PR #11, opened 2024-02-01
Update SlotViz tile behavior

D13ce:
This PR updates the SlotViz's tile behavior: Missed slots without duties now show a green box with a red border.

The mocked data for the SlotViz has been improved to show that. It also now contains a missed slot with mixed duties (not sure if that can ever happen but the frontend is able to deal with that).


MauserBitfly (approved):
LGTM

PR #172, opened 2024-04-03
api: return error instead of silently capping if limit is too high

invis-bitfly:
```json
{
"error": "given value '100000' for parameter 'limit' is too large, maximum limit is 100"
}
```


LuccaBitfly (approved):
lgtm

PR #200, opened 2024-04-12
Nobids/cursor reverse flag

invis-bitfly:
no more direction comparisons, simply check if `cursor.IsReverse()` is `true`


Eisei24 (approved):
lgtm, merge conflicts still need to be handled.

PR #297, opened 2024-05-14
execution rewards

invis-bitfly:
adds exporter task to fetch missing rewards to postgres each time an epoch is justified. it automatically fills whatever is missing. no manual backfill should be needed. migration files is there. there is a cached materialized view that joins with the relay rewards for dashboard usage

depends on https://github.com/gobitfly/eth2-beaconchain-explorer/pull/2879


invis-bitfly (commented):
i just did a self review and the conclusion is that it indeed looks like my code

PR #361, opened 2024-05-31
invis merge only: clickhouse db connect

invis-bitfly:
this should be enough if you want to mess with clickhouse over the weekend. if you need any information such as how to access clickhouse from home, can you make me a db acc, etc let me know


invis-bitfly (commented):
ye

PR #380, opened 2024-06-05
Bids 3086/swipe away

MauserBitfly:
This PR:
- Adds a `useSwipe`composable to track swiping of elelemnts
- The (dynamic) dialogs use this composable to close when swiped 

To make sure you don't swipe on content in the modal (like in lists) we prevent swiping inside the `p-dialog-content` ...this logic can be improved in the future.


thib-wien:
In this implementation, it is hard to swipe because most of the area of the modal is ignored (`div` of class _p-dialog-content_ ). The swipable area is limited to the narrow padding around the modal.

An idea that I have (probably wrong) would be to remove `invalidSwipeClasses` and to change the logic of `hasClassOrParentWithClass()` :
this function would keep its recursive logic, but instead of searching for a class name over the element and in its parents, it would check whether they have `offsetWidth < scrollWidth` or `offsetHeight < scrollHeight` (meaning that the finger is on a scrollable area).

thib-wien (changes requested):
impressive work that you did so fast :)

MauserBitfly (commented):
Thanks for the input, checking for a scroll-able container (+ I added input to the check) makes it much easier to close the modal on swipe

PR #436, opened 2024-06-11
mark scheduled blocks of the past as probably missed

MauserBitfly:
This PR:
- marks scheduled blocks that are before the current slot from the latest state as (probably) missed* 

image


D13ce (approved):
LGTM (which is no surprise really as you have simply implemented **MY** brilliant idea)

PR #443, opened 2024-06-12
don't default to shared dashboard

MauserBitfly:
This PR:
- prevents from defaulting to the shared dashboard if you navigate to /dashboard


D13ce (approved):
LGTM

PR #450, opened 2024-06-12
(NOBIDS) hand-cursor over dashboard menus + larger filter-input in tables

thib-wien:
This PR:

Shows a hand cursor anywhere over the buttons + in the drop-down of the MenuBar:
image

Enlarges by +22% the input field of the filter at the top of tables, so that its placeholders are never truncated like that:
filter


MauserBitfly (approved):
LGTM

PR #451, opened 2024-06-12
ensure disabled chart toggle uses the correct style

MauserBitfly:
This PR:
- fixes the disabled toggle style on initial page load 

info for tester:
The chart toggle did not correctly apply the disabled style on intial page load ... it only worked after switiching between tabs
image


frontend/components/bc/BcIconToggle.vue
```diff
@@ -48,7 +48,7 @@ const toggle = () => {
   height: 23px;
   cursor: pointer;
 
-  &[disabled='true']{
+  &[disabled]{
```
thib-wien, line 51: Sehr sehr praktisch feature of Vue!
I will use it to simplify my code.

thib-wien (approved):

PR #454, opened 2024-06-12
Improve text logic on shared dashboard view modal

D13ce:
This PR
* Causes the "Shared Dashboard View"-modal to have a different text whenever the user already has at least 1 local/private dashboard


MauserBitfly (approved):
LGTM

PR #457, opened 2024-06-12
several small fixes

MauserBitfly:
This PR:
- routes the footer imprint to https://beaconcha.in/imprint
- removes the footer link to the API Docs
- fixes some paddings in the blocks table so that the rows have the same height as in the other tables
- fixes the transparent menu button (when a disabled button is a child of the main button on mobile) 
image


D13ce (approved):
LGTM

PR #471, opened 2024-06-13
hide tooltip in a scrolling scroll container

MauserBitfly:
This PR:
- hides the Tooltip if any of it's scrollable parents is scrolled
- it also hides the tooltip on window resize 

Add. Infos:
- I left the `scrollContainer` (class provided by the outside) functionality in (even dough the new logic should tackle most cased needed)

### DIFF
--- a/frontend/components/bc/BcTooltip.vue
+++ b/frontend/components/bc/BcTooltip.vue
@@ -16,7 +16,7 @@ interface Props {
 const props = defineProps<Props>()
 const bcTooltipOwner = ref<HTMLElement | null>(null)
 const bcTooltip = ref<HTMLElement | null>(null)
-let scrollParent: HTMLElement | null = null
+let scrollParents: HTMLElement[] = []
 const tooltipAddedTimeout = ref<NodeJS.Timeout | null>(null)
 const { selected, doSelect } = useTooltipStore()
 const { width, height } = useWindowSize()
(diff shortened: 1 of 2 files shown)


thib-wien (commented):
I see that it works well when resizing the window, everywhere I tried 👍🏻

I found on configuration with which the tooltip remained displayed while I scrolled.
To reproduce:

1. search for an address returning several results. Example:
`0x778F5F13C4Be78A3a4d7141BCB26999702f407CF`
2. decrease the height of the window enough to force the content div of the management modal to become scrollable:
Screenshot from 2024-06-13 13-37-13
3. Click the diamond:
Screenshot from 2024-06-13 13-38-05
4. A page opens. Close it. Now, the tooltip is sill displayed. Scroll down:
5. 
Screenshot from 2024-06-13 13-41-25

MauserBitfly:
> I see that it works well when resizing the window, everywhere I tried 👍🏻
> 
> I found one configuration with which the tooltip remained displayed while I scrolled. To reproduce:
> 
> 1. search for an address returning several results. Example:
>    `0x778F5F13C4Be78A3a4d7141BCB26999702f407CF`
> 2. decrease the height of the window enough to force _the content div of the management modal_ to become scrollable:
>    Screenshot from 2024-06-13 13-37-13
> 3. Click the diamond:
>    Screenshot from 2024-06-13 13-38-05
> 4. A page opens. Close it.
> 5. The tooltip is still displayed. Scroll down: the tooltip does not disappear:
>    Screenshot from 2024-06-13 13-41-25

I don't think that this hight is supported by the explorer and you can't use it. But what the heck now also this should work.

thib-wien (approved):

PR #472, opened 2024-06-13
convert nano average to time duration

MauserBitfly:
This PR
- correctly converts the luck average from nano seconds to a time duration


D13ce (approved):
LGTM

PR #474, opened 2024-06-13
Hide timeframe related values on /pricing

D13ce:
This PR
* Hides all values, where we don't know what we can support yet, by writing "Soon" in italic instead

### DIFF
--- a/frontend/components/pricing/PremiumCompare.vue
+++ b/frontend/components/pricing/PremiumCompare.vue
@@ -31,7 +31,7 @@ const showContent = ref(false)
 const rows = computed(() => {
   const sorted = products.value?.premium_products?.sort((a, b) => a.price_per_month_eur - b.price_per_month_eur) ?? []
   const rows: CompareRow[] = []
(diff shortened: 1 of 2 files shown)


frontend/components/pricing/PremiumCompare.vue
```diff
@@ -47,10 +49,11 @@ const rows = computed(() => {
 
     return {
       value,
-      tooltip
+      tooltip,
+      class: undefined as string | undefined
```
MauserBitfly, line 53: that is not pretty, I would suggest to define the return type CompareValue
```
const mapValue = (property: string, perks: PremiumPerks):CompareValue => {
```

then you should be able to remove the whole row here.

MauserBitfly (commented):
looks good, just one Nitpick

frontend/components/pricing/PremiumCompare.vue
D13ce: Obviously better. I couldn't wrap my head around that at first. Thanks!

MauserBitfly (approved):
LGTM

PR #476, opened 2024-06-13
(NOBIDS) small improvements

thib-wien:
Now:

- _BcTooltip.vue_ activates the listeners only when the tooltip is shown and turns them off when the tooltip is hidden.
This makes the UI to run less listeners than before (dozens on some pages).

- `refreshLatestState()` and `getUser()` return the data that they load in case of success or `undefined` in case of problem.
This allows `useAsyncData()` to call the API less often.


MauserBitfly (approved):
LGTM

PR #479, opened 2024-06-14
min height for nav bar 2 on desktop

MauserBitfly:
This PR
- fixes the problem that on desktop on the initial page load the second row of the top header is too small and only fixes itself after hydration.


frontend/components/bc/header/MainHeader.vue
D13ce, line 1: **Bug:** The height of the row changed slightly. Here is a side by side comparison (left: your branch; right: production):
image

D13ce (changes requested):
The bug itself is changed, well done. But the height of the row changed too (oh so slightly). PTAL :point_down:

frontend/components/bc/header/MainHeader.vue
D13ce: As discussed, it was wrong BEFORE and now is correct. Well done.

D13ce (approved):
LGTM

PR #483, opened 2024-06-14
(NOBIDS) small fix in the logic of getUser()

thib-wien:
If undefined data was received from the API, `getUser()` would return `undefined` to inform `useAsyncData` that it must recall the API from the client.
The user was not set to `undefined` in the meantime. This was a mistake.

Also:
The bottom border of the header should be visible when `NUXT_PUBLIC_SHOW_IN_DEVELOPMENT` is `"false"` and the screen is large.
It was not the case. Fixed.


MauserBitfly (approved):
LGTM

PR #485, opened 2024-06-14
(BIDS-2959) pressing Enter or the + button does not bypass premium restrictions anymore

thib-wien:
To test it:

1. In the management modal, perform a premium-restricted search, for example `0x778F5F13C4Be78A3a4d7141BCB26999702f407CF`.
2. Press `Enter` or `+`.
3. The UI does not react anymore. The dropdown stays open and shows the premium diamonds.

If you set `mock: true` in _types/customFetch.ts_, then you can test what happens when the dropdown contains a mix of premium and non-premium results:
the premium rows are skipped when the user presses `Enter` or `+`, the first non-premium row gets added to the dashboard.


MauserBitfly (approved):
LGTM

PR #486, opened 2024-06-17
Remove "per dashboard" on guppy

D13ce:
This PR
* Removes the "per dashboard" text for the validator amount on the `Guppy` card (as `Guppy` only has a single validator dashboard)


MauserBitfly (approved):
LGTM

PR #487, opened 2024-06-17
show hide icon

MauserBitfly:
This PR:
- adds a 'visible' icon to the Slot Viz filters: if disabled then no icons in the slot viz will  be shown 
- set the visible icon per default to hidden, if the dashboard has more then 60 validators. *
- adds tooltips to the Slot Viz filter icons

* Note: this will only be applied the first time the user navigates to the dashboard then it will be stored in the cookie

### DIFF
--- a/frontend/components/bc/toggle/MultiBarButton.vue
+++ b/frontend/components/bc/toggle/MultiBarButton.vue
@@ -19,7 +19,7 @@ const icon = computed(() => {
 </script>
 
 <template>
-  <BcTooltip :dont-open-permanently="true">
+  <BcTooltip :dont-open-permanently="true" :hover-delay="350">
     <template #tooltip>
       <div class="button-tooltip">
         <div>{{ selected ? $t('filter.enabled'): $t('filter.disabled') }}</div>
--- a/frontend/components/bc/toggle/MultiBarButton.vue
+++ b/frontend/components/bc/toggle/MultiBarButton.vue
@@ -19,7 +19,7 @@ const icon = computed(() => {
 </script>
 
 <template>
-  <BcTooltip>
+  <BcTooltip :dont-open-permanently="true">
     <template #tooltip>
       <div class="button-tooltip">
         <div>{{ selected ? $t('filter.enabled'): $t('filter.disabled') }}</div>
--- a/frontend/components/bc/toggle/MultiBarButton.vue
+++ b/frontend/components/bc/toggle/MultiBarButton.vue
@@ -1,10 +1,12 @@
 <script setup lang="ts">
 import type { IconDefinition } from '@fortawesome/fontawesome-svg-core'
 import { FontAwesomeIcon } from '@fortawesome/vue-fontawesome'
+import BcTooltip from '../BcTooltip.vue'
 
 interface Props {
   icon?: IconDefinition,
   falseIcon?: IconDefinition,
+  tooltip?: string,
 }
 
 const props = defineProps<Props>()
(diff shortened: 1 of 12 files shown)


frontend/stores/dashboard/useValidatorDashboardOverviewStore.ts
```diff
@@ -47,5 +47,15 @@ export function useValidatorDashboardOverviewStore () {
...
 
-  return { overview, refreshOverview, hasValidators }
+  const validatorCount = computed(() => {
+    if (!overview.value) {
+      return undefined
+    }
+    if (!overview.value?.validators) {
+      return 0
+    }
+    return overview.value?.validators.exited + overview.value?.validators.offline + overview.value?.validators.online + overview.value?.validators.pending + overview.value?.validators.slashed
```
D13ce, line 57: **Nitpick:** No need for the `?` here, you have already checked its availability above.
```suggestion
    return overview.value.validators.exited + overview.value.validators.offline + overview.value.validators.online + overview.value.validators.pending + overview.value.validators.slashed
```

frontend/components/bc/toggle/MultiBarButton.vue
```diff
@@ -17,16 +19,33 @@ const icon = computed(() => {
...
-      </slot>
+  <BcTooltip>
+    <template #tooltip>
+      <div class="button-tooltip">
+        <div>{{ selected ? $t('filter.enabled'): $t('filter.disabled') }}</div>
+        <div v-if="tooltip" class="individual">
+          {{ tooltip }}
+        </div>
+      </div>
     </template>
```
D13ce, line 30: **Bug:** Clicking the button now forces the tooltip to be visible until the users clicks somewhere else again.

frontend/components/bc/toggle/MultiBarButton.vue
```diff
@@ -17,16 +19,33 @@ const icon = computed(() => {
...
-    <template #icon="slotProps">
-      <slot name="icon" v-bind="slotProps">
-        <FontAwesomeIcon v-if="icon" :icon="icon" />
-      </slot>
+  <BcTooltip>
+    <template #tooltip>
+      <div class="button-tooltip">
+        <div>{{ selected ? $t('filter.enabled'): $t('filter.disabled') }}</div>
+        <div v-if="tooltip" class="individual">
+          {{ tooltip }}
```
D13ce, line 27: **Request:** I know it is designed like this, but please ask the product owner/designer if the order of these two texts can be changed. Reading "If this is enabled, it does X" BEFORE "Filter is enabled" flows better I believe.

D13ce (changes requested):
It is working as requested but I have left some comments below, PTAL :point_down: 

Furthermore: Personally, I am a bit confused by this feature (not your fault of course). It does not feel intuitive and that isn't good if its objective is to make things clearer.

Maybe move the eye symbol to the very left as? And/Or maybe improve its tooltip to something like "If enabled, all duty icons for the duties selected to the right will be shown."

Feel free to discuss this with the product owner/designer.

frontend/stores/dashboard/useValidatorDashboardOverviewStore.ts
MauserBitfly: true

frontend/components/bc/toggle/MultiBarButton.vue
MauserBitfly: fixed

frontend/components/bc/toggle/MultiBarButton.vue
MauserBitfly: changed

MauserBitfly (commented):
Thanks, I also added the 3rd point of the BIDS, requesting a delay when you hover over the slot viz icons

PR #488, opened 2024-06-17
(BIDS-3135) pricing page: the premium/API switch is in a sticky banner

thib-wien:


frontend/pages/pricing.vue
D13ce, line 1: **Bug:** On mobile (Firefox; resolution used in video: 400px / 877px) it is possible to scroll down far enough so that the sticky element is hidden:
 
[Screencast from 2024-06-17 15:41:24.webm](https://github.com/gobitfly/beaconchain/assets/114066135/67074d80-96be-4ba1-8acf-feb807176b52)

There is a short break once the bottom is reached, continuing scrolling down will cause the toggle to disappear.

Once this state is reached, scrolling up and keeping scrolling once the top is reached, will re-show the element eventually.

D13ce (changes requested):
Looks good, only found one weird bug.

frontend/pages/pricing.vue
thib-wien: This is the normal behavior of sticky divs: they cannot leave their parent.
To work around it, I am afraid that we would need to replace the Premium/API switch with an empty div (whose height would be hard-coded), and move the switch into _PageWrapper.vue_ at the same DOM level as the footer.

Is it necessary? None of the smartphones simulated by Firefox cause the problem. Here is how far I can scroll down with the smallest of all screens and the banner is still visible so the bug does not happen:
image
This screen has a height of 667 pixels so I wonder why your browser bugs with 877 px.

Maybe you have a zoom, so the canvas height in your test is smaller than the video height?

D13ce: As discussed, this was caused because an error on my side: My client window's height was set to 877px but because of my open Inspector, the browser window was not high enough.

D13ce (approved):
LGTM

PR #489, opened 2024-06-17
let's do some maintenance

MauserBitfly:
This PR:
- introduces the new .env var `NUXT_PUBLIC_MAINTENANCE_TS` 
- if set we show a maintenance banner (text depends on if the ts is in the past or not)


frontend/components/bc/BcMaintenanceBanner.vue
```diff
@@ -0,0 +1,47 @@
...
+const { public: { maintenanceTS } } = useRuntimeConfig()
+const { tick } = useInterval(60)
+const { t: $t } = useI18n()
+
+const maintenanceLabel = computed(() => {
+  if (!maintenanceTS) {
+    return
+  }
+  const parsed = typeof maintenanceTS === 'number' ? maintenanceTS : parseInt(maintenanceTS)
+  if (isNaN(parsed)) {
```
thib-wien, line 13: `parseInt()` is more tolerant than `Int()`. If the string contains "129,394,653" it returns 129 and ignores the rest... I believe.
`Int()` would return `Nan`.

Maybe, the `? :` is not necessary, I imagine that this line would work the same way:
```TS
const parsed = Int(maintenanceTS)
```

frontend/components/bc/BcMaintenanceBanner.vue
```diff
@@ -0,0 +1,47 @@
...
+    warn('NUXT_PUBLIC_MAINTENANCE_TS is not convertable to an intager, a unix ts is expected')
```
thib-wien, line 14: The words might be misspelled. I would write "convert**i**ble" and "int**e**ger".

frontend/components/bc/BcMaintenanceBanner.vue
```diff
@@ -0,0 +1,47 @@
...
+    return undefined
+  } else if (parsed === 0) {
```
thib-wien, line 16: maybe `parsed <= 0`, so if Peter schedules a maintenance before 1970 we do not display the banner because we hope it is finished now.

frontend/components/bc/BcMaintenanceBanner.vue
```diff
@@ -0,0 +1,47 @@
...
+    return
+  }
+  const ts = new Date(parseInt(maintenanceTS) * 1000).getTime()
```
thib-wien, line 19: Would `parsed` fit here, instead of `parseInt(maintenanceTS)` ?

frontend/components/bc/BcMaintenanceBanner.vue
```diff
@@ -0,0 +1,47 @@
...
+
+const { public: { maintenanceTS } } = useRuntimeConfig()
+const { tick } = useInterval(60)
+const { t: $t } = useI18n()
+
+const maintenanceLabel = computed(() => {
+  if (!maintenanceTS) {
+    return
+  }
+  const parsed = typeof maintenanceTS === 'number' ? maintenanceTS : parseInt(maintenanceTS)
```
thib-wien, line 12: Maybe, the `? :` is not necessary, I imagine that this line would work the same way:
```TS
const parsed = Int(maintenanceTS)
```

frontend/components/bc/BcMaintenanceBanner.vue
MauserBitfly: yes that was the plan, but somehow got lost ;-)

frontend/components/bc/BcMaintenanceBanner.vue
MauserBitfly: fixed

frontend/components/bc/BcMaintenanceBanner.vue
MauserBitfly: Did not think about that, but I think it would be ok that... so at least we show the banner if someone added it with `,`

frontend/components/bc/BcMaintenanceBanner.vue
MauserBitfly: I don't think there is an `Int` constructor in JS. Do you mean Number?

thib-wien: 🤦🏻

thib-wien (approved):
:)

PR #492, opened 2024-06-18
fix watch error in SlotVizViewer

MauserBitfly:
This PR
- fixes an error thrown by a watch in the SlotVizViewer

image

### DIFF
--- a/frontend/components/slot/viz/SlotVizViewer.vue
+++ b/frontend/components/slot/viz/SlotVizViewer.vue
@@ -71,7 +71,7 @@ const currentSlotId = computed(() => {
   return Math.max(mostRecentScheduledSlotId.value ?? 0, tsToSlot((props.timestamp ?? 0) / 1000) - 1)
 })
 
-watch(props, () => {
+watch(() => props, () => {
   if (props.initiallyHideVisible !== undefined) {
     const initialIndex = selectedCategories.value.indexOf('initial')
     if (initialIndex < 0) {


D13ce (changes requested):
Error is gone, good catch :+1: 

**BUT!** There are several more locations in the code where your change should most likely be applied.

Please search for `watch(props`, this will show some results, for example
https://github.com/gobitfly/beaconchain/blob/068946bd529f3ee533f4e0feab9bb4d6fda29db4/frontend/components/dashboard/DashboardRenameModal.vue#L18

---

Furthermore, here is am example where I believe we might have to add a `deep: true` to the watcher:
https://github.com/gobitfly/beaconchain/blob/b1b803855d8572cd0c7009e4195a1d4547047961/frontend/components/bc/toggle/MultiBar.vue#L19

Please check other watchers too.

---

:information_source: Please note that I did not find any error log (neither in VSC for SSR nor in my browser) for the mentioned code parts. So maybe I am misunderstanding this completely :upside_down_face:

MauserBitfly:
> Error is gone, good catch 👍
> 
> **BUT!** There are several more locations in the code where your change should most likely be applied.
> 
> Please search for `watch(props`, this will show some results, for example
> 
> https://github.com/gobitfly/beaconchain/blob/068946bd529f3ee533f4e0feab9bb4d6fda29db4/frontend/components/dashboard/DashboardRenameModal.vue#L18
> 
> Furthermore, here is am example where I believe we might have to add a `deep: true` to the watcher:
> 
> https://github.com/gobitfly/beaconchain/blob/b1b803855d8572cd0c7009e4195a1d4547047961/frontend/components/bc/toggle/MultiBar.vue#L19
> 
> Please check other watchers too.
> 
> ℹ️ Please note that I did not find any error log (neither in VSC for SSR nor in my browser) for the mentioned code parts. So maybe I am misunderstanding this completely 🙃

these should be fine as they are reactive properties.

D13ce (approved):
LGTM

PR #495, opened 2024-06-18
Bids 3090/mobile adaptions for pricing

D13ce:
This PR
* Increases some font sizes in the premium product boxes on `/pricing` on mobile
* Changes the `cost per validator` calculation from [`price` / `validators per dashboard`] to [`price` / `validators per dashboard` / `validator dashboards`] and shows an extra digit for that value

All changes have been discussed with @Buttaa.


MauserBitfly (approved):
LGTM

PR #497, opened 2024-06-18
(BIDS-3040) Improves the chart of rewards

thib-wien:
This PR:

- changes the broken curve for a smooth curve,
- removes the snapshot dots from the curve,
- adjusts the scale of the chart automatically so the empty space under the curve is limited.


MauserBitfly (approved):
LGTM

PR #509, opened 2024-06-20
Bids 3167/user creation v2

remoterami:


backend/pkg/api/data_access/user.go
```diff
@@ -6,19 +6,33 @@ import (
...
 
+	"github.com/gobitfly/beaconchain/pkg/api/services"
 	t "github.com/gobitfly/beaconchain/pkg/api/types"
 	"github.com/gobitfly/beaconchain/pkg/commons/utils"
 	"github.com/pkg/errors"
 )
 
 type UserRepository interface {
+	GetUserExists(email string) (bool, error)
+	AddUser(email, password string) error
```
LuccaBitfly, line 17: Minor:
```suggestion
	CreateUser(email, password string) error
```

backend/pkg/api/data_access/user.go
```diff
@@ -6,19 +6,33 @@ import (
...
 	GetUserValidatorDashboardCount(userId uint64) (uint64, error)
 }
 
+func (d *DataAccessService) GetUserExists(email string) (bool, error) {
+	// TODO @patrick
+	return d.dummy.GetUserExists(email)
+}
+
+func (d *DataAccessService) AddUser(email, password string) error {
+	// TODO @patrick
```
LuccaBitfly, line 31: Minor: These probably shouldn't have a `@patrick` :)

backend/pkg/api/data_access/user.go
```diff
@@ -6,19 +6,33 @@ import (
...
+	d.services.QueueEmail(services.EMail{Recipient: email, Message: "Click to confirm email"}, 60*5)
```
LuccaBitfly, line 32: Sending the e-mail is not part of the data access and should be done inside the handler, since you're not fetching or manipulating data in our databases. This also means that the handler service needs direct access to the email service and this func is going to have to return some confirmation hash for the message though.

backend/pkg/api/handlers/auth.go
```diff
@@ -85,6 +85,60 @@ func (h *HandlerService) InternalPostApiKeys(w http.ResponseWriter, r *http.Requ
...
+		Email    string `json:"email"`
+		Password string `json:"password"`
+	}{}
+	if err := v.checkBody(&req, r); err != nil {
+		handleErr(w, err)
+		return
+	}
+
+	// validate email
+	email := v.checkEmail(req.Email)
```
LuccaBitfly, line 101: Should probably make sure the email is converted to lowercase in this check.

LuccaBitfly (changes requested):
Changes look mostly good, I have a few small issues though.

backend/pkg/api/data_access/user.go
remoterami: Anyone else instead, like `recy21` or `DATA-ACCESS`? Or just `TODO`? I'll just suggest `@DATA-ACCESS` for now

LuccaBitfly: both recy and data access are fine, or even a naked `// TODO` as we're creating a specific follow up bids anyways

LuccaBitfly (approved):
lgtm

PR #511, opened 2024-06-21
(BIDS-3166) API: Indicate archived dashboards

remoterami:


backend/pkg/api/types/validator_dashboard.go
```diff
@@ -32,6 +32,11 @@ type VDBOverviewData struct {
 
 type InternalGetValidatorDashboardResponse ApiDataResponse[VDBOverviewData]
 
+type VDBPostArchivingReturnData struct {
+	Id       uint64 `json:"id"`
+	Archived bool   `json:"archived"`
```
LuccaBitfly, line 37: Minor:
```suggestion
	IsArchived bool   `json:"is_archived"`
```

backend/pkg/api/types/dashboard.go
```diff
@@ -5,10 +5,15 @@ type AccountDashboard struct {
 	Name string `json:"name"`
 }
 type ValidatorDashboard struct {
-	Id        uint64        `json:"id"`
-	Name      string        `json:"name"`
-	PublicIds []VDBPublicId `json:"public_ids,omitempty"`
+	Id             uint64        `json:"id"`
+	Name           string        `json:"name"`
+	PublicIds      []VDBPublicId `json:"public_ids,omitempty"`
+	Archived       bool          `json:"archived"`
```
LuccaBitfly, line 11: Minor:
```suggestion
	IsArchived       bool          `json:"is_archived"`
```

backend/pkg/api/handlers/internal.go
```diff
@@ -293,6 +293,31 @@ func (h *HandlerService) InternalDeleteValidatorDashboard(w http.ResponseWriter,
 	returnNoContent(w)
 }
 
+func (h *HandlerService) InternalPutValidatorDashboardArchiving(w http.ResponseWriter, r *http.Request) {
+	var v validationError
+	dashboardId := v.checkPrimaryDashboardId(mux.Vars(r)["dashboard_id"])
+	req := struct {
+		Archived bool `json:"archived"`
```
LuccaBitfly, line 300: Minor:
```suggestion
		IsArchived bool `json:"is_archived"`
```

backend/pkg/api/handlers/internal.go
```diff
@@ -293,6 +293,31 @@ func (h *HandlerService) InternalDeleteValidatorDashboard(w http.ResponseWriter,
 	returnNoContent(w)
 }
 
+func (h *HandlerService) InternalPutValidatorDashboardArchiving(w http.ResponseWriter, r *http.Request) {
```
LuccaBitfly, line 296: This is func is missing checks as to whether unarchiving is allowed. It should go through all perks that could be violated and check against them before unarchiving (may require new data access funcs)

LuccaBitfly (changes requested):

backend/pkg/api/handlers/internal.go
remoterami: Right missed that, added handling for this.

But while thinking about permissions, I noticed that the other way should also be checked - if a dashboard is archived, basically no VDB endpoint should return any actual data, right? Unless I'm missing what "dashboard archival" is even about.

So in order to handle this without copy pasting the same archival status handling 20+ times for all endpoints, I also added a suggestion by messing with the `GetVDBAuthMiddleware` (hopefully not breaking staging until implemented by data access). If that's not clean enough we could move the `/archival` endpoint to another route, but lmk what you think

backend/pkg/api/types/dashboard.go
```diff
@@ -5,10 +5,15 @@ type AccountDashboard struct {
...
 }
 type ValidatorDashboard struct {
-	Id        uint64        `json:"id"`
-	Name      string        `json:"name"`
-	PublicIds []VDBPublicId `json:"public_ids,omitempty"`
+	Id             uint64        `json:"id"`
+	Name           string        `json:"name"`
+	PublicIds      []VDBPublicId `json:"public_ids,omitempty"`
+	IsArchived     bool          `json:"is_archived"`
+	ArchivedReason string        `json:"archived_reason,omitempty"` // dashboard_limit, validator_limit, group_limit, none
```
LuccaBitfly, line 12: Frontend would probably like this:
```suggestion
	ArchivedReason string        `json:"archived_reason,omitempty" tstype:"'dashboard_limit' | 'validator_limit' | 'group_limit'"`
```
Don't forget to convert to ts

LuccaBitfly (changes requested):

LuccaBitfly (approved):
lgtm

PR #514, opened 2024-06-24
Bids 3085/summary table v2 5

MauserBitfly:
This PR:
- adapts the Summary Table (rows + details) according to the 2.5 spec

This PR:
- does not adapt the changes for the `DashboardValidatorSubsetModal` this will be done in a follow up PR


thib-wien:
test

frontend/components/dashboard/table/SummaryDetails.vue
```diff
@@ -1,60 +1,77 @@
 <script setup lang="ts">
 import type { VDBSummaryTableRow } from '~/types/api/validator_dashboard'
```
thib-wien, line 2: Another example here, it is very unclear what are 16 and 42 here:
image
16 and 42 do not add up to the "past" number.
So what are they?

frontend/components/dashboard/table/SummaryDetails.vue
```diff
@@ -1,60 +1,77 @@
...
-import { type SummaryDetailsEfficiencyCombinedProp, type SummaryRow } from '~/types/dashboard/summary'
```
thib-wien, line 3: A small detail on mobile:
image
the tip of the bubble points to the void. Is there a way to move it above the `i`?

frontend/components/dashboard/table/SummaryDetails.vue
```diff
@@ -1,60 +1,77 @@
...
-import { TimeFrames, type TimeFrame } from '~/types/value'
+import { type SummaryDetailsEfficiencyCombinedProp, type SummaryRow, type SummaryTableVisibility, type SummaryTimeFrame } from '~/types/dashboard/summary'
```
thib-wien, line 4: A question regarding the mobile view:
image
when the screen width is so tight, Figma shows that that three Power buttons spread vertically:
image

thib-wien (commented):
Very beautiful and consequent work.
Not surprising

frontend/components/bc/format/FormatPercent.vue
```diff
@@ -31,18 +43,60 @@ const data = computed(() => {
...
+    } else if (percent > props.comparePercent) {
+      className = 'text-positive'
+      leadingIcon = faArrowUp
+      compareResult = 'higher'
+    } else {
+      className = 'text-negative'
+      leadingIcon = faArrowDown
+      compareResult = 'lower'
+    }
+  } else if (props.colorBreakPoint) {
```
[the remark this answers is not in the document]
MauserBitfly: As it makes no sense `comparePercent`and `colorBreakPoint`at the same time I would prefer to keep it that way. 

I added a comment to the interface definition

frontend/components/dashboard/table/DashboardTableSummary.vue
```diff
@@ -1,10 +1,16 @@
 <script setup lang="ts">
```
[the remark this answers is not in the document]
MauserBitfly: The visibility of the 'add validators +' component depends on the fact that the user has added validators to the dashboard and not if the summary list has items. This information is taken from the overview data.

frontend/components/dashboard/table/SummaryDetails.vue
```diff
@@ -1,60 +1,77 @@
 <script setup lang="ts">
```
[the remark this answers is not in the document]
MauserBitfly: renamed to 'Validators'

frontend/components/dashboard/table/SummaryDetails.vue
MauserBitfly: the 
`green / red ` scheme is used for `good / bad`through out the UI ... feel free to discuss this with Inan at one point (but this would effect the whole application) .

The numbers don't add up because they are randomly mocked by the backend

frontend/components/dashboard/table/SummaryDetails.vue
MauserBitfly: Yes this is a minor Bug in the Tooltip component, as the tooltip tries to center itself over the parent but ensures it stays visible. The small triangle is just centered relative to the tooltip and does not consider when the tooltip has to move. 

As this is a general Bug and not related to this PR, I would prefer to fix that separately. Would be cool if you could create a Bug for it so it does not get forgotten.

frontend/components/dashboard/table/SummaryDetails.vue
MauserBitfly: good catch a `wrap`was missing ;-)

frontend/components/dashboard/table/SummaryDetails.vue
```diff
@@ -64,97 +81,108 @@ const rowClass = (data:SummaryRow) => {
...
+        </div>
+        <DashboardTableSummaryValue
+          v-if="prop.prop"
+          class="value"
+          :data="summary"
+          :absolute="absolute"
+          :property="prop.prop"
+          :time-frame="timeFrame"
+          :row-x="props.row"
+          :row="{ ...props.row, reward: { el: '1239082134123213', cl: '02394' } }"
```
thib-wien, line 98: maybe write a `/* TODO: blabla */` between the last `}` and `"` ?

frontend/components/dashboard/table/SummaryStatus.vue
```diff
@@ -0,0 +1,113 @@
...
+      key,
+      className: count > 0 ? `${key.replace('_', '-')} active` : '',
+      tooltip: $t(`dashboard.validator.summary.status.${key}`, {}, count)
+    }
+  }
+
+  // TODO: replace with v2.5 logic (once we got the api structs)
+  const scheduledSyncCount = props.status?.next_sync_count
+  const currentSyncCount = props.status?.current_sync_count
+  const slashedCount = props.status?.slashed_count
```
thib-wien, line 26: On lines 24, 25 and 26, the `?` is ineffective because line `7` makes `prop.status` mandatory.

frontend/components/dashboard/table/SummaryValidators.vue
```diff
@@ -0,0 +1,155 @@
...
+}
+
+const groupName = computed(() => {
+  return getGroupLabel($t, props.groupId, groups.value)
+})
+
+const mapped = computed(() => {
+  const list: { count: number, key: string }[] = []
+  const addCount = (key: string, count?: number) => {
+    if (count) {
```
thib-wien, line 49: If `count` is `0`, the corresponding Power icon will not be displayed. Oder?
As a user, I would like to always see the 3 icons, to see that I have 0 slashing or exited validator, so I know that the UI will tell me if it happens one day.

MauserBitfly: That is by design. But I enhanced the logic for the absolute check (where we usually display only online validators) to display offline/exited if there is no online validator .

thib-wien (approved):
🚀

PR #518, opened 2024-06-25
(BIDS-3165) Api: User Account Management

remoterami:


backend/pkg/api/router.go
```diff
@@ -94,8 +94,11 @@ func addRoutes(hs *handlers.HandlerService, publicRouter, internalRouter *mux.Ro
 		{http.MethodGet, "/product-summary", nil, hs.InternalGetProductSummary},
 
 		{http.MethodPost, "/users", nil, hs.InternalPostUsers},
-
 		{http.MethodGet, "/users/me", nil, hs.InternalGetUserInfo},
+		//{http.MethodDelete, "/users/me", nil, hs.InternalDeleteUser},
+		//{http.MethodPut, "/users/me/email", nil, hs.InternalPutUserEmail},
+		//{http.MethodPut, "/users/me/password", nil, hs.InternalPutUserPassword},
+		//{http.MethodPut, "/users/me/subscriptions", nil, hs.InternalPutUserSubscriptions},
```
LuccaBitfly, line 101: Changing the subscription is not really part of this BIDS (that will be done on the pricing page I guess), other endpoints look fine

LuccaBitfly (changes requested):

backend/pkg/api/handlers/auth.go
```diff
@@ -253,6 +289,146 @@ func (h *HandlerService) InternalPostLogout(w http.ResponseWriter, r *http.Reque
...
+		handleErr(w, err)
+		return
+	}
+	userInfo, err := h.dai.GetUserInfo(user.Id)
+	if err != nil {
+		handleErr(w, err)
+		return
+	}
+	if !userInfo.Confirmed {
+		handleErr(w, errors.New("cannot update email for an unconfirmed address"))
```
LuccaBitfly, line 322: We should probably allow changing the email here, e.g. if the user has used a wrong email during signup. Otherwise, this line would cause a 500 to be returned, should rather be 409.

backend/pkg/api/handlers/auth.go
```diff
@@ -253,6 +289,146 @@ func (h *HandlerService) InternalPostLogout(w http.ResponseWriter, r *http.Reque
...
+	}
+
+	// validate new email
+	newEmail := v.checkEmail(req.Email)
+	if v.hasErrors() {
+		handleErr(w, v)
+		return
+	}
+	if newEmail == userInfo.Email {
+		handleErr(w, errors.New("can't reuse current email"))
```
LuccaBitfly, line 344: This returns 500, should return 409

backend/pkg/api/handlers/auth.go
```diff
@@ -253,6 +289,146 @@ func (h *HandlerService) InternalPostLogout(w http.ResponseWriter, r *http.Reque
...
+		return
+	}
+
+	userExists, err := h.dai.GetUserExists(newEmail)
+	if err != nil {
+		handleErr(w, err)
+		return
+	}
+	if userExists {
+		handleErr(w, errors.New("email already registered"))
```
LuccaBitfly, line 354: Should return 409

backend/pkg/api/router.go
```diff
@@ -94,8 +94,12 @@ func addRoutes(hs *handlers.HandlerService, publicRouter, internalRouter *mux.Ro
 		{http.MethodGet, "/product-summary", nil, hs.InternalGetProductSummary},
 
 		{http.MethodPost, "/users", nil, hs.InternalPostUsers},
-
+		{http.MethodPut, "/users/confirm/{token}", nil, hs.InternalPostUserConfirm},
```
LuccaBitfly, line 97: Should be a POST request.
Minor: Also I'd like to stay RESTful and avoid verbs in URLs if possible, so I'd prefer this path:
```suggestion
		{http.MethodPost, "/users/email-confirmations/{token}", nil, hs.InternalPostUserConfirm},
```

backend/pkg/api/types/user.go
```diff
@@ -3,6 +3,7 @@ package types
 type UserInfo struct {
 	Id            uint64             `json:"id"`
 	Email         string             `json:"email"`
+	Confirmed     bool               `json:"email_confirmed"`
```
LuccaBitfly, line 6: Minor: IMO struct fields should have the same name as their respective JSON tags
```suggestion
	EmailConfirmed     bool               `json:"email_confirmed"`
```

backend/pkg/api/handlers/common.go
```diff
@@ -158,6 +159,10 @@ func (v *validationError) checkPassword(password string) string {
 	return v.checkRegex(rePassword, password, "password")
 }
 
+func (v *validationError) checkConfirmationHash(hash string) string {
+	return v.checkRegex(reEmailConfirmationHash, hash, "token")
+}
```
LuccaBitfly, line 164: Question: you have made sure that new tokens match this regex?

backend/pkg/api/handlers/auth.go
```diff
@@ -253,6 +289,146 @@ func (h *HandlerService) InternalPostLogout(w http.ResponseWriter, r *http.Reque
...
+}
+
+func (h *HandlerService) InternalPutUserEmail(w http.ResponseWriter, r *http.Request) {
+	// validate user
+	user, err := h.getUserBySession(r)
+	if err != nil {
+		handleErr(w, err)
+		return
+	}
+	userInfo, err := h.dai.GetUserInfo(user.Id)
```
LuccaBitfly, line 316: Should rather call `GetUserCredentialInfo` and add the "EmailConfirmed" field to that struct. Fetching all premium perks, api subscriptions etc. here is maybe overkill.

backend/pkg/api/handlers/auth.go
```diff
@@ -253,6 +289,146 @@ func (h *HandlerService) InternalPostLogout(w http.ResponseWriter, r *http.Reque
...
+	err = h.dai.RemoveUser(user.Id)
+	if err != nil {
+		handleErr(w, err)
+		return
+	}
+
+	returnNoContent(w)
+}
+
+func (h *HandlerService) InternalPutUserEmail(w http.ResponseWriter, r *http.Request) {
```
LuccaBitfly, line 309: Btw, if you add the 409 response codes this should be documented in the API spec 😅

LuccaBitfly (changes requested):
From the business logic side of things everything looks fine, just some issues regarding API consistency.

backend/pkg/api/handlers/auth.go
remoterami: I took this from v1 behaviour: https://github.com/gobitfly/eth2-beaconchain-explorer/blob/master/handlers/user.go#L1300
> We should probably allow changing the email here, e.g. if the user has used a wrong email during signup.

Typically you can't even login to web pages unless your email is confirmed though, right? So I'd rather stick to that, that's also how we did it in v1 (https://github.com/gobitfly/eth2-beaconchain-explorer/blob/master/handlers/auth.go#L278).

That's also not there yet, I'll add it for now; LMK if you think we should deviate from this common/v1 behaviour anyways

backend/pkg/api/handlers/common.go
remoterami: Yes should match, at least unless the generation charset (that function was taken from v1) is changed https://github.com/gobitfly/beaconchain/blob/staging/backend/pkg/commons/utils/user.go#L31

backend/pkg/api/handlers/auth.go
LuccaBitfly: You have convinced me

LuccaBitfly (approved):
lgtm

PR #522, opened 2024-06-26
Bids 2828/rewards chart font weight legend padding

MauserBitfly:
This PR:
- adds some padding to the Rewards Chart Legend
- sets the medium font weight for the x/y axis labels
- tries to fix the font family issue on mac by adding a backup font and default family (if we can't load it from the document) ... this is not tested on a mac and is just a wild guess.

### DIFF
--- a/frontend/components/dashboard/chart/RewardsChart.vue
+++ b/frontend/components/dashboard/chart/RewardsChart.vue
@@ -46,7 +46,7 @@ const { fetch } = useCustomFetch()
 
 const { dashboardKey, isPrivate: groupsEnabled } = useDashboardKey()
 
-const data = ref<ChartData<number, string> | undefined >()
+const data = ref<ChartData<number, string> | undefined>()
 const isLoading = ref(false)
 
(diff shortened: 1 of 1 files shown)


frontend/components/dashboard/chart/RewardsChart.vue
```diff
@@ -205,6 +207,7 @@ const option = computed<ECBasicOption | undefined>(() => {
       silent: true,
       axisLabel: {
         formatter: valueFormatter.value,
+        fontWeight: fontWeightMedium,
         fontSize: textSize,
```
thib-wien, line 211: We have 3 occurences of `fontSize: textSize`
Would it be safer to write `fontSize: textSize + 'px'` ?

MauserBitfly: Fontsize in the ECharts is a number ... so adding 'px' would not be valid
https://echarts.apache.org/en/option.html#textStyle.fontSize

thib-wien (approved):

PR #537, opened 2024-07-01
Add "Bulk Adding" to premium compare

D13ce:
This PR
* Adds a row for "Bulk Adding" to the compare component used on `/pricing`
* Re-factors the compare component slightly


MauserBitfly (commented):
Code changes look fine, but on mobile it is currently really hard to figure out which text - and sub text - belong together and to which row.

image


Guess to problem is that the new text+sub-text take up 3 rows

PR #547, opened 2024-07-02
add index for cached proposal rewards view

peterbitfly:


invis-bitfly (approved):
prob not a large speedup as the multi column indexes should already get used for `dashboard_id` filters but i guess it cant hurt

PR #539, opened 2024-07-02
(BIDS-3186) summary chart v2.5

LuccaBitfly:
This PR does 2 things:
1. Add the new premium perks for the chart.
image
3. Implement the v2.5 version of the summary chart
image

PR #556, opened 2024-07-04
Bids 3186/pricing adaptations

MauserBitfly:
This PR,
based on [PR 539](https://github.com/gobitfly/beaconchain/pull/539)
- adapts the new API Chart Types for the pricing page. 

Note for tester: 
- the new API is not deployed yet so you will not see valid data for the chart props
- the heatmap part is reduced to the label with coming soon (compared to the Figma file) 
- values >= 1 day and < 1 year will be displayed in days (compared to the Figma file)

### DIFF
--- a/frontend/components/pricing/PremiumCompare.vue
+++ b/frontend/components/pricing/PremiumCompare.vue
@@ -16,7 +16,7 @@ type CompareValue = {
   class?: string
 }
 
-type RowType = 'header' | 'group' | 'perc'
+type RowType = 'header' | 'group' | 'perc' | 'label'
 
 type CompareRow = {
   type: RowType,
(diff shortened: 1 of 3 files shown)


D13ce (changes requested):
Almost LGTM.

But: As discussed in DMs, the new rows should not write something like "1 hour" but "last 1 hour":
image

MauserBitfly:
> Almost LGTM.
> 
> But: As discussed in DMs, the new rows should not write something like "1 hour" but "last 1 hour": image

Thanks, should be fixed now (note: also the machine history now labels it with 'Last ...'

D13ce (approved):
LGTM

PR #557, opened 2024-07-04
(BIDS 3195) API: bulk add check premium perks

remoterami:


backend/pkg/api/handlers/internal.go
```diff
@@ -478,6 +478,10 @@ func (h *HandlerService) InternalPostValidatorDashboardValidators(w http.Respons
 		return
 	}
 	limit := userInfo.PremiumPerks.ValidatorsPerDashboard
+	if req.Validators == nil && !userInfo.PremiumPerks.BulkAdding {
+		returnConflict(w, errors.New("bulk adding not allowed with current subscription plan"))
+		return
+	}
```
LuccaBitfly, line 484: We actually want to have the opposite of that. All users should be able to add with the `validators` field, only adding by deposit address, withdrawal credentials or graffiti should be checked with the premium perk.

LuccaBitfly (changes requested):

backend/pkg/api/handlers/internal.go
remoterami: Yes I'm aware, that's what I tried to do, maybe I'm missing something..? If the user supplied `validators`, this if will return false and won't be executed. According to https://github.com/gobitfly/beaconchain/blob/staging/backend/pkg/api/handlers/internal.go#L451, we're doing a bulk add if we're not adding by `validators`, so it must be any of `depositAddress`, `WithdrawalAddress` or `Graffiti`, and we need to check the premium perks

LuccaBitfly: You're right, misread `== nil` as `!= nil`, sorry

PR #565, opened 2024-07-08
(BIDS-3193) API: added pool mode

remoterami:


backend/pkg/api/enums/enums.go
```diff
@@ -608,3 +608,32 @@ var ChartAggregations = struct {
 	IntervalDaily,
 	IntervalWeekly,
 }
+
+// Protocol Modes
```
LuccaBitfly, line 612: I'm wondering whether making an enum for this still makes sense, since we're not passing a single (out of multiple possible) states, but rather a list of enabled states, which can each individually be active or disabled. What do you think of passing some struct which has a boolean field for each protocol instead of an enum?

LuccaBitfly (changes requested):

backend/pkg/api/enums/enums.go
remoterami: Usually I'm a fan of using enums wherever possible, mostly because of their explicitness and extensibility. And even though I could imagine introducing "sub-modes" later (e.g. to present "what could I earn if I did xyz"-data, split `rocket_pool` mode into "off", "show 'actual' protocol data" (=on), "show validators as if they all were LEB8s", "show if they were all LEB16s"), that would prob get too messy in golang once again.
Also I imagine data access could be happier with checking some bools instead of extracting enums from an unsorted list.

So in summary I don't really mind here, will give it a go :+1:

LuccaBitfly (approved):

LuccaBitfly:
lgtm!

PR #567, opened 2024-07-08
Fix overflowing text in `SlotVizToolTip`

marcel-bitfly:
See: BIDS-3203

### DIFF
--- a/frontend/components/slot/viz/SlotVizTooltip.vue
+++ b/frontend/components/slot/viz/SlotVizTooltip.vue
@@ -300,6 +300,16 @@ const data = computed(() => {
         align-items: flex-start;
         word-wrap: nowrap;
         white-space: nowrap;
+        .validators {
+          display: grid;
+          grid-template-columns: 1fr 1fr 1fr;
+          gap: var(--padding-tiny);
+          // otherwise the text: `...and XX more` will mess up the first column
+          > span:not(:has(> a)) {
+              grid-column: span 2;
+              text-align: left
+          }
+        }
       }
     }
   }


frontend/components/slot/viz/SlotVizTooltip.vue
```diff
@@ -300,6 +300,15 @@ const data = computed(() => {
...
         white-space: nowrap;
+        .validators {
+          display: grid;
+          grid-template-columns: 1fr 1fr 1fr;
+          gap: var(--padding-tiny);
+          span:last-child {
+            grid-column: span 3;
+            text-align: left
+          }
+        }
```
D13ce, line 311: **Bug:** This change causes exactly two validators to be written in a column.

**Current staging**
image

**Your branch**
image

There is no need for two validators to be placed below each other. If there are only two validators, it should look like current staging.

It also looks odd when there are 5 validators for example:
image

D13ce (changes requested):

frontend/components/slot/viz/SlotVizTooltip.vue
D13ce: @marcel-bitfly , I think we should not use a grid for this. Here is a screenshot showing a tooltip using your latest commit (a26ec4b5b0a44a053a1381a95e9d23f1f8117186):
image
As you can see, the lower validator line is not aligned correctly, there is an empty space at the beginning.

Shouldn't it suffice to simply allow wrapping for this element? I made a quick test with this code:
```css
.validators {
  text-align: left;
  white-space: wrap;
}
```
and it looks good. Is there anything against that?

marcel-bitfly: it does not work for popups with containing `...and XX more`

But it could stay as is:
https://bitfly1.atlassian.net/browse/BIDS-3203?focusedCommentId=16465

D13ce (approved):
As discussed, the outcome was approved by our designer so:

LGTM

PR #568, opened 2024-07-08
(BIDS-3180) small visual fixes in the group and validator modals

thib-wien:
This PR:

- fixes the column title "Group Id" that was wrapped over two lines on very small screens:
Screenshot from 2024-07-08 16-12-13

- on mobile again, corrects the position of the gem and the Close button at the bottom of tables. Before, they were stacked vertically instead of spread left and right. Now:

image


MauserBitfly (approved):
LGTM

PR #570, opened 2024-07-09
Nobids/fix rewards view el and use goqu

Eisei24:
This fixes the el_reward calculations. It would have resulted in wrong data for everything except the el_reward if a validator proposed more than one block within an epoch.

Furthermore changed from string manipulation to goqu for query creation.

### DIFF
--- a/backend/pkg/api/data_access/vdb_rewards.go
+++ b/backend/pkg/api/data_access/vdb_rewards.go
@@ -878,84 +878,9 @@ func (d *DataAccessService) GetValidatorDashboardDuties(ctx context.Context, das
 		}
 	}
 
-	// Use a subquery to allow access to total_reward in the where clause
-	// rewardsQuery := fmt.Sprintf(`
-	// 	SELECT *
-	// 		FROM (
-	// 		WITH rewards AS(
-	// 			SELECT
-	// 				e.validator_index,
-	// 				(CAST((
-	// 					COALESCE(e.attestations_reward, 0) +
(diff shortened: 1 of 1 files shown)


backend/pkg/api/data_access/vdb_rewards.go
```diff
@@ -71,63 +74,41 @@ func (d *DataAccessService) GetValidatorDashboardRewards(ctx context.Context, da
...
-					queryParams = append(queryParams, currentCursor.Epoch)
-					whereQuery += fmt.Sprintf(" AND e.epoch%[1]s$%[2]d", sortSearchDirection, len(queryParams))
+					condition = fmt.Sprintf("e.epoch %s ?", sortSearchDirection)
 				} else {
 					// The cursor is on a specific group, get the data for the whole epoch since we could need it for the total rewards
-					queryParams = append(queryParams, currentCursor.Epoch)
-					whereQuery += fmt.Sprintf(" AND e.epoch%[1]s=$%[2]d", sortSearchDirection, len(queryParams))
+					condition = fmt.Sprintf("e.epoch %s= ?", sortSearchDirection)
 				}
+				ds = ds.Where(goqu.L(condition, currentCursor.Epoch))
```
peterbitfly, line 102: maybe set the `ds.Where` directly in the if branches, skipping the condition variable?

peterbitfly (commented):
still quite complex but goqu makes it more readable from my point of view

PR #573, opened 2024-07-10
(BIDS-3209) API: notification endpoints

LuccaBitfly:


frontend/types/api/notifications.ts
```diff
@@ -0,0 +1,176 @@
...
+  realtime_mode: boolean;
+}
+export interface NotificationSettingsValidatorDashboard {
+  webhook_url: string;
+  webhook_discord: boolean;
+  real_time_mode: boolean;
+  subscribed_events: NotificationEventsValidatorDashboard;
+}
+export type InternalPutNotificationSettingsValidatorDashboardResponse = ApiDataResponse<NotificationSettingsValidatorDashboard>;
+export interface NotificationEventsAccountDashboard {
```
thib-wien, line 151: Is this the part that tells us how to send the API the preferences of the user in this dialog?
image
The user can select networks and can ask us to ignore spam transactions.
I do not see those fields in this json :)

frontend/types/api/notifications.ts
```diff
@@ -0,0 +1,176 @@
...
+  incoming_transactions: boolean;
+  outgoing_transactions: boolean;
+  track_erc20_token_transfers: boolean;
```
thib-wien, line 154: I think here we need a float because the user types a  value in dollars.

thib-wien:
PS: I adopted this convention for numbers (the percentage of offline validators and the threshold for ERC20 transaction) :
- positive number: the user entered a number and activated the setting with the checkbox
- negative number: the user entered a number and deactivated the setting with the checkbox
- NaN : the user has left the input field empty

frontend/types/api/notifications.ts
LuccaBitfly: Network selection is in `NotificationSettingsAccountDashboard`, this type is supposed to only contain subscribable events. But @MauserBitfly would prefer flat types, so I will add all fields from the events types to their respective settings type.

frontend/types/api/notifications.ts
```diff
@@ -0,0 +1,176 @@
...
+  enable_email: boolean;
+  enable_push: boolean;
+  subscribed_events: NotificationEventsGeneral;
+  networks: NotificationSettingsNetwork[];
+  paired_devices: NotificationSettingsPairedDevice[];
+}
+export type InternalGetNotificationSettingsResponse = ApiDataResponse<NotificationSettings>;
+export interface NotificationEventsValidatorDashboard {
+  validator_offline: boolean;
+  group_offline: number /* float64 */;
```
thib-wien, line 135: As JSON does not support `NaN`, can you make tygo generate `number|null` here?
`null` is when the user hasn't set a value (the user never opened the setting dialog or has left the input field empty)

frontend/types/api/notifications.ts
```diff
@@ -149,6 +149,7 @@ export interface NotificationSettingsAccountDashboard {
   subscribed_chain_ids: number /* uint64 */[];
   is_incoming_transactions_subscribed: boolean;
   is_outgoing_transactions_subscribed: boolean;
+  is_erc20_token_transfers_subscribed: number /* float64 */;
```
thib-wien, line 152: Maybe here we want a boolean (the next member contains the amount).
Or this member disappears and we use `< 0` to disable

LuccaBitfly: Yes it should be a boolean, thx

backend/pkg/api/handlers/notifications.go
```diff
@@ -0,0 +1,429 @@
...
+	"errors"
+	"math"
+	"net/http"
+
+	"github.com/gobitfly/beaconchain/pkg/api/enums"
+	"github.com/gobitfly/beaconchain/pkg/api/types"
+	"github.com/gorilla/mux"
+)
+
+// TODO move to internal.go before merging
```
remoterami, line 13: Don't forget

backend/pkg/api/handlers/notifications.go
```diff
@@ -0,0 +1,429 @@
...
+		handleErr(w, err)
+		return
+	}
+	checkMinMax(&v, req.MachineStorageUsageThreshold, 0, 1, "machine_storage_usage_threshold")
+	checkMinMax(&v, req.MachineCpuUsageThreshold, 0, 1, "machine_cpu_usage_threshold")
+	checkMinMax(&v, req.MachineMemoryUsageThreshold, 0, 1, "machine_memory_usage_threshold")
+	checkMinMax(&v, req.RocketPoolMaxCollateralThreshold, 0, 1, "rocket_pool_max_collateral_threshold")
+	checkMinMax(&v, req.RocketPoolMinCollateralThreshold, 0, 1, "rocket_pool_min_collateral_threshold")
+	// TODO: check validity of clients
+	err := h.dai.UpdateNotificationSettingsGeneral(r.Context(), userId, req)
```
remoterami, line 234: Add `v.hasErrors()` check

backend/pkg/api/handlers/notifications.go
```diff
@@ -0,0 +1,429 @@
...
+		Name                   string `json:"name,omitempty"`
+		IsNotificationsEnabled bool   `json:"is_notifications_enabled"`
+	}{}
+	if err := v.checkBody(&req, r); err != nil {
+		handleErr(w, err)
+		return
+	}
+	// TODO use a better way to validate the paired device id
+	pairedDeviceId := v.checkRegex(reNonEmpty, mux.Vars(r)["paired_device_id"], "paired_device_id")
+	v.checkNameNotEmpty(req.Name)
```
remoterami, line 290: Consider using return value of `checkNameNotEmpty` as name

Also add `v.hasErrors()` check

backend/pkg/api/handlers/notifications.go
```diff
@@ -0,0 +1,429 @@
...
+		},
+	}
+
+	returnOk(w, response)
+}
+
+func (h *HandlerService) InternalDeleteUserNotificationSettingsPairedDevices(w http.ResponseWriter, r *http.Request) {
+	var v validationError
+	// TODO use a better way to validate the paired device id
+	pairedDeviceId := v.checkRegex(reNonEmpty, mux.Vars(r)["paired_device_id"], "paired_device_id")
```
remoterami, line 311: `v.hasErrors()` check again - or did you leave them out intentionally?

remoterami (changes requested):

backend/pkg/api/handlers/notifications.go
LuccaBitfly: No, it was unintentional 😅

backend/pkg/api/handlers/notifications.go
LuccaBitfly: Will move after approval

remoterami (approved):
lgtm

PR #574, opened 2024-07-10
Fix `clickable area` of `dashboard header buttons`

marcel-bitfly:
This is a hack. Ideally buttons should be constructed in the correct format.
For example the text could define the `button size`, by having its own padding.
Preventing this issue.

See: BIDS-3200

### DIFF
--- a/frontend/assets/css/prime_menubar.scss
+++ b/frontend/assets/css/prime_menubar.scss
@@ -15,17 +15,25 @@
       border: 1px solid var(--container-border-color);
       border-radius: var(--border-radius);
       height: 30px;
-      cursor: pointer;
       &:has(.router-link-active),
       &:has(.p-active) {
         color: var(--primary-color);
         border-color: var(--primary-color);
-      }
-
+      }     
       > .p-menuitem-content {
         padding: var(--padding-small) var(--padding-large);
         border-radius: var(--border-radius);
         height: 28px;
+
+        .text {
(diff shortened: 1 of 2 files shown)


frontend/assets/css/prime_menubar.scss
```diff
@@ -26,6 +26,14 @@
...
         border-radius: var(--border-radius);
         height: 28px;
+        .text {
+          // this hack is necessary to make the clickable area as big as the parent
+          &::before{
+            content: "";
+            position: absolute;
+            inset: 0;
+          }
+        }
```
D13ce, line 36: **Suggestion:** You might want to do the same for this scenario:
image
i.e. when the drop-down menu is shown. The grey hover area OUTSIDE of the red rectangle shows a pointer but is not clickable.

For this scenario to happen, you have to be logged in and you have to have at least 2 validator dashboards.

D13ce (commented):
The issue mentioned in the ticket is fixed, but there is pretty much the same issue for users that have more than one dashboard causing a drop-down. Please see my comment below :point_down: 

My suggestion would be to fix this other issue in this PR too. If you want to have a new ticket, I can create one of course. You choose :)

thib-wien:
On request of Marcel, I reviewed his change following the observation of @D13ce.
LG**TM**.

thib-wien (approved):

PR #577, opened 2024-07-11
Bids 3162/summary chart v2 5

MauserBitfly:
This PR:
Implements the Summary 2.5 chart changes
- New Filter options
  - aggregation (some options only available for permium users charts *)
  - efficiency
  - groups: Multi select + 2 special options (network average and total)
 - data points now based on timestamp an no longer on epochs

This PR did not
- Test the chart with new data (the backend does not throw an error with the new options, but still returns the old data) -> currently the dates / epochs in the chart are wrong displayed as they are based on timestamps but get epochs from the backend. 
- Remove the `showInDevelopment` flag as the Chart should only be made public after tests against the new backend. 
- Implement loading of partial time frame loading. We will first see how the chart performs when loading all the data and only when needed implement partial loading. As this would decrease usability and should only be done when we get into performance issues. 
  
  
*) add. info: the permissions for the chart depend on the user how created the dashboard and not on the logged-in user - important for shared dashboards.

### DIFF
--- a/frontend/components/dashboard/chart/DashboardChartSummaryChartFilter.vue
+++ b/frontend/components/dashboard/chart/DashboardChartSummaryChartFilter.vue
@@ -30,8 +30,8 @@ const efficiencyList = EfficiencyTypes.map(e => ({
 watch(efficiency, (e) => { chartFilter.value.efficiency = e })
 
 /** groups */
-const total = ref(chartFilter.value.groupIds.includes(SUMMARY_CHART_GROUP_TOTAL))
-const average = ref(chartFilter.value.groupIds.includes(SUMMARY_CHART_GROUP_NETWORK_AVERAGE))
+const total = ref(!chartFilter.value.initialised || chartFilter.value.groupIds.includes(SUMMARY_CHART_GROUP_TOTAL))
+const average = ref(!chartFilter.value.initialised || chartFilter.value.groupIds.includes(SUMMARY_CHART_GROUP_NETWORK_AVERAGE))
 const groups = computed(() => {
   if (!overview.value?.groups) {
     return []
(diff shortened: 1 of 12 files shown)


frontend/components/dashboard/chart/SummaryChart.vue
D13ce, line 1: **Suggestion:** This is not related to this PR, but the tooltip should probably only show two decimal points, right?
image

frontend/components/dashboard/chart/DashboardChartSummaryChartFilter.vue
D13ce, line 1: **Suggestion:** Should we maybe debounce this? So if a user changes the group settings quickly, there are not that many requests.

D13ce (changes requested):
Well done. Only found minor issues, PTAL below :point_down:

frontend/components/dashboard/chart/SummaryChart.vue
```diff
@@ -109,17 +129,16 @@ const option = computed(() => {
         fontSize: textSize,
         lineHeight: 20,
         formatter: (value: number) => {
-          const date = formatEpochToDate(value, $t('locales.date'))
-          if (date === undefined) {
-            return ''
+          const date = formatGoTimestamp(value, undefined, 'absolute', 'narrow', $t('locales.date'), false)
+          if (aggregation.value === 'epoch') {
+            return `${date}\n${$t('common.epoch')} ${tsToEpoch(value)}`
           }
```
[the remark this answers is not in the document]
MauserBitfly: It is a Minus sign due to the data beeing epoch and not a timestamp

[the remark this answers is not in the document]
D13ce: Ah, of course. Thanks for the clarification, resolved.

frontend/components/dashboard/chart/SummaryChart.vue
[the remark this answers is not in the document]
MauserBitfly: should be fixed now

MauserBitfly: Had the same thought, but then got distracted in the office and forgot about it ... changed ;-)

frontend/components/dashboard/chart/DashboardChartSummaryChartFilter.vue
MauserBitfly: good point ... sending now a request on the first change (so if he only chances one item it is fast) and then debounce if he makes multiply changes

D13ce (approved):
LGTM

PR #581, opened 2024-07-11
align arrow icon for tooltips on the border of the screen

MauserBitfly:
This PR:

- Tries to align the tooltip arrow icon with it's parent as good as possible  - like on the borders of the browser.


thib-wien (approved):
Nice improvement

PR #583, opened 2024-07-14
Nobids/use context in db queries

peterbitfly:
Switches all db calls in the `data_access` package to use the provided `context.Context`


remoterami (approved):
Please remove https://github.com/gobitfly/beaconchain/blob/staging/backend/.golangci.yml#L190-L194 for good measure

Otherwise, lgtm!

PR #585, opened 2024-07-15
Bids 3218 token header to ssr requests

benji-bitfly:
This PR introduces a conditional header addition for server-side requests in the useCustomFetch function.


frontend/composables/useCustomFetch.ts
```diff
@@ -38,6 +38,7 @@ export function useCustomFetch () {
     const { public: { apiClient, legacyApiClient, apiKey, domain, logIp }, private: pConfig } = runtimeConfig
     const path = map.mock ? `${pathName}.json` : map.getPath?.(pathValues) || map.path
     let baseURL = map.mock ? '../mock' : map.legacy ? legacyApiClient : apiClient
+    const ssrSecret = process.env.NUXT_PRIVATE_SSR_SECRET
```
peterbitfly, line 41: i believe this approach will only allow us to specify the secret as runtime environment variables, but we need to be able to specifiy it both during runtime (for our own deployments) and during compile time (for deployments via cloudflare pages)

peterbitfly (changes requested):

frontend/composables/useCustomFetch.ts
benji-bitfly: Understood.

frontend/composables/useCustomFetch.ts
```diff
@@ -38,6 +38,7 @@ export function useCustomFetch () {
     const { public: { apiClient, legacyApiClient, apiKey, domain, logIp }, private: pConfig } = runtimeConfig
     const path = map.mock ? `${pathName}.json` : map.getPath?.(pathValues) || map.path
     let baseURL = map.mock ? '../mock' : map.legacy ? legacyApiClient : apiClient
+    const ssrSecret = pConfig?.ssrSecret
```
marcel-bitfly, line 41: nitpick: all `runtimeConfigs` that are not part of `public` are `private`, so this could be simplified
```suggestion
    const { public: { apiClient, legacyApiClient, apiKey, domain, logIp }, ssrSecret } = runtimeConfig
    const path = map.mock ? `${pathName}.json` : map.getPath?.(pathValues) || map.path
    let baseURL = map.mock ? '../mock' : map.legacy ? legacyApiClient : apiClient
```

frontend/nuxt.config.ts
```diff
@@ -46,7 +46,8 @@ export default defineNuxtConfig({
     },
     private: {
       apiServer: process.env.PRIVATE_API_SERVER,
-      legacyApiServer: process.env.PRIVATE_LEGACY_API_SERVER
+      legacyApiServer: process.env.PRIVATE_LEGACY_API_SERVER,
+      ssrSecret: process.env.PRIVATE_SSR_SECRET || ''
```
marcel-bitfly, line 50: nitpick:
```suggestion
      ssrSecret: ''
```
Except `ssrSecret` should be `undefined` on purpose

marcel-bitfly (approved):

marcel-bitfly:
Nitpick your `commit messages` could be more consistent or even be squashed and give more helpful context information for `git line blaming`:

e.g. See: https://www.gitkraken.com/learn/git/best-practices/git-commit-message#using-imperative-verb-form

peterbitfly (approved):

PR #588, opened 2024-07-16
(BIDS-3032) Dashboard creation flow depends on parameters passed by the caller

thib-wien:
This PR adds two _optional_ parameters to exposed function `.show()` of _DashboardCreationController.vue_.

Those parameters allow the parent of the dashboard creation flow to force the creation of either an account or a validator dashboard, and on a certain network only.
It does not change anything for the parts of the explorer already using the dashboard creation flow because the parameters are optional. They will be used by future pages to come.

This can be tried on the playground.


MauserBitfly (approved):
LGTM.

the new forced dashboard type only works in combination with mode `modal` and not with `panel` + ìnitiallyVisible`  ... but I think that's fine as we don't have a usecase where we need to force the type in the 'panel' mode.

PR #613, opened 2024-07-23
fix mobile summary detail scroll, hide chart in non dev mode

MauserBitfly:
This PR
- hides the chart in the summary table in non prod
- fixes the scroll issue in the summary details table on mobile if there are multiple validators.


thib-wien (approved):
✅

PR #618, opened 2024-07-24
optimize the switch from a shared dashboard to /dashboard

MauserBitfly:
This PR:
- makes sure that we don't set the shared dashboard key in the cookie dashboard
- fixes the issue in the Validator Summary tables when you switch from a shared dashboard to an empty dashboard .*
- fixes an error in the html utils where we check for an elements parent that sometimes cause an error in the console

.* to recreate the problem: Be not logged in. Create an empty dashboard (don't add validators). Open a shared dashboard and then click the 'go to dashboard' button in the bottom. Result before the bugfix: the shard key was put in the dashboard cookie and the summary table still showed the shared dashboard data. And sometimes this was leading to the browser to be frozen.

### DIFF
--- a/frontend/stores/dashboard/useUserDashboardStore.ts
+++ b/frontend/stores/dashboard/useUserDashboardStore.ts
@@ -120,7 +120,7 @@ export function useUserDashboardStore () {
   // Update the hash (=hashed list of id's) of a specific local dashboard
   function updateHash (type: DashboardType, hash: string) {
     if (!isPublicDashboardKey(hash) || isSharedKey(hash)) {
-      warn('updateHash: invalid public hashed key: ', hash)
+      warn('invalid public hashed key: ', hash)
       return
     }
     if (type === 'validator') {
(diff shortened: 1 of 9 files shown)


frontend/stores/dashboard/useUserDashboardStore.ts
```diff
@@ -118,6 +119,10 @@ export function useUserDashboardStore () {
 
   // Update the hash (=hashed list of id's) of a specific local dashboard
   function updateHash (type: DashboardType, hash: string) {
+    if (!isPublicDashboardKey(hash) || isSharedKey(hash)) {
+      warn('updateHash: invalid public hashed key: ', hash)
```
marcel-bitfly, line 123: nitpick: call stack should appear in `vue warn` anyway: but on the other hand I get that it is more convenient to read
```suggestion
      warn('invalid public hashed key: ', hash)
```

marcel-bitfly (approved):

frontend/stores/dashboard/useUserDashboardStore.ts
MauserBitfly: sure ;-)

PR #624, opened 2024-07-24
(BIDS-3229) Shows the currency selector in the header (rather than burger menu) if there is enough room

thib-wien:
To try it:
1. Use a mobile view,
2. Log out. The login/signup buttons take all the room in the header so the currency selector is in the burger menu.
3. Log in. The currency selector is now in the header because there is room for it.


benji-bitfly:
LGTM :+1:

PR #626, opened 2024-07-24
(BIDS-3044) Slotviz tooltips: slot numbers open slot pages -- version B

thib-wien:
This is a proposal.

In this version, the changes require 7 additional lines of code and the code is more complicated.

### DIFF
--- a/frontend/components/slot/viz/SlotVizTooltip.vue
+++ b/frontend/components/slot/viz/SlotVizTooltip.vue
@@ -18,7 +18,7 @@ const data = computed(() => {
 
   const status = slot.status === 'scheduled' && slot.slot < (props.currentSlotId ?? 0) ? 'scheduled-past' : slot.status
 
-  const networkLabel = $t(`slotViz.tooltip.network.${status}`, { slot: formatNumber(slot.slot) })
+  const networkLabel = tAll($t, `slotViz.tooltip.network.${status}`)
 
   const hasDuties = !!slot?.proposal || !!slot?.slashing || !!slot?.attestations || !!slot?.sync
   let hasSuccessDuties = false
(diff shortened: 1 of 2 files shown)


frontend/components/slot/viz/SlotVizTooltip.vue
```diff
@@ -164,7 +164,14 @@ const data = computed(() => {
...
           <div class="row network">
-            <BcFormatNumber :text="data.networkLabel" />
+            {{ data.networkLabel[0] }}&nbsp;
+            <BcFormatNumber v-if="props.data.status === 'scheduled'" :text="formatNumber(props.data.slot)" />
+            <BcLink v-else :to="`/slot/${props.data.slot}`" target="_blank" class="link">
+              <BcFormatNumber :text="formatNumber(props.data.slot)" />
+            </BcLink>
+            <span v-if="data.networkLabel[1]">
+              &nbsp;{{ data.networkLabel[1] }}
+            </span>
```
marcel-bitfly, line 174: if we stick to the default solution of our [package](https://vue-i18n.intlify.dev/guide/advanced/component.html#slots-syntax-usage) it could be simplified to:
```suggestion
            <i18n-t :keypath="`slotViz.tooltip.network.${props.data.status}`" tag="span">
              <template #slot>
                <BcLink :to="`/slot/${props.data.slot}`" target="_blank" class="link">
                  <BcFormatNumber :text="`${props.data.slot}`" />
                </BcLink>
              </template>
            </i18n-t>
```
you could think about wrapping `<i18n-t>` into a `BcComponent` as it is `third party code`

marcel-bitfly:
your solution worked fine, smart idea with the `array syntax` :muscle: 
But imo it would be best to stick to the `default solution` of the `packages and frameworks` we are using. 

In this case it would be [component interpolation](https://vue-i18n.intlify.dev/guide/advanced/component.html) for `translations`

frontend/components/slot/viz/SlotVizTooltip.vue
thib-wien: I do not understand you code.

What does _en.json_ would look like?
How does `i18n-t` know what part is outside `<bcLink></BcLink>` and what part is inside?
Why does `formatNumber(props.data.slot)` disappear?

marcel-bitfly: - plz revert `en.json`
- try the code I suggested, it should work fine

for reference: https://vue-i18n.intlify.dev/guide/advanced/component.html

thib-wien: Thanks for your proposal of a 3rd way to implement the BIDS.
I implemented it after correcting two details:
- `:keypath="'slotViz.tooltip.network.${props.data.status}'"` would lead to bugs because the status in the props is not exactly the status that _en.json_ needs (see [this line](https://github.com/gobitfly/beaconchain/blob/243d90cd0c281a38cf242c2cb214647f5de2c041/frontend/components/slot/viz/SlotVizTooltip.vue#L18))
- `<BcFormatNumber :text="'${props.data.slot}'"/>` would not format the number as we want (with separation between thousands), what we need is `:value="props.data.slot"`

marcel-bitfly (approved):
nice work :+1:

PR #630, opened 2024-07-25
Fix `DashboardOverviewBox` does not show data

marcel-bitfly:
Request  to `DASHBOARD_VALIDATOR_INDICES` needed  `period query parameter`. 
For now it was `hard coded` to `last_24h`.

See: BIDS-3227

### DIFF
--- a/frontend/components/dashboard/OverviewBox.vue
+++ b/frontend/components/dashboard/OverviewBox.vue
@@ -22,7 +22,8 @@ const openValidatorModal = () => {
     data: {
       context: 'dashboard',
       dashboardName: getDashboardLabel(dashboardKey.value, 'validator'),
-      dashboardKey: dashboardKey.value
+      dashboardKey: dashboardKey.value,
+      timeFrame: 'last_24h'
     }
   })
 }


frontend/components/dashboard/OverviewBox.vue
```diff
@@ -22,7 +22,8 @@ const openValidatorModal = () => {
     data: {
       context: 'dashboard',
       dashboardName: getDashboardLabel(dashboardKey.value, 'validator'),
-      dashboardKey: dashboardKey.value
+      dashboardKey: dashboardKey.value,
+      timeFrame: 'last_24h'
```
marcel-bitfly, line 26: @LuccaBitfly could the backend be adapted not to require a `period parameter`

image
image

LuccaBitfly: @marcel-bitfly it used to be like that, but at some point someone requested that the API should return an error if this parameter is omitted. Can revert that again if you want to. pls lmk

marcel-bitfly: @MauserBitfly suggested to leave it like that, as we might break other things.  Thx for the information

MauserBitfly (approved):
LGTM

PR #631, opened 2024-07-25
Bids 3230/summary chart improvements

MauserBitfly:
This PR:
- updates the styles of the tooltips (they are now different in light and dark mode)
- makes the summary chart filter aggregation more reactive
- refactors the summary chart 
  - we should now initially trigger only one load instead of two
  - loading animation is now above the chart instead of replacing the chart while loading 
  - we are now loading only the data for the selected range (instead of the whole data at once)
  - the selected range is now limited to 200 data points
  - lines are no longer smoothed 
  - highlight the data entry of the line closed to the mouse in the tooltip
  - fix date format of the range slider
  - clear table if we get an error 
  - depending on which input changes we update the chart instantly or bounce (like when the user moves around the range slider) 
  - re-enabling the summary chart in prod mode


frontend/components/dashboard/chart/SummaryChart.vue
```diff
@@ -221,14 +323,140 @@ const option = computed(() => {
...
+    ...zoomValues,
+    toIndex,
+    toTs: categories.value[toIndex],
+    fromIndex,
+    fromTs: categories.value[fromIndex]
+  }
+}
+
+// validate and adjust zoom settings
+const validateDataZoom = (instant?: boolean) => {
```
thib-wien, line 357: When I go too far in the past, the chart becomes empty and I think that that it would confuse users:
image
Maybe we want to show an error tooltip or a dialog inviting to subscribe?

frontend/components/dashboard/chart/SummaryChart.vue
```diff
@@ -155,7 +247,7 @@ const option = computed(() => {
       type: 'value',
       minInterval: 10,
       maxInterval: 20,
-      min: (range: any) => Math.max(0, 10 * Math.ceil(range.min / 10 - 1)),
```
thib-wien, line 158: Without this lower bound at 0, the scale of the chart shows "-10%" when a validator has been offline, for example:
image

thib-wien:
This is not related to the PR but I believe that there is a little bug in the routing.

9 times out of 10 with Gnome Web and maybe 1 time out of 3 with Firefox, the browser shows an empty dashboard when I enter this URL:
https://local.beaconcha.in:3000/dashboard/v-80d7edaa-74fb-4129-a41e-7700756961cf
(when it happens, first I see the dashboard creation dialog for a few seconds and the browser redirects automatically to an empty dashboard)

thib-wien (commented):
:)

MauserBitfly:
> This is not related to the PR but I believe that there is a little bug in the routing.
> 
> 9 times out of 10 with Gnome Web and maybe 1 time out of 3 with Firefox, the frontend does not show the correct dashboard when I enter this URL: https://local.beaconcha.in:3000/dashboard/v-80d7edaa-74fb-4129-a41e-7700756961cf (when it happens, first I see the dashboard creation dialog for a few seconds and the browser redirects automatically to an empty dashboard)

switching to your own dashboard (from a shared dashboard) should happen when the shared dashboard cannot be loaded (because maybe it was deleted or something) ... so maybe the internet connection is not stable ... it's also possible that there is a bug, so let's keep en eye on it (but not within this PR)

frontend/components/dashboard/chart/SummaryChart.vue
```diff
@@ -247,7 +247,7 @@ const option = computed(() => {
       type: 'value',
       minInterval: 10,
       maxInterval: 20,
-      min: (range: any) => 10 * Math.ceil(range.min / 10 - 1),
+      min: (range: any) => range.min >= 0 ? Math.max(0, 10 * Math.ceil(range.min / 10 - 1)) : 10 * Math.ceil(range.min / 10 - 1),
```
thib-wien, line 250: I think that the minimum value is never negative in our use case, so maybe the test and the second part are not useful here. Maybe just `Math.max(0, 10 * Math.ceil(range.min / 10 - 1))` is enough in our case.

frontend/components/bc/BcTooltip.vue
```diff
@@ -332,7 +332,7 @@ onUnmounted(() => {
     opacity: 1;
```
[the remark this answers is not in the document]
thib-wien: Did I misunderstand the need for `opacity`?

thib-wien (approved):

PR #636, opened 2024-07-29
fix chart ts

invis-bitfly:
group by timestamp column instead of epoch_start to accurately dedupplicate fragments in clickhouse data. this makes it so the function actually returns the intended count of data points
add function to get last aggregation ts for each table


LuccaBitfly (approved):
lgtm

PR #638, opened 2024-07-29
(BIDS-3250) add chart datapoint limits

LuccaBitfly:
for the summary chart we will have two limits enforced by the api:

- the dashboard owner's premium perk, which, for each aggregation type, decides the earliest possible datapoint that could be returned. if the client requests data that goes further back than this, the api returns an error
- a general limit of 200 datapoints, which limits how far after_ts and before_ts can lie apart. if the client requests data that would exceed this, the api returns an error.


backend/pkg/api/handlers/internal.go
```diff
@@ -966,24 +967,26 @@ func (h *HandlerService) InternalGetValidatorDashboardSummaryChart(w http.Respon
...
+	if err != nil {
+		handleErr(w, err)
 		return
 	}
-	if maxAge == 0 {
-		returnConflict(w, fmt.Errorf("requested aggregation is not available for dashboard owner's premium subscription"))
+	minAllowedTs := latestExportedTs - min(maxAge, latestExportedTs)                                      // min to prevent underflow
+	secondsPerEpoch := uint64(12 * 32)                                                                    // TODO: fetch dashboards chain id and use correct value for network once available
+	maxAllowedInterval := chartDatapointLimit*uint64(aggregation.Duration(secondsPerEpoch).Seconds()) - 1 // -1 to make sure we don't go over the limit
+	afterTs, beforeTs := v.checkTimestamps(q.Get("after_ts"), q.Get("before_ts"), latestExportedTs, minAllowedTs, maxAllowedInterval)
```
LuccaBitfly, line 983: Maybe it's possible to make all this logic for timestamps easier

remoterami: Prob not, but could try to plan for minimizing code duplication with other charts later. I'd probably wrap it all up in a general `chartTimestampHandler`<sup>TM</sup> method (or maybe 2 if we'll have charts without aggregation), and incorporate `checkTimestamps` into that function as I doubt that it will ever be used in another context.

backend/pkg/api/handlers/common.go
```diff
@@ -605,20 +606,39 @@ func isValidNetwork(network intOrString) (uint64, bool) {
...
-	}
-	beforeTs := uint64(time.Now().Unix())
-	if beforeParam != "" {
-		beforeTs = v.checkUint(beforeParam, "before_ts")
-	}
-	if afterTs > beforeTs {
-		v.add("after_ts", "parameter `after_ts` must not be greater than `before_ts`")
+func (v *validationError) checkTimestamps(afterParam string, beforeParam string, latestExportedTs uint64, minAllowedTs uint64, maxAllowedInterval uint64) (after uint64, before uint64) {
+	switch {
+	// If both parameters are e mpty, return the latest data
```
remoterami, line 611:
```suggestion
	// If both parameters are empty, return the latest data
```

remoterami (changes requested):

backend/pkg/api/handlers/internal.go
LuccaBitfly: we'll do that once it happens :)

remoterami (approved):

remoterami:
lgtm

PR #645, opened 2024-07-30
summary chart tooltip: add opacity to non highlighted elements

MauserBitfly:
This PR:
- adds a 0.5 opacity for non highlighted elements in the summary chart toolitp


thib-wien (approved):
perfect

PR #647, opened 2024-07-30
(BIDS-2872) copy ratelimit over from v1

guybrush:


LuccaBitfly (changes requested):
Everything regarding ratelimiting logic seems sound. When reading through the code, one notices how well thought-of all of it is, props! Nonetheless, I have some minor suggestions. Please have a look :)

LuccaBitfly (approved):
utAck

PR #651, opened 2024-07-31
(BIDS-3032) Refactoring to follow incoming new eslint rules - mini bug fix

thib-wien:
When I replaced the weak binding between v-models with a bridge, I forgot to tell the bridge that `0` as a `ChainID` corresponds to `''` as a `string` (they represent the state "no network selected" on the respective v-models).


frontend/components/dashboard/creation/DashboardCreationNetworkMask.vue
```diff
@@ -6,12 +6,14 @@ import { useNetworkStore } from '~/stores/useNetworkStore'
 const { currentNetwork, availableNetworks, isNetworkDisabled } = useNetworkStore()
 const { t: $t } = useI18n()
 
+const emit = defineEmits<{(e: 'next' | 'back'): void }>()
 const network = defineModel<ChainIDs>('network', { required: true })
-const selection = usePrimitiveRefBridge<ChainIDs, `${ChainIDs}`|''>(network)
+const selection = usePrimitiveRefBridge<ChainIDs, `${ChainIDs}`|''>(network, o => `${o || ''}`, c => Number(c || 0))
```
marcel-bitfly, line 11: what is `o =>` for?

thib-wien: In the vocabulary of our bridge system, we have the "original ref" and the "created ref". They correspond to the ref that you give to the bridge and the ref that it creates.

Maybe I should rename `o` and `c` to something like `net` and `sel` to match the names of the v-models in the file?

thib-wien: I renamed. Tell me if the names are still not clear enough.

marcel-bitfly: imo best would be to write it out completely

thib-wien: I disagree, they would have the same names as the global varianbles although they are local. Confusing practice.

marcel-bitfly: in this case I would suggest `_network` and `_selection`?

MauserBitfly: I think we should reserve starting with `_` only for variable that are not used.  (Example: `..forEach((_value,index)=>console.log('i am only interested in the', index))`

`net` and `sel`should be sufficient here to get the meaning out of the context in that line.

marcel-bitfly (approved):

PR #654, opened 2024-07-31
Bids 3232/update validator subset states

MauserBitfly:
This PR:
- fixes some issues in the Validator Subset Modal
   - updates translations for the missing categories
   - tries to get the count information out of the group data (defaulting to the row data that was used before) 
   - using the new count logic for the dialog sub header and the accordion header
   - splits up the `withrawing` / `withdrawn` categories into `exited` and `slashed`
   - we no longer open the first group
   - only adds the 'All' category if we have more then one group and are in the dashboard/group context


frontend/i18n/en.json
```diff
@@ -580,9 +580,15 @@
...
           "proposal_missed": "Missed",
           "exiting": "Exiting",
-          "exited":"Exited",
-          "withdrawing":"Withdrawing",
-          "withdrawn":"Withdrawn"
+          "exited": "Exited",
+          "withdrawing": "Withdrawing",
+          "withdrawn": "Withdrawn",
+          "slashing": "Slashing",
+          "slashed": "Slashed",
```
thib-wien, line 587: The translation seems present but I got this:
Screenshot from 2024-08-02 11-41-25
🤔

frontend/i18n/en.json
```diff
@@ -769,7 +775,12 @@
     "password": "Password",
     "choose_password": "Choose a password",
     "confirm_password": "Confirm your password",
-    "please_agree": ["I agree to the", "Terms and Conditions", "and", "Privacy Policy"],
+    "please_agree": [
+      "I agree to the",
+      "Terms and Conditions",
```
thib-wien, line 780: This is outdated. `staging` has different words (for example, "Term and Conditions" is now "T&C").

I guess you will get the correct strings after updating your branch from `staging`.

I suggest that you update from staging **before you solve the conflicts** otherwise you might erase the new strings.

thib-wien (commented):
With apologies for the delay, here is my review.

I think that the branch has conflicts.
The problem is that I see at least one change that you made starting from an outdated version of staging. If you solve the conflicts before updating from staging, I am afraid that you will erase the new version of these lines that is on staging.

frontend/i18n/en.json
MauserBitfly: Strange, works for my ... are you on my branch?
image

frontend/utils/dashboard/validator.ts
```diff
@@ -25,3 +26,27 @@ export function sortSummaryValidators (list?: VDBSummaryValidator[]): VDBSummary
   }
   return [...list].sort((a, b) => a.index - b.index)
 }
+
```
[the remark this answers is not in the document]
MauserBitfly: Sure you are on my branch?

thib-wien (approved):
Let's Go To Merge

PR #657, opened 2024-08-01
Bids 3236/ch for summary

Eisei24:


backend/pkg/api/data_access/vdb_summary.go
```diff
@@ -732,62 +738,138 @@ func (d *DataAccessService) GetValidatorDashboardGroupSummary(ctx context.Contex
 	return ret, nil
 }
 
-func (d *DataAccessService) internal_getElClAPR(ctx context.Context, validators []t.VDBValidator, days int) (elIncome decimal.Decimal, elAPR float64, clIncome decimal.Decimal, clAPR float64, err error) {
```
Eisei24, line 735: The reason I had to change the whole method from a validator list `validators` to the `dashboardId` (+group filter) is because I could not change the query for clickhouse to
`WHERE validator_index IN ($1)`

For a large dashboard, e.g. 250,000 validators, you get a "Max query size exceeded" error.

backend/pkg/api/data_access/vdb_summary.go
```diff
@@ -114,113 +113,123 @@ func (d *DataAccessService) GetValidatorDashboardSummary(ctx context.Context, da
...
-			From(goqu.T(table).As("r")).
-			GroupBy(goqu.L("result_group_id"))
-
-		if len(validators) > 0 {
+	ds := goqu.Dialect("postgres").
+		From(goqu.L(fmt.Sprintf(`%s AS r FINAL`, clickhouseTable))).
+		With("validators", goqu.L("(SELECT dashboard_id, group_id, validator_index FROM users_val_dashboards_validators WHERE dashboard_id = ?)", dashboardId.Id)).
+		Select(
+			goqu.L("ARRAY_AGG(r.validator_index) AS validator_indices"),
+			goqu.L("SUM(COALESCE(r.attestations_reward, 0) + COALESCE(r.blocks_cl_reward, 0) + COALESCE(r.sync_rewards, 0) + COALESCE(r.blocks_cl_slasher_reward, 0)) AS cl_rewards"),
```
invis-bitfly, line 125:
```suggestion
			goqu.L("COALESCE(SUM(r.attestations_reward + r.blocks_cl_reward + r.sync_rewards + r.blocks_cl_slasher_reward), 0) AS cl_rewards"),
```
these columns arent nullable in clickhouse so this should work

backend/pkg/api/data_access/vdb_summary.go
```diff
@@ -114,113 +113,123 @@ func (d *DataAccessService) GetValidatorDashboardSummary(ctx context.Context, da
...
+	for _, row := range queryResult {
+		if row.MinEpochStart < epochMin {
+			epochMin = row.MinEpochStart
+		}
+		if row.MaxEpochEnd > epochMax {
+			epochMax = row.MaxEpochEnd
+		}
+	}
 	// ------------------------------------------------------------------------------------------------------------------
 	// Get the EL rewards
```
invis-bitfly, line 186: `validator_dashboard_view_execution_rewards_epoch` is a proxied view that exists in clickhouse that exposes `value` `proposer` and `epoch` - in theory this can even be combined in the above query.


but not really important enough to have this block the pr

backend/pkg/api/data_access/vdb_summary.go
```diff
@@ -733,62 +739,138 @@ func (d *DataAccessService) GetValidatorDashboardGroupSummary(ctx context.Contex
...
+			goqu.L("(SUM(COALESCE(r.balance_end,0)) + SUM(COALESCE(r.withdrawals_amount,0)) - SUM(COALESCE(r.deposits_amount,0)) - SUM(COALESCE(r.balance_start,0))) AS reward")).
+		From(goqu.L(fmt.Sprintf("%s AS r FINAL", table)))
+
+	if len(dashboardId.Validators) > 0 {
+		rewardsDs = rewardsDs.
+			Where(goqu.L("validator_index IN ?", dashboardId.Validators))
+	} else {
+		rewardsDs = rewardsDs.
+			InnerJoin(goqu.L("users_val_dashboards_validators v"), goqu.On(goqu.L("r.validator_index = v.validator_index"))).
+			Where(goqu.L("v.dashboard_id = ?", dashboardId.Id))
```
invis-bitfly, line 782: this is prob slow as it isn't using the `with` subquery optimization that gets used above

backend/pkg/api/data_access/vdb_summary.go
```diff
@@ -733,62 +739,138 @@ func (d *DataAccessService) GetValidatorDashboardGroupSummary(ctx context.Contex
...
 
-	query = `
-	SELECT
-		COALESCE(SUM(COALESCE(rb.value / 1e18, fee_recipient_reward)), 0)
-	FROM blocks
-	LEFT JOIN execution_payloads ON blocks.exec_block_hash = execution_payloads.block_hash
-	LEFT JOIN relays_blocks rb ON blocks.exec_block_hash = rb.exec_block_hash
-	WHERE proposer = ANY($1) AND status = '1' AND slot >= (SELECT MIN(epoch_start) * $2 FROM %s WHERE validator_index = ANY($1));`
-	err = db.AlloyReader.GetContext(ctx, &elIncome, fmt.Sprintf(query, table), validators, utils.Config.Chain.ClConfig.SlotsPerEpoch)
+	err = d.alloyReader.GetContext(ctx, &elIncome, query, args...)
```
invis-bitfly, line 857: same as above, could be rewriten to use the proxy view in clickhouse

invis-bitfly (approved):
only minor things found, nothing critical really, so marking as approved

backend/pkg/api/data_access/vdb_summary.go
Eisei24: I will keep it in mind for the future.

backend/pkg/api/data_access/vdb_summary.go
Eisei24: Changed.

backend/pkg/api/data_access/vdb_summary.go
Eisei24: I changed it to the subquery for here and also the validator slashing list which should now be all clickhouse queries.
However I also removed the WHERE condition after the INNER JOIN since it was redudant, or was there an unknown performance trick that needed it?

invis-bitfly: > However I also removed the WHERE condition after the INNER JOIN since it was redudant, or was there an unknown performance trick that needed it?

doesnt seem like the clickhouse query planner is smart enough to optimize the join that way. without the where filter the query always scans the complete `from` table

Eisei24: Ah, I see. I reverted the removal of the WHERE.

PR #658, opened 2024-08-01
Bids 3187/api archived dashboard auth

remoterami:
based on #511


backend/pkg/api/handlers/auth.go
```diff
@@ -779,3 +780,66 @@ func (h *HandlerService) ManageViaApiCheckMiddleware(next http.Handler) http.Han
...
+		return err
+	}
+	if dashboard.IsArchived {
+		return newForbiddenErr("dashboard with id %v is archived", dashboardId)
+	}
+	return nil
+}
+
+// allows to return a middleware auth check which excludes specified endpoints under the same subrouter
+func (h *HandlerService) GetMiddlewareExcludedRoutes(middlewareCheckFunc func(context.Context, *http.Request, ...interface{}) error, excludedRoutes []*mux.Route, params interface{}) func(http.Handler) http.Handler {
```
LuccaBitfly, line 801: This seems very complicated for such a minimal use case, as I don't think we'll need this anywhere else. I'd rather just have separate subrouters with their required middlewares instead of doing this exclusion logic.

backend/pkg/api/router.go
```diff
@@ -239,16 +239,22 @@ func addValidatorDashboardRoutes(hs *handlers.HandlerService, publicRouter, inte
...
+		{http.MethodPut, "/{dashboard_id}/archiving", hs.PublicPutValidatorDashboardArchiving, hs.InternalPutValidatorDashboardArchiving},
+	}
+	archivalRoutes := addEndpointsToRouters(archivalEndpoints, publicDashboardRouter, internalDashboardRouter)
+
 	// add middleware to check if user has access to dashboard
 	if !cfg.Frontend.Debug {
-		publicDashboardRouter.Use(hs.VDBAuthMiddleware, hs.ManageViaApiCheckMiddleware)
-		internalDashboardRouter.Use(hs.VDBAuthMiddleware)
+		publicDashboardRouter.Use(hs.VDBAuthMiddleware, hs.ManageViaApiCheckMiddleware, hs.GetMiddlewareExcludedRoutes(hs.GetVDBArchivedCheckMiddleware, archivalRoutes, nil))
+		internalDashboardRouter.Use(hs.VDBAuthMiddleware, hs.GetMiddlewareExcludedRoutes(hs.GetVDBArchivedCheckMiddleware, archivalRoutes, nil))
```
LuccaBitfly, line 253: Addendum to my last comment, you can just do this here:
```suggestion
		publicDashboardRouter.Use(hs.VDBAuthMiddleware, hs.ManageViaApiCheckMiddleware)
		internalDashboardRouter.Use(hs.VDBAuthMiddleware)
		// create new subrouters for archived check middleware, will be used for all endpoints added after this
		publicDashboardRouter = publicDashboardRouter.NewRoute().Subrouter()
		internalDashboardRouter = internalDashboardRouter.NewRoute().Subrouter()
		publicDashboardRouter.Use(hs.VDBArchivedCheckMiddleware)
		internalDashboardRouter.Use(hs.VDBArchivedCheckMiddleware)
```

LuccaBitfly (changes requested):

backend/pkg/api/handlers/auth.go
remoterami: Agreed it does, but encapsulated like this it's prob not much worse than fine-grained custom subrouter creation.

Also I could imagine making use of it again. Not sure if that's good examples, but here are some random thoughts:
1. Maybe we want to introduce in-depth premium-only endpoints (hard to think of a good example.. `/{dashboard_id}/name/history` as a stupid one); instead of checking the users premium status in all associated data access methods, we'd just have one single place to manage
2. Or we want to exclude a number of subroutes for mobile access for some reason, like `/{dashboard_id}/heatmap` because that data can't be presented anyways because the screen is too small or sth.

I can remove it, lmk if you like that version more. But will bookmark this for later :stuck_out_tongue:

backend/pkg/api/router.go
```diff
@@ -283,6 +276,24 @@ func addValidatorDashboardRoutes(hs *handlers.HandlerService, publicRouter, inte
...
+		internalDashboardRouter = internalDashboardRouter.NewRoute().Subrouter()
+		publicDashboardRouter.Use(hs.VDBArchivedCheckMiddleware)
+		internalDashboardRouter.Use(hs.VDBArchivedCheckMiddleware)
+	}
+
+	archivalEndpoints := []endpoint{
+		{http.MethodDelete, "/{dashboard_id}", hs.PublicDeleteValidatorDashboard, hs.InternalDeleteValidatorDashboard},
+		{http.MethodPut, "/{dashboard_id}/archiving", hs.PublicPutValidatorDashboardArchiving, hs.InternalPutValidatorDashboardArchiving},
+	}
+
```
LuccaBitfly, line 295: You first need to add the archiving endpoints, then do the middleware-adding logic, and then add the other endpoints. If you create a new (sub)router and add a middleware to it, only the endpoints added to that router will be affected by the middleware.

LuccaBitfly (changes requested):

LuccaBitfly (approved):
lgtm

PR #661, opened 2024-08-01
Bids 3175/data access user creation

Eisei24:


backend/pkg/api/data_access/user.go
```diff
@@ -96,13 +143,27 @@ func (d *DataAccessService) UpdatePasswordResetTime(ctx context.Context, userId
...
+	var result string
+
+	err := d.userReader.GetContext(ctx, &result, `
+    	SELECT
+			email_confirmation_hash
+		FROM users
+		WHERE id = $1`, userId)
+
+	return result, err
 }
```
remoterami, line 155: I'm honestly not even sure why this method exists, I don't think it's needed. My bad, would you remove it? :)

remoterami (changes requested):
ptal

backend/pkg/api/data_access/user.go
```diff
@@ -37,43 +37,84 @@ type UserRepository interface {
...
-	// TODO @DATA-ACCESS
 	// (password is already hashed)
-	return d.dummy.UpdateUserPassword(ctx, userId, password)
+
+	_, err := d.userWriter.ExecContext(ctx, `
+		UPDATE users 
+		SET 
+			password = $1
+		WHERE id = $2
+	`, password, userId)
```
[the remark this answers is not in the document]
Eisei24: Right, makes sense to have it like for the email.

backend/pkg/api/data_access/user.go
Eisei24: Sure, removed.

remoterami (approved):
lgtm

PR #663, opened 2024-08-02
(BIDS-3256) login and singup pages: removal of placeholders

thib-wien:
according to the new design of Butta


benji-bitfly:
LGTM

PR #665, opened 2024-08-05
fix(dashboard): enable 1h timeframe, fix types error and loading bug

MauserBitfly:
This PR: 
- enabled the 1h timeframe for the dashboard summary
- fixes a type error introduced by new props on the Validator Dashboard api types
- fixes a bug where we infinitely try to load our own validator dashboards when we encounter an error while loading the overview. 
  - new logic only tries to load our own validator dashboard if the dashboard we tried was not our own (like shared) or was archived.

BIDS-3293


benji-bitfly:
LGTM :heavy_check_mark:

PR #670, opened 2024-08-06
Bids 3184/restructure validator dashboard header

MauserBitfly:
This PR
- changes the arrangement of the dashboard header and control rows.
  - changes the `create new dashboard button` to secondary
  - moves some buttons from the dashboard header to the control row
  - combines dashboard configuration buttons into one settings menu
  - changing the label of the Dashboard selections to `Validators` / `Accounts` instead of the selected name
  - adds the `global` Rocket Pool (RP) Switch to the settings - currently disabled in production. 
- creates a `BcMenuBar` component wrapping the Primevue `Menubar`to be used in the header and control row
  - changes the default style for the menu bar and uses the previous style only for highlighted menu bar items
- uses the secondary button style for the content filter button

This PR:
- does not change the logic when buttons are shown only where and how.


benji-bitfly (commented):
LGTM :rocket:

PR #671, opened 2024-08-06
docs(decision): use `conventional commits`

marcel-bitfly:
Add `decision record` with first `decision proposal`.

See: BEDS-129


MauserBitfly (approved):
Looks good to me ... I think we already can set the status to `accepted` for the conventional commits

PR #674, opened 2024-08-06
feat(dashboard): add current name to rename dialog title

MauserBitfly:
This PR:
- adds the current dashboard name to the title in the rename dashboard dialog

### DIFF
--- a/frontend/components/dashboard/DashboardRenameModal.vue
+++ b/frontend/components/dashboard/DashboardRenameModal.vue
@@ -48,7 +48,7 @@ const rename = async () => {
 
 <template>
   <div class="dashboard_rename_modal_container">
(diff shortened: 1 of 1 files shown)


frontend/components/dashboard/DashboardRenameModal.vue
```diff
@@ -56,6 +58,12 @@ const rename = async () => {
 </template>
 
 <style lang="scss" scoped>
+@use "~/assets/css/utils.scss";
+
+:global(.dashboard-rename-modal-header .p-dialog-title) {
```
thib-wien, line 63: As I do not know it:
Is it sure that this selector will be removed from global rules by Vue when this dialog component is unmounted ?

MauserBitfly: yes that should be the case, but just to be sure it has a distinct name (`dashboard-rename-modal-header`) so it should never bother any other styles.

thib-wien (approved):

PR #676, opened 2024-08-07
fix(dashboard): improve summary chart from handling

MauserBitfly:
This PR:
- fixes the dashboard summary chart time handling
  - omits the from value if you place the slider on the far left to prevent us from going to far in the past and cause a ws error
  - refreshes the current slot that is based for the time frame range calculation when the user input changes
  - prevents the UI from being locked if we get a WS error (previously the error message was placed over the UI and prevented the user from interacting with the chart to change the settings. 

BEDS-164

### DIFF
--- a/frontend/components/dashboard/chart/SummaryChart.vue
+++ b/frontend/components/dashboard/chart/SummaryChart.vue
@@ -46,7 +46,7 @@ const { overview } = useValidatorDashboardOverviewStore()
 const { groups } = useValidatorDashboardGroups()
 const { latestState } = useLatestStateStore()
 const latestSlot = ref(latestState.value?.current_slot || 0)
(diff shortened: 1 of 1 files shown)


frontend/components/dashboard/chart/SummaryChart.vue
```diff
@@ -388,10 +391,19 @@ const validateDataZoom = (instant?: boolean) => {
...
+  // in that case the backend will go back depending on the max secons of the dashboard settings
+  if(timestamps.fromIndex <= bufferSteps){
+    fromTs = undefined
+  }
   const newTimeFrames = {
-    from: timestamps.fromTs,
+    from: fromTs,
     to: timestamps.toTs
   }
+  console.log('timestamps', timestamps.fromIndex, newTimeFrames.from)
```
benji-bitfly, line 406: :eyes:

MauserBitfly: fixed ;-)

PR #677, opened 2024-08-07
fix(dashboard): place hover menubar on the correct element

MauserBitfly:
This Pr:
- places the hover background on the correct element of the BcMenuBar elements


benji-bitfly (commented):
LGTM :checkered_flag:

PR #682, opened 2024-08-08
(BEDS-137) API heatmap aggregation

remoterami:
Based on #638


backend/pkg/api/handlers/common.go
```diff
@@ -337,6 +336,30 @@ func (h *HandlerService) handleDashboardId(ctx context.Context, param string) (*
 	return dashboardId, nil
 }
 
+const chartDatapointLimit uint64 = 200
+
+// helper function to retrieve allowed chart timestamp boundaries according to the users premium perks at the current point in time
+func (h *HandlerService) getCurrentChartTimeLimitsForUser(ctx context.Context, dashboardId *types.VDBId, aggregation enums.ChartAggregation) (uint64, uint64, uint64, error) {
```
LuccaBitfly, line 342: Minor: IMO when there are this many return values, the values should either be documented in the function doc or be named variables, or maybe you could encapsulate them in a struct.

LuccaBitfly, line 342: Minor: For clarity
```suggestion
func (h *HandlerService) getCurrentChartTimeLimitsForDashboard(ctx context.Context, dashboardId *types.VDBId, aggregation enums.ChartAggregation) (uint64, uint64, uint64, error) {
```
Since this is dependent on the dashboard owner (i.e. in case of a shared dasboard), not the requesting user.

backend/pkg/api/handlers/internal.go
```diff
@@ -1038,35 +1036,22 @@ func (h *HandlerService) InternalGetValidatorDashboardSummaryChart(w http.Respon
...
 	}
-	minAllowedTs := latestExportedTs - min(maxAge, latestExportedTs)                                      // min to prevent underflow
-	secondsPerEpoch := uint64(12 * 32)                                                                    // TODO: fetch dashboards chain id and use correct value for network once available
-	maxAllowedInterval := chartDatapointLimit*uint64(aggregation.Duration(secondsPerEpoch).Seconds()) - 1 // -1 to make sure we don't go over the limit
-	afterTs, beforeTs := v.checkTimestamps(q.Get("after_ts"), q.Get("before_ts"), latestExportedTs, minAllowedTs, maxAllowedInterval)
+	afterTs, beforeTs := v.checkTimestamps(r, maxAvailableTs, minAllowedTs, maxAllowedInterval)
 	if v.hasErrors() {
-		handleErr(w, v)
-		return
+		handleErr(w, err)
```
LuccaBitfly, line 1051: missing return after this

LuccaBitfly (changes requested):
Looks good, some minor issues

PR #683, opened 2024-08-08
style(eslint): add `plugin perfectionist`

marcel-bitfly:
See: https://perfectionist.dev/guide/introduction
See: BEDS-128


frontend/components/bc/BcCurrencySelection.vue
```diff
@@ -2,10 +2,12 @@
...
 
 defineProps<{
-  showCurrencyIcon: boolean
+  showCurrencyIcon: boolean,
 }>()
 
-const { currency, withLabel, setCurrency } = useCurrency()
+const {
+  currency, setCurrency, withLabel,
+} = useCurrency()
```
marcel-bitfly, line 10: fyi: should be like that, but atm could not get autofixed, as the rule: 
https://eslint.style/rules/js/object-property-newline#object-property-newline
lacks a proper destructuring logic
```suggestion
const {
  currency,
  setCurrency,
  withLabel,
} = useCurrency()
```

MauserBitfly (approved):
LGTM

PR #685, opened 2024-08-08
fic(dashboard): summary chart - use default steps and make loading mo…

MauserBitfly:
This PR:
- fixes the summary chart loading
  - adds default time steps per time range
  - makes the loading / checking mechanism more resilient

BEDS-178


thib-wien (approved):
🚀

PR #688, opened 2024-08-09
feat(notifications): add machine notifications table

MauserBitfly:
This PR:
- adds the Machine table to the notifications dashboard

This PR is not:
- tested with real data (the backend returns mocked data for now)
- adding the real subscription count (this will come from the overview data, that is WIP)


benji-bitfly:
Lgtm :heavy_check_mark:  There are conflicts to solve.

PR #689, opened 2024-08-09
(BEDS-205) API: added possible values for event types in notificaitons

remoterami:


backend/pkg/api/types/notifications.go
```diff
@@ -37,7 +37,7 @@ type NotificationDashboardsTableRow struct {
 	GroupName          string   `json:"group_name"`
 	NotificationId     uint64   `json:"notification_id"` // may be string? db schema is not defined afaik
 	EntityCount        uint64   `json:"entity_count"`
-	EventTypes         []string `json:"event_types"`
+	EventTypes         []string `json:"event_types" tstype:"'validator_offline' | 'group_offline' | 'attestation_missed' | 'proposal_success' | 'proposal_missed' | 'proposal_upcoming' | 'sync' | 'withdrawal' | 'slashed_own'| 'incoming_tx' | 'outgoing_tx' | 'transfer_erc20' | 'transfer_erc721' | 'transfer_erc1155'" faker:"oneof: validator_offline, group_offline, attestation_missed, proposal_success, proposal_missed, proposal_upcoming, sync, withdrawal, slashed_own, incoming_tx, outgoing_tx, transfer_erc20, transfer_erc721, transfer_erc1155"`
```
LuccaBitfly, line 40: Slash events should be `got_slashed` and `has_slashed`. `validator_online` and `group_online` should prob also be included.

LuccaBitfly (changes requested):

LuccaBitfly (approved):
lgtm, don't forget FE approval

frontend/types/api/notifications.ts
```diff
@@ -43,7 +43,7 @@ export interface NotificationDashboardsTableRow {
   group_name: string;
   notification_id: number /* uint64 */; // may be string? db schema is not defined afaik
   entity_count: number /* uint64 */;
-  event_types: string[];
+  event_types: 'validator_online' | 'validator_offline' | 'group_online' | 'group_offline' | 'attestation_missed' | 'proposal_success' | 'proposal_missed' | 'proposal_upcoming' | 'sync' | 'withdrawal' | 'got_slashed' | 'has_slashed' | 'incoming_tx' | 'outgoing_tx' | 'transfer_erc20' | 'transfer_erc721' | 'transfer_erc1155';
```
MauserBitfly, line 46: I think this should be a string.

MauserBitfly (commented):
thanks for the types, but I think I found an issue

frontend/types/api/notifications.ts
remoterami: Not sure if I understand correctly, the point of this was to not use a `string` but a union type, right? But I see that it's not an array of event types any more, just a single value - that wasn't intended, changed it (thanks!). Was this was you meant or should I change it to something else?

MauserBitfly: Yes exactly that, thanks ;-)

MauserBitfly:
@remoterami @LuccaBitfly : Pushed a small frontend fix to fix a type error

MauserBitfly (approved):
LGTM

PR #690, opened 2024-08-12
fix: update node version to 22-alpine

MauserBitfly:
This PR:
- updates the node version to 22-alpine in the docker file

BEDS-204


marcel-bitfly (approved):
👍

PR #695, opened 2024-08-13
fix(dashboard): shrink summary chart filter font and paddings on smal…

MauserBitfly:
This PR:
- shrinks the font size and paddings of the summary chart filter dropdowns for smaller screens, so that more of the label is visible.

BEDS-206


benji-bitfly:
lgtm :rocket:

benji-bitfly (approved):

PR #696, opened 2024-08-13
fix(slot exporter): improve validator online / offline detection

peterbitfly:


backend/pkg/exporter/modules/slot_exporter.go
```diff
@@ -619,54 +668,6 @@ func ExportSlot(client rpc.Client, slot uint64, isHeadEpoch bool, tx *sqlx.Tx) e
...
-	}
-
-	// save the proposal to bigtable
-	err = db.BigtableClient.SaveProposal(block)
-	if err != nil {
-		return fmt.Errorf("error exporting proposal to bigtable for slot %v: %w", block.Slot, err)
-	}
-
-	// save the block data to the db
-	err = edb.SaveBlock(block, false, tx)
```
Eisei24, line 660: **issue:** This part was also moved before the big if but we change the content of block within the if, e.g. in the currently adjusted `SaveValidators`.
AFAIK we only wanted to move the bigtable duties to the front.

If a `SaveBlock` at the start is fine please let me know.

Eisei24 (commented):
Please check the changed `SaveBlock` handling.

backend/pkg/exporter/modules/slot_exporter.go
peterbitfly: i think its fine as we don't touch the mutated fields in the `SaveBlock` function. if you found a place where `SaveBlock` touches one of those fields please let me know.

Eisei24 (approved):
lgtm

PR #697, opened 2024-08-13
chore: add `flag` to disable `unauthorized tls certs` to `dev script`

marcel-bitfly:
In order to run `dev server` with `https` we have to create a `certificate`. But this certificate is not `authorized` by a known and trusted `authority`. This is why we have to `disable` the rejection of that `certificate`

See: BEDS-97


MauserBitfly (approved):
LGTM

PR #698, opened 2024-08-13
(BEDS-138) api password resetting

remoterami:


backend/pkg/api/handlers/auth.go
```diff
@@ -99,6 +100,58 @@ Best regards,
...
+		return newForbiddenErr("password reset not allowed")
+	}
+
+	// 1. check last confirmation time to enforce ratelimit
+	lastTs, err := h.dai.GetPasswordResetTime(ctx, userId)
+	if err != nil {
+		return errors.New("error getting confirmation-ts")
+	}
+	if lastTs.Add(authResetEmailRateLimit).After(time.Now()) {
+		return errors.New("rate limit reached, try again later")
```
LuccaBitfly, line 121: This would return 500 to the user, something like 429 would be more appropriate.

backend/pkg/api/handlers/auth.go
```diff
@@ -239,9 +292,95 @@ func (h *HandlerService) InternalPostUserConfirm(w http.ResponseWriter, r *http.
...
+		handleErr(w, err)
+		return
+	}
+
 	returnOk(w, nil)
 }
 
+func (h *HandlerService) InternalPostUserPasswordResetHash(w http.ResponseWriter, r *http.Request) {
+	var v validationError
+	resetToken := v.checkUserEmailToken(mux.Vars(r)["token"])
```
LuccaBitfly, line 337: The token only needs to be part of the URL in FE, for the API endpoint we can move this into the request payload.

backend/pkg/api/handlers/auth.go
```diff
@@ -99,6 +100,58 @@ Best regards,
 	return nil
 }
 
+// TODO move to service?
+func (h *HandlerService) sendResetEmail(ctx context.Context, userId uint64, email string) error {
```
LuccaBitfly, line 104:
```suggestion
func (h *HandlerService) sendPasswordResetEmail(ctx context.Context, userId uint64, email string) error {
```

backend/pkg/api/handlers/auth.go
```diff
@@ -239,9 +292,95 @@ func (h *HandlerService) InternalPostUserConfirm(w http.ResponseWriter, r *http.
...
+		handleErr(w, err)
+		return
+	}
+	resetTime, err := h.dai.GetPasswordResetTime(r.Context(), userId)
+	if err != nil {
+		handleErr(w, err)
+		return
+	}
+	if resetTime.Add(authEmailExpireTime).Before(time.Now()) {
+		handleErr(w, errors.New("reset link expired"))
```
LuccaBitfly, line 363: This would return 500 to the user, more approriate would be 400 or 410 maybe

backend/pkg/api/handlers/auth.go
```diff
@@ -239,9 +292,95 @@ func (h *HandlerService) InternalPostUserConfirm(w http.ResponseWriter, r *http.
...
+		handleErr(w, errors.New("error hashing password"))
+		return
+	}
+	err = h.dai.UpdateUserPassword(r.Context(), userId, string(passwordHash))
+	if err != nil {
+		handleErr(w, err)
+		return
+	}
+
+	// TODO: purge all user sessions
```
LuccaBitfly, line 379: Pls implement, see https://github.com/alexedwards/scs?tab=readme-ov-file#enumerate-all-sessions

LuccaBitfly (changes requested):

backend/pkg/api/router.go
```diff
@@ -103,11 +103,12 @@ func addRoutes(hs *handlers.HandlerService, publicRouter, internalRouter *mux.Ro
 
 		{http.MethodPost, "/users", nil, hs.InternalPostUsers},
 		{http.MethodPost, "/users/email-confirmations/{token}", nil, hs.InternalPostUserConfirm},
+		{http.MethodPut, "/users/password/reset", nil, hs.InternalPutUserPasswordReset},
```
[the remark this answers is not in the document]
remoterami: The reasons I've chosen `PUT` here are:
- most examples you can find online around this suggest using it
- the endpoints should be idempotent: no matter how often you call them, the result should be the same. This is the case here imo
- it further distinguishes the endpoint from the `POST` version
- we're also using it for email change requests (`/users/me/email`)

So I'd keep it, if you've a strong opinion about it pls lmk why.

backend/pkg/api/handlers/auth.go
remoterami: Will do in #716 after some testing

backend/pkg/api/router.go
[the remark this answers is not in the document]
LuccaBitfly: Generally, `PUT` is used to update/replace an existing resource. Although it can be used to create new resources, this is rather an exception and we currently only use it for replacing existing resources in our API. I would argue that this endpoint creates a new password reset request (and does not replace an existing one), so from a consistency perspective it should be `POST`. 
But, more importantly: You are correct that `PUT` endpoints must be idempotent, which is not the case for this endpoint. Calling the endpoint twice will either return a forbidden error (due to the password reset ts) or set a new, distinct password reset token for the user. Both of these are not the same result as the first call. If you try to call `PUT /users/me/email` or even `PUT /users/me/password` with the same parameters, it will always result in the same DB state.

LuccaBitfly (approved):
lgtm

PR #699, opened 2024-08-13
(BEDS-274) fix auth

LuccaBitfly:
This PR fixes the bug that prevents users from creating a dashboard, aswell as cleaning up the middlewares a bit. With these changes, the user ID (if available) is always loaded into the request context at the beginning. Further handlers should then only need to load user ID by context and throw unauthorized themselves, if they expect it to be in context.

### DIFF
--- a/backend/pkg/api/handlers/auth.go
+++ b/backend/pkg/api/handlers/auth.go
@@ -101,15 +101,13 @@ Best regards,
 
 func (h *HandlerService) GetUserIdBySession(r *http.Request) (uint64, error) {
 	user, err := h.getUserBySession(r)
-	if err != nil {
-		return 0, err
-	}
-	return user.Id, nil
+	return user.Id, err
 }
 
 const authHeaderPrefix = "Bearer "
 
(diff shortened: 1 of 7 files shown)


backend/pkg/api/handlers/auth.go
```diff
@@ -715,45 +719,42 @@ func GetUserIdStoreMiddleware(userIdFunc func(r *http.Request) (uint64, error))
...
-			}
-			// store user id in context
-			ctx := r.Context()
-			ctx = context.WithValue(ctx, ctxUserIdKey, userId)
-			r = r.WithContext(ctx)
+		userId, err := GetUserIdByContext(r)
+		if err != nil {
+			handleErr(w, err)
+			return
+		}
```
remoterami, line 737: Would use the same check in `ManageViaApiCheckMiddleware` (instead of `r.Context().Value(ctxUserIdKey).(uint64)`)

remoterami (approved):
utACK, just a small suggestion

PR #700, opened 2024-08-13
fix(dashboard): make summary chart range selection more resilient

MauserBitfly:
This PR:
- Tries to optimize the range selection in the dashboard summary chart
  - changing the MAX_DATA_POINTS to 199
  - introducing a min range
  - refactoring the time range checking logic
  - increasing the default range for the Epoch Range 
 
BEDS-217


benji-bitfly (approved):
lgtm

PR #705, opened 2024-08-13
fix(dashboard): harmonize calculation of total cl rewards

peterbitfly:


backend/pkg/api/data_access/vdb_summary.go
```diff
@@ -122,7 +122,7 @@ func (d *DataAccessService) GetValidatorDashboardSummary(ctx context.Context, da
 		With("validators", goqu.L("(SELECT dashboard_id, group_id, validator_index FROM users_val_dashboards_validators WHERE dashboard_id = ?)", dashboardId.Id)).
 		Select(
 			goqu.L("ARRAY_AGG(r.validator_index) AS validator_indices"),
-			goqu.L("COALESCE(SUM(r.attestations_reward + r.blocks_cl_reward + r.sync_rewards + r.blocks_cl_slasher_reward), 0) AS cl_rewards"),
+			goqu.L("(SUM(COALESCE(r.balance_end,0)) + SUM(COALESCE(r.withdrawals_amount,0)) - SUM(COALESCE(r.deposits_amount,0)) - SUM(COALESCE(r.balance_start,0))) AS cl_rewards"),
```
Eisei24, line 125: **suggestion:** Because none of the clickhouse types have `Nullable` all these columns should never be `NULL`.
The `COALESCE` is then not required.

Eisei24, line 125: **issue:** Isn't this then the whole reward including the el part and not just cl?

Eisei24 (commented):
Please have a look at my comments.

Eisei24:
This way of calculating the cl_reward is also used twice for the rewards view if you want to change it there too.

backend/pkg/api/data_access/vdb_summary.go
peterbitfly: no, as el rewards are not counted towards the validator cl balance, they are separately handled on the evm level

Eisei24: Thanks for the info.
I also tested it now for a dashboard with proposals and the old and new ways give identical values.

PR #717, opened 2024-08-14
feat(pricing): add promo code support

MauserBitfly:
This PR:
- adds promo code via url (?promoCode=ABCD-EFGH-...)  support
  - shows overlay on pricing page if promo code is provided
  - passes on the promo code between the `pricing`, `login` and `register`page
  - uses the promo code when creating a purchase session

Not for the reviewer: You will not be able to create a checkout session when you work localy, but you should see in the failed network request body the `promotionCode`.  
If you already have an active subscription you can test by putting a `!` in row 97 in the `PremiumProductBox`


benji-bitfly (approved):
Incredible job! LGTM :heavy_check_mark: :raised_hands:

PR #718, opened 2024-08-14
fix(pricing): minor fixes for the promo overlay

MauserBitfly:
This Pr:
- makes some small tweeks for the pricing overlay
  - you can now click on the left / right side of the overlay content
  - border radius added
  - increased z-index to 6 to be over the gradient blur


benji-bitfly (approved):
LGTM :raised_hands:

PR #720, opened 2024-08-14
Beds 181/add balance to header

marcel-bitfly:


frontend/components/dashboard/DashboardValidatorOverview.vue
```diff
@@ -0,0 +1,176 @@
...
+<style lang="scss" scoped>
+@use "~/assets/css/main.scss";
+.container {
+  @include main.container;
+  display: flex;
+  align-items: center;
+  justify-content: space-between;
+  overflow-x: auto;
+  margin-top: var(--padding-large);
+  transform: translateY(0px); // hack: on safari top-border is not shown
```
marcel-bitfly, line 166: fyi: also copied latest change over

frontend/components/dashboard/DashboardValidatorOverview.vue
```diff
@@ -0,0 +1,176 @@
...
+      >
+        <FontAwesomeIcon
+          class="link optical-correction"
+          :icon="faArrowUpRightFromSquare"
+        />
+      </BcButtonIcon>
+      <template #additionalInfo>
+        {{ $t('dashboard.validator.overview.validators_balance.balance_total') }}
+        <span class="bold">
+          {{ formatValueWei(overview?.balances.total ?? 0) }}
```
MauserBitfly, line 127: Nitpick: even dough the formatValueWei can handle `number` and `string` the total is of type string so I would default to `'0'`

MauserBitfly (approved):
LGTM, just one small Nitpick you might consider before merging.

frontend/components/dashboard/DashboardValidatorOverview.vue
marcel-bitfly: In general I would second that, but in template strings in between `{{ }}` mustache syntax, there is no difference

PR #722, opened 2024-08-16
chore(checkout): using `nuxt`s `ssl handling`

marcel-bitfly:
Nuxt can create `selfsinged certificates` on its own. This is especially heplful when checking out the repo or when using `git worktree`s feature, as there a fewer steps to get the dev server running.

See: BEDS-97


benji-bitfly (approved):
LGTM :rocket:

PR #724, opened 2024-08-16
Beds 162/display message if api returns empty

benji-bitfly:
This PR:

- adds message for when the vlidators are added but do not fulfill any tasks.
- adds text and Component for the message 
 
Screenshot from 2024-08-14 16-16-35


MauserBitfly (approved):
LGTM, nice job

PR #725, opened 2024-08-19
fix(dashboard): have a min button size for the dashboard header butto…

MauserBitfly:
This PR:
- sets a min width for the dashboard header buttons 
- centers the content of the dashboard header buttons if they are not a dropdown

image


benji-bitfly (approved):
LGTM :rocket:

PR #726, opened 2024-08-19
feat: use clickhouse to source data for the average network efficienc…

peterbitfly:
…y, add support for the 1h time interval


Eisei24 (approved):
lgtm

PR #729, opened 2024-08-19
fix(dashboard): center disabled header buttons, use plural for Valida…

MauserBitfly:
This PR:
- fixes some minor issues in the Dashboard Header Buttons
  - use Plural for the Validators / Accounts button
    - adjust the gap in the button on mobile so that the plural text is not truncated
  - fix the padding of disabled buttons


benji-bitfly (approved):
LGTM :heavy_check_mark:

PR #730, opened 2024-08-19
(BEDS-298) add network to dashboards

remoterami:
Based on #692


LuccaBitfly (changes requested):
lgtm, pls make sure that the fix for FE is pushed onto this branch before merging

LuccaBitfly (approved):
lgtm, pls make sure that the fix for FE is pushed onto this branch before merging

guybrush:
the cf-pipeline had some problems
```
error: failed to download node-v22.8.0-linux-x64.tar.gz
```

it worked after restarting the pipeline

PR #733, opened 2024-08-19
(BEDS-155) Data Access: adjusted address struct

remoterami:
Returning `is_contract` and optional `label` in all cases where an address is retrieved.


backend/pkg/api/data_access/vdb_blocks.go
```diff
@@ -382,7 +384,13 @@ func (d *DataAccessService) GetValidatorDashboardBlocks(ctx context.Context, das
 				Hash: t.Hash(hexutil.Encode(proposal.FeeRecipient)),
 			}
 			data[i].RewardRecipient = &rewardRecp
-			ensMapping[hexutil.Encode(proposal.FeeRecipient)] = ""
+			addressMapping[hexutil.Encode(proposal.FeeRecipient)] = nil
+			contractStatusRequests = append(contractStatusRequests, db.ContractInteractionAtRequest{
+				Address:  string(proposal.FeeRecipient),
```
Eisei24, line 389: **issue:** I don't see this giving you the correct string formatting for `GetAddressContractInteractionsAt`.
It seems to request the address strings in the `fmt.Sprintf("%x", byteSlice)` format (so without the "0x") not as `string(byteSlice)`.
This affects all new uses of `GetAddressContractInteractionsAt` but please also see my comment in withdrawals.

backend/pkg/api/data_access/vdb_withdrawals.go
```diff
@@ -393,12 +408,28 @@ func (d *DataAccessService) getNextWithdrawalRow(queryValidators []t.VDBValidato
 		withdrawalAmount = 0
 	}
 
+	ens_name, err := db.GetEnsNameForAddress(*address, utils.SlotToTime(nextWithdrawalSlot))
+	if err != sql.ErrNoRows {
+		return nil, err
+	}
+
+	contractStatusReq := []db.ContractInteractionAtRequest{{
+		Address: address.String(),
```
Eisei24, line 417: **question:** I am not sure whether `address.String()` returns the string with the "0x" prefix, please check.

backend/pkg/commons/db/migrations/postgres/20240822134034_add_address_tags.sql
```diff
@@ -0,0 +1,15 @@
+-- +goose Up
+-- +goose StatementBegin
+SELECT('up SQL query - create address_tags table');
+CREATE TABLE address_tags (
```
Eisei24, line 4: **minor:** Afaik should include `IF NOT EXISTS`.

backend/pkg/commons/db/migrations/postgres/20240822134034_add_address_tags.sql
```diff
@@ -0,0 +1,15 @@
...
+    address bytea NOT NULL UNIQUE,
+    tag CHARACTER VARYING(100) NOT NULL,
+    PRIMARY KEY (address, tag)
+);
+-- +goose StatementEnd
+
+-- +goose Down
+-- +goose StatementBegin
+SELECT('down SQL query - drop address_tags table');
+DROP TABLE address_tags;
```
Eisei24, line 14: **minor:** Afaik should include `IF EXISTS`.

Eisei24 (changes requested):
Please check the address string conversion.

backend/pkg/api/data_access/vdb_withdrawals.go
remoterami: It does, also it's the checksum hashed address which would not match

backend/pkg/api/data_access/vdb_blocks.go
remoterami: Ah you're right, good, ty

Eisei24 (approved):
utACK

backend/pkg/commons/db/migrations/postgres/20240822134034_add_address_tags.sql
```diff
@@ -0,0 +1,15 @@
+-- +goose Up
+-- +goose StatementBegin
+SELECT('up SQL query - create address_tags table');
+CREATE TABLE IF NOT EXISTS address_tags (
+    address bytea NOT NULL UNIQUE,
```
guybrush, line 5: maybe make this non-unique to able to have multiple tags per address?

guybrush (approved):

PR #739, opened 2024-08-19
(BEDS-306) api: add basic healthz endpoint

invis-bitfly:
status can be checked by response code (500=bad, 200=good) or the root `is_ok` flag
breakdown of reports is in `status_reports`
`tags` would allow for dynamic routing of alerts (for example all clickhouse alerts go to person a while all alloy alerts go to person b)
image

### DIFF
--- a/backend/.golangci.yml
+++ b/backend/.golangci.yml
@@ -111,9 +111,6 @@ linters-settings:
         msg: Use log.Infof instead
       - p: ^fmt\.Printf.*$
         msg: Use log.Infof instead
-      # ban usage of db.ClickHouseWriter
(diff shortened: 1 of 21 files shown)


peterbitfly:
This will carry out a health check on each request to the endpoint (which can be a lot given that the endpoint is public & usually health check services request the endpoint from multiple locations in short intervals to check geo-availability). I suggest to adopt the approach outlined in https://github.com/gobitfly/eth2-beaconchain-explorer/blob/master/services/monitoring.go (have the service to regularly write the status to the `service_status` table and have the api access that table to determine service health).

guybrush (approved):

PR #740, opened 2024-08-20
fix(dashboard): chart tooltip position, line color, min pre selection…

MauserBitfly:
This PR fixes:
- Tab pre selection
  - new default selection prop
  - use the correct default on validator and notifications dashboard
- summary chart
  - have an initial 3 % min range selection 
  - set an 0.3 opacity for the chart background lines 
  - keep the tooltip in the boundries of the chart
- try to fix the cut border top on the validator overview


benji-bitfly (approved):
LGTM :rocket:

PR #744, opened 2024-08-20
Fixed the summary for empty dashboards

Eisei24:


backend/pkg/api/data_access/vdb_summary.go
```diff
@@ -171,7 +171,12 @@ func (d *DataAccessService) GetValidatorDashboardSummary(ctx context.Context, da
 		return nil, nil, fmt.Errorf("error retrieving data from table %s: %v", clickhouseTable, err)
 	}
 
-	epochMin := int64(math.MaxInt64)
+	if len(queryResult) == 0 {
+		// No groups to show
+		return result, &paging, nil
+	}
+
+	epochMin := int64(math.MaxInt32)
```
Eisei24, line 179: Both of these changes would fix the bug on their own but I think it is good to have both.

peterbitfly (approved):

PR #745, opened 2024-08-20
Beds 262/data access password reset

Eisei24:


backend/pkg/api/data_access/user.go
```diff
@@ -76,8 +76,20 @@ func (d *DataAccessService) GetEmailConfirmationTime(ctx context.Context, userId
 }
 
 func (d *DataAccessService) GetPasswordResetTime(ctx context.Context, userId uint64) (time.Time, error) {
-	// TODO @DATA-ACCESS
-	return d.dummy.GetPasswordResetTime(ctx, userId)
+	result := time.Time{}
+
+	var queryResult sql.NullTime
+	err := d.userWriter.GetContext(ctx, &queryResult, `
```
remoterami, line 82: `userReader` should suffice, also see for `IsPasswordResetAllowed` and `GetUserIdByResetHash`

remoterami (changes requested):
ptal, small copy paste remark

backend/pkg/api/data_access/user.go
Eisei24: Right, changed for the 3 methods.

PR #748, opened 2024-08-20
(BEDS-295) implement automated api doc generation

LuccaBitfly:


backend/pkg/api/handlers/public.go
```diff
@@ -95,8 +114,66 @@ func (h *HandlerService) PublicPutAccountDashboardTransactionsSettings(w http.Re
...
+//	@Description	Create a new validator dashboard. **Note**: New dashboards will automatically have a default group created.
+//	@Security 		ApiKeyInHeader || ApiKeyInQuery
+//	@Tags			Validator Dashboards
+//	@Accept			json
+//	@Produce		json
+//	@Param			request		body		handlers.PublicPostValidatorDashboards.request	true	"`name`: Specify the name of the dashboard.<br>`network`: Specify the network for the dashboard. Possible options are:<ul><li>`ethereum`</li><li>`gnosis`</li></ul>"
+//	@Success		201			{object}	types.PostValidatorDashboardsResponse
+//	@Failure		400			{object}	types.ApiErrorResponse
+//	@Failure		409			{object}	types.ApiErrorResponse	"Conflict. The request could not be performed by the server because the authenticated user has already reached their dashboard limit."
+//	@Router			/validator-dashboards [post]
```
remoterami, line 128: Would be cool if there was a way to even generate stubs of these. Like at least the route, return objects etc.

remoterami (approved):
lgtm

backend/pkg/api/handlers/public.go
LuccaBitfly: best we can get is prob copilot auto completion 🤷

peterbitfly (approved):

PR #749, opened 2024-08-20
BEDS-311/port_InternalGetValidatorDashboardTotalWithdrawals_to_clickhouse

peterbitfly:
…khouse


invis-bitfly (approved):
could be simplified (checks thrice if it has validators or a dashboard id) but otherwise lgtm

PR #755, opened 2024-08-21
(BEDS-313) don't pass VDBId struct in archiving middleware

LuccaBitfly:
This PR changes the parameter of a data access function, so that it can never take a validator set / guest dashboard as an argument. also renames some stuff for clarity.


remoterami (approved):
lgtm

PR #759, opened 2024-08-21
Fixed GetValidatorDashboardGroupSummary for empty

Eisei24:
I also cleaned up the method by using goqu to avoid duplicated query parts.

### DIFF
--- a/backend/pkg/api/data_access/vdb_summary.go
+++ b/backend/pkg/api/data_access/vdb_summary.go
@@ -521,65 +521,44 @@ func (d *DataAccessService) GetValidatorDashboardGroupSummary(ctx context.Contex
 		return nil, err
 	}
 
-	query := `
-			WITH validators AS (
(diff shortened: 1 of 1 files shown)


backend/pkg/api/data_access/vdb_summary.go
```diff
@@ -606,18 +585,21 @@ func (d *DataAccessService) GetValidatorDashboardGroupSummary(ctx context.Contex
...
+	var rows []*QueryResult
+	err = d.clickhouseReader.SelectContext(ctx, &rows, query, args...)
 	if err != nil {
 		return nil, fmt.Errorf("error retrieving validator dashboard group summary data: %v", err)
 	}
 
+	if len(rows) == 0 {
+		return ret, nil
+	}
+
```
Eisei24, line 602: This alone is the fix.

remoterami (approved):
utACK

PR #761, opened 2024-08-21
(BEDS-332) log endpoint+params on api error

LuccaBitfly:


backend/pkg/api/handlers/common.go
```diff
@@ -724,108 +722,121 @@ func isUserAdmin(user *types.UserInfo) bool {
...
+		map[string]interface{}{
+			"endpoint": r.Method + " " + r.URL.Path,
+			"query":    r.URL.RawQuery,
+			"body":     string(body),
+		},
+	)
+}
+
+func returnInternalServerError(w http.ResponseWriter, r *http.Request, err error) {
+	logApiError(r, err)
```
LuccaBitfly, line 814: This is prob the most significant change

remoterami: The fixed `callerSkip` value of 3 might lead to unclear logging depending on how the error is handled. If handled via `handleErr` a depth of 4 would be required to log the expected function in the call stack.

Could solve this by forbidding all `returnXXX` shortcuts in all handlers (and always using `handleErr(w, r, newXXXErr("<error>"))`) or by passing the `callerSkip` to `logApiError`

remoterami (changes requested):
ptal

backend/pkg/api/handlers/common.go
LuccaBitfly: You're right. Added the `callerSkip` param. `returnInternalServerError` should be used either only by `handleErr` or only by handlers directly. Since I don't want to prevent handlers from using these quick return funcs, I'll choose the latter. And because no handler uses `returnInternalServerError`, I removed it.

remoterami (approved):
lgtm :+1:

PR #762, opened 2024-08-21
chore: update nuxt to 3.12.4

MauserBitfly:
This Pr: 
- updates Nuxt to Version 3.12.4 

Reason for the Update: the 3.12.3 had an issue with the cookie reactivity (example: switching currency did not update the formatted values on the current page) 

Note: 
the update was done by modifying the `package.json` and not with `npx nuxi upgrade` as this leads to type problems in the current version.


marcel-bitfly (approved):
👍

PR #763, opened 2024-08-21
Fixed the wrong EL APR calculation

Eisei24:


backend/pkg/api/data_access/vdb_management.go
```diff
@@ -342,10 +342,6 @@ func (d *DataAccessService) GetValidatorDashboardOverview(ctx context.Context, d
 			return nil
 		})
 	}
-	validators, err := d.getDashboardValidators(ctx, dashboardId, nil)
-	if err != nil {
-		return nil, fmt.Errorf("error retrieving validators from dashboard id: %v", err)
-	}
```
Eisei24, line 348: This just looked like a pointless query to have.

backend/pkg/api/data_access/vdb_summary.go
```diff
@@ -871,6 +875,9 @@ func (d *DataAccessService) internal_getElClAPR(ctx context.Context, dashboardId
 	}
 	elIncomeFloat, _ := elIncome.Float64()
 	elAPR = ((elIncomeFloat / float64(aprDivisor)) / (float64(32e18) * float64(rewardsResultTable.ValidatorCount))) * 24.0 * 365.0 * 100.0
+	if math.IsNaN(elAPR) {
+		elAPR = 0
+	}
```
remoterami, line 880: could log an error here so we can debug, just to be safe

remoterami (approved):
lgtm

backend/pkg/api/data_access/vdb_summary.go
Eisei24: I would leave it like it was written for the cl part for now and not add error logging.

PR #766, opened 2024-08-22
Beds 195/upgrade primevue

MauserBitfly:
This PR
- updates the primevue version to `4.0.4`
  - Replace `Dropdown` with `Select`
  - Create new `BcTabList` component for the `Tabs`component and replacing all usages of the old `TabList`
  - Replace `InputSwitch`with `ToggleSwitch`
  - Adapted to  a lot of PrimeVue class names changes
  - Adjusted default styles (as they now have more default values) of several comonents
- Fix small translation but in `NotificationsNetworkTable`

This PR did not:
- make use of CSS variables introduced with the new version, but rather overwrites them to get to status quo


frontend/assets/css/prime.scss
```diff
@@ -25,12 +26,11 @@
 
   transition: background-color 0.2s, color 0.2s, border-color 0.2s;
 
-  &:not([data-secondary]) {
+  &:not([ddata-p-severity="secondary"]) {
```
marcel-bitfly, line 29: I think there is a typo

MauserBitfly: good catch, fixed ;-)

frontend/components/bc/toggle/MultiBarButton.vue
```diff
@@ -48,22 +39,12 @@ const icon = computed(() => {
       </div>
     </template>
     <ToggleButton
-      v-model="selected"
-      class="bc-toggle"
-      :class="readonlyClass"
-      on-label="''"
-      off-icon="''"
+      v-model="selected" class="bc-toggle" :class="readonlyClass"
```
marcel-bitfly, line 42: nitpick: why is this all in one line now?

frontend/package.json
```diff
@@ -15,7 +15,7 @@
     "lodash-es": "^4.17.21",
     "luxon": "^3.4.4",
     "pinia": "^2.1.7",
-    "primevue": "3.46.0",
+    "primevue": "^4.0.4",
```
marcel-bitfly, line 18: should we pin this version?

frontend/types/hashTabs.ts
```diff
@@ -1,5 +1,12 @@
+import type { IconDefinition } from '@fortawesome/fontawesome-svg-core'
+import type { Component } from 'vue'
+
 export type HashTab = {
+  component?: Component,
   disabled?: boolean,
-  index: number,
+  icon?: IconDefinition,
+  index: string,
```
marcel-bitfly, line 8: this should be a number

marcel-bitfly, line 8:
```suggestion
  index: number,
```
this should stay a number

frontend/types/hashTabs.ts
```diff
@@ -1,5 +1,12 @@
...
+  placeholder?: string,
+  title?: string,
 }
 export type HashTabs = Record<string, HashTab>
```
marcel-bitfly, line 12: couldn't that be an array instead of an object, then we would spare us the struggle of sorting this

MauserBitfly: originally it was easier with a Record to map the url hash ... but now that the PrimeVue tabs no longer use an index but a string as key it's much easier and cleaner with an array.

frontend/types/hashTabs.ts
MauserBitfly: renamed to key:string and refactored the whole thing once more

frontend/components/bc/tab/BcTabList.vue
```diff
@@ -0,0 +1,61 @@
...
+
+const sortedTabs = computed(() => orderBy(Object.entries(props.tabs).map(([
+  key,
+  tab,
+]) => {
+  return {
+    key,
+    ...tab,
+  }
+}), tab => parseInt(tab.index)))
```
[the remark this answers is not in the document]
MauserBitfly: sorting no longer needed

marcel-bitfly (approved):
nice work...thanks for taking over this hard task 💪

PR #767, opened 2024-08-22
fix(dashboard): label the time range overlay as date

MauserBitfly:
This PR:
- labels the reward chart overlay as a date instead of the epoch

image

BEDS-309


benji-bitfly (approved):
LGTM :heavy_check_mark:

PR #769, opened 2024-08-22
fix: tootlip sizing and and border in validator list modal

MauserBitfly:
This Pr:
- fixes the tooltip sizing
- borders around the groups in the validator dialog


benji-bitfly (approved):
lgtm :ocean:

PR #771, opened 2024-08-23
Beds 157/add balance to overview

Eisei24:


backend/pkg/api/data_access/vdb_management.go
```diff
@@ -343,56 +341,109 @@ func (d *DataAccessService) GetValidatorDashboardOverview(ctx context.Context, d
 		})
 	}
 
-	// Validator Status
+	// Validator status and balance
 	wg.Go(func() error {
-		var query string
```
Eisei24, line 348: No need to get the validator status from the db if we have it in the cache.

backend/pkg/api/data_access/vdb_management.go
```diff
@@ -343,56 +341,111 @@ func (d *DataAccessService) GetValidatorDashboardOverview(ctx context.Context, d
...
+				data.Balances.Total = data.Balances.Total.Add(validatorBalance)
+
+				nonRpDashboardId.Validators = append(nonRpDashboardId.Validators, validator)
+			}
+			data.Balances.Effective = data.Balances.Effective.Add(effectiveBalance)
+		}
+
+		// Get the total cl deposits for non-rocketpool validators
+		if len(nonRpDashboardId.Validators) > 0 {
+			totalNonRpDeposits, err := d.GetValidatorDashboardTotalClDeposits(ctx, nonRpDashboardId)
```
Eisei24, line 442: I assumed `StakedEth` means CL deposits for non-rocketpool validators.
If that is not the case please comment here.

remoterami (changes requested):
I'd do some small changes, ptal

invis-bitfly (approved):

guybrush (approved):

remoterami (approved):

PR #772, opened 2024-08-26
(BEDS-345) move healthz queries to data access

LuccaBitfly:


invis-bitfly (approved):
code changes lgtm, did not test if it actually works

PR #773, opened 2024-08-26
fix(mainHeader): `touch target`s were reduced to the `text` instead o…

marcel-bitfly:
…f the `line`

See: BEDS-346

### DIFF
--- a/frontend/components/bc/header/MainHeader.vue
+++ b/frontend/components/bc/header/MainHeader.vue
@@ -68,20 +68,20 @@ const toggleMegaMenu = (evt: Event) => {
 
 const isMobileMegaMenuOpen = computed(() => megaMenu.value?.isMobileMenuOpen)
 
-const userMenu = computed(() => {
-  return [
-    {
-      command: async () => {
-        await navigateTo('../user/settings')
-      },
-      label: $t('header.settings'),
-    },
-    {
-      command: () => doLogout(),
-      label: $t('header.logout'),
-    },
-  ]
-})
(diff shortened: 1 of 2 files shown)


frontend/components/bc/header/MainHeader.vue
```diff
@@ -68,20 +68,25 @@ const toggleMegaMenu = (evt: Event) => {
...
+const userMenu: UserMenuItem[] = [
+  {
+    label: labelSettings,
+  },
+  {
+    label: labelLogout,
+  },
+]
+const handleUserMenuSelect = async (value: UserMenuItem) => {
+  if (value.label === labelSettings) {
```
MauserBitfly, line 83: I am not sure if this solution is salable if we have more options and also using the translation as Identifier is not the best choice I think.  

Here is a suggestions:
```suggestion
type UserMenuItem = { command?: () => (Promise<void> | void), label: string }
const userMenu: UserMenuItem[] = [
  {
    command: async () => { await navigateTo('/user/settings') },
    label: $t('header.settings'),
  },
  {
    command: () => doLogout(),
    label: $t('header.logout'),
  },
]
const handleUserMenuSelect = async (value: UserMenuItem) => {
  await value.command?.()
}
```

MauserBitfly (commented):
Works fine, but I left a suggestion for better scalability.

frontend/components/bc/header/MainHeader.vue
marcel-bitfly: I am following yr suggestion...thx

MauserBitfly (approved):
LGTM

PR #774, opened 2024-08-26
(BEDS-347) refactor dummy

LuccaBitfly:
refactors dummy service to use generic func to avoid code duplication


remoterami (approved):
lgtm

PR #775, opened 2024-08-26
fix(dashboard): deduplicate blocks

peterbitfly:


remoterami:
This does not solve the problem in all cases, see e.g. dashboard ID `v-839509ee-bd2d-43cd-aa5a-032c82ea0be3`
The underlying issue is coming from the `relays_blocks` table: when multiple relays report the same block, the left join causes multiple rows to appear. However it's also possible that they report the same block but with a different bribe, so the `group by` won't suffice (that's the example from above).

It's kind of impossible to know for sure which relay offer was accepted, but I suggest to pick the maximum as a best guess; I'll push this suggestion, ptal.

Note: I think this should be true for multiple other usages of that table, too, namely all the tables related to rewards. Any duplicated entries result in counting the relay rewards too often.

PR #777, opened 2024-08-26
fix(qrCode): use shorter url syntax for dashboards with many `valid… 

marcel-bitfly:
…ators`

`validator ids` get encoded in `base64`.
 Which triggerd the `data capacity limit` of the `qr library`.

 So we use the `unencoded url version` (?validators=<validatorID or validatorIndex>),
 as a workaround.

See: BEDS-350


MauserBitfly (approved):
LGTM

PR #782, opened 2024-08-27
fix: multi + select primevue update styles

MauserBitfly:
This PR:
- fixes the Styles after the PrimeVue Update for 
  - Select
  - MultiSelect 

image


marcel-bitfly:
nitpick: commit-msg could be prettier

PR #786, opened 2024-08-27
fix(table): fix interaction of table pager buttons on big screen

MauserBitfly:
This PR:
- makes sure that the last row does not prevent the table pager buttons to be clickable.


marcel-bitfly (approved):
👍

PR #791, opened 2024-08-28
chore: add `self signed` ssl certificates

marcel-bitfly:
When we create our own `certs`, which have a `validity` of one year. We do not have to `click proceed` everytime we restart the `dev server`


MauserBitfly (approved):
LGTM

PR #803, opened 2024-08-29
(BEDS-374) validator list changed to body param

remoterami:


backend/pkg/api/handlers/internal.go
LuccaBitfly, line 1: Removing the ability to delete all validators in a previous PR and now moving the validators param to the body (not really done for `DELETE` method) introduces arbitrary un-RESTful elements to the endpoint. I'd like to change the route to `POST /validator-dashboards/{id}/validators/bulk-deletions` retain some RESTfullness.

backend/pkg/api/handlers/internal.go
```diff
@@ -824,12 +824,17 @@ func (h *HandlerService) InternalDeleteValidatorDashboardValidators(w http.Respo
...
-			return
-		}
+	req := struct {
+		Validators []intOrString `json:"validators"`
+	}{}
+	if err := v.checkBody(&req, r); err != nil {
+		handleErr(w, r, err)
+		return
+	}
+	indices, publicKeys = v.checkValidators(req.Validators, allowEmpty)
```
LuccaBitfly, line 834: should not allow empty param

LuccaBitfly (changes requested):

PR #816, opened 2024-09-02
Adjusted latest proposal check

Eisei24:


backend/pkg/api/data_access/vdb_slotviz.go
```diff
@@ -35,6 +34,23 @@ func (d *DataAccessService) GetValidatorDashboardSlotViz(ctx context.Context, da
...
+		} else {
+			// No more data available
+			break
+		}
+
+		if slot == 0 {
+			break
+		}
+	}
+
```
invis-bitfly, line 53: thoughts about moving this to the slot viz service so we dont have to replicate this when something else needs the same logic? maybe a new `LatestProposedSlot` field in the `SyncData` struct?

Eisei24: Sure, it's just one extra value in the cache.

PR #821, opened 2024-09-03
Set HTML attributes.

benji-bitfly:
Set HTML attributes to prevent popups from third-party apps.

BEDS-368


benji-bitfly:
> does fix the problem for me (side note, the [linter complains about the order of the attributes](https://github.com/gobitfly/beaconchain/pull/821/files))

fixed. :raised_hands:

PR #822, opened 2024-09-03
Beds 401/api integration tests

peterbitfly:
Tests can be run with `go test -v -cover ./pkg/api/handlers/* -config <path_to_config_file>`


backend/pkg/api/handlers/handlers_test.go
```diff
@@ -0,0 +1,288 @@
...
+
+	if err := goose.Up(tempDb.DB, "../../../pkg/commons/db/migrations/postgres"); err != nil {
+		log.Fatal(err, "error running migrations", 0)
+	}
+
+	// insert dummy user for testing (email: admin@admin, password: admin)
+	pHash, _ := bcrypt.GenerateFromPassword([]byte("admin"), 10)
+	_, err = tempDb.Exec(`
+      INSERT INTO users (password, email, register_ts, api_key, email_confirmed)
+      VALUES ($1, $2, TO_TIMESTAMP($3), $4, $5)`,
```
LuccaBitfly, line 126: Could insert this test user with some large number as user ID, which would make sure that new dashboards created for this user never interfere with real user dashboards.

backend/pkg/api/handlers/handlers_test.go
```diff
@@ -0,0 +1,288 @@
...
+	// check if email is censored
+	assert.Equal(t, meResponse.Data.Email, "a***n@a***n.com", "email should be a***n@a***n.com")
+
+	// check if logout works
+	code, _ = ts.post(t, "/api/i/logout", bytes.NewBufferString(``))
+	assert.Equal(t, http.StatusOK, code, "logout should be successful")
+
+	// check if user is logged out
+	code, _ = ts.get(t, "/api/i/users/me")
+	assert.Equal(t, http.StatusUnauthorized, code, "call to users/me should be unauthorized")
```
LuccaBitfly, line 224: Minor: I believe for cases like these, where you make multiple requests to the same endpoint, it's common practice to structure each request into it's own `t.Run`, e.g.
```go
    t.Run("login with email in wrong format", func(t *testing.T) {
        code, body := ts.post(t, "/api/i/login", bytes.NewBuffer([]byte(`{"email": "admin", "password": "admin"}`)))
        assert.Equal(t, http.StatusBadRequest, code)
        resp := ts.parseErrorResonse(t, body)
        assert.Equal(t, "email: given value 'admin' has incorrect format", resp.Error, "unexpected error message")
    })

    t.Run("login with correct user and wrong password", func(t *testing.T) {
        code, body := ts.post(t, "/api/i/login", bytes.NewBufferString(`{"email": "admin@admin.com", "password": "wrong"}`))
        assert.Equal(t, http.StatusUnauthorized, code, "login should not be successful")
        resp := ts.parseErrorResonse(t, body)
        assert.Equal(t, "unauthorized: invalid email or password", resp.Error, "unexpected error message")
    })
    // ...
```

backend/pkg/api/handlers/handlers_test.go
```diff
@@ -0,0 +1,288 @@
...
+}
+
+func TestInternalSearchHandler(t *testing.T) {
+	// search for validator with index 5
+	code, body := ts.post(t, "/api/i/search", bytes.NewBufferString(`{"input":"5","networks":[17000],"types":["validators_by_deposit_ens_name","validators_by_deposit_address","validators_by_withdrawal_ens_name","validators_by_withdrawal_address","validators_by_withdrawal_credential","validator_by_index","validator_by_public_key","validators_by_graffiti"]}`))
```
LuccaBitfly, line 229: I'd prefer if we properly formatted these long JSON strings for better readability.

backend/pkg/api/handlers/handlers_test.go
```diff
@@ -0,0 +1,288 @@
...
+		t.Fatal(err)
+	}
+	defer rs.Body.Close()
+	body, err := io.ReadAll(rs.Body)
+	if err != nil {
+		t.Fatal(err)
+	}
+	bytes.TrimSpace(body)
+	return rs.StatusCode, string(body)
+}
```
LuccaBitfly, line 67: Can extract common logic:
```suggestion
func (ts *testServer) request(t *testing.T, method, urlPath string, data io.Reader) (int, string) {
	req, err := http.NewRequest(method, ts.URL+urlPath, data)
	if err != nil {
		t.Fatal(err)
	}
	rs, err := ts.Client().Do(req)
	if err != nil {
		t.Fatal(err)
	}
	defer rs.Body.Close()
	body, err := io.ReadAll(rs.Body)
	if err != nil {
		t.Fatal(err)
	}
	bytes.TrimSpace(body)
	return rs.StatusCode, string(body)
}

func (ts *testServer) get(t *testing.T, method, urlPath string) (int, string) {
	return ts.request(t, http.MethodGet, urlPath, nil)
}

func (ts *testServer) post(t *testing.T, urlPath string, data io.Reader) (int, string) {
	return ts.request(t, http.MethodPost, urlPath, data)
}
```

LuccaBitfly (changes requested):
looks very nice to me so far, some small comments though

backend/pkg/api/handlers/handlers_test.go
```diff
@@ -0,0 +1,288 @@
...
+	cfg.Frontend.ReaderDatabase.Password = "postgres"
+	cfg.Frontend.ReaderDatabase.Username = "postgres"
+
+	cfg.Frontend.WriterDatabase.Host = "localhost"
+	cfg.Frontend.WriterDatabase.Port = "5432"
+	cfg.Frontend.WriterDatabase.Name = "postgres"
+	cfg.Frontend.WriterDatabase.Password = "postgres"
+	cfg.Frontend.WriterDatabase.Username = "postgres"
+
+	utils.Config = cfg
```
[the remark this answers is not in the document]
peterbitfly: can be set in the provided config file

LuccaBitfly (approved):
lgtm

PR #825, opened 2024-09-04
💄Temporary Fix: Resolve Display Shift and Background Color Issue When Opening Modal

benji-bitfly:
fix: :lipstick: Add temporary workaround to fix shifting display issue and background color when opening the modal.

BEDS-406


LuccaBitfly (approved):
utAck

PR #838, opened 2024-09-06
feat(dashboard): add basic integration tests

peterbitfly:
requires playwright to be installed and a local instance of the frontend to be running at port `3000`. tests can be run with `npm run test`.

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -22,9 +22,11 @@
   },
   "devDependencies": {
     "@nuxt/eslint": "^0.3.13",
+    "@nuxt/test-utils": "^3.14.1",
     "@nuxtjs/color-mode": "^3.3.2",
     "@nuxtjs/i18n": "^8.0.0",
     "@nuxtjs/style-resources": "^1.2.2",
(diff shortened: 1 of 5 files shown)


marcel-bitfly:
nitpick `commit msg`: 

test(dashboard): add basic `e2e`-tests

frontend/package.json
```diff
@@ -22,9 +22,11 @@
   },
   "devDependencies": {
     "@nuxt/eslint": "^0.3.13",
+    "@nuxt/test-utils": "^3.14.1",
     "@nuxtjs/color-mode": "^3.3.2",
     "@nuxtjs/i18n": "^8.0.0",
     "@nuxtjs/style-resources": "^1.2.2",
+    "@playwright/test": "^1.47.0",
```
marcel-bitfly, line 29: dunno whats best, but just that you have seen it, could have also gone with `playwright-core` here 🤷

marcel-bitfly (approved):
thanks for making a start.

In the next step we should wrap `expect` and `test` with our own `utility` function

PR #841, opened 2024-09-09
fix(dashboard): source head epoch from slot viz duties response

peterbitfly:
sources the head epoch from the slot viz duties response to avoid data races when setting the `latestEpoch` cached variable


Eisei24 (approved):
lgtm

Eisei24 (approved):
lgtm

PR #847, opened 2024-09-09
Beds 97/add notifications

marcel-bitfly:


frontend/components/dashboard/DashboardHeader.vue
```diff
@@ -90,19 +90,18 @@ const items = computed<MenuBarEntry[]>(() => {
...
-  const disabledTooltip = !showInDevelopment ? $t('common.coming_soon') : undefined
-  const onNotificationsPage = dashboardType.value === 'notifications'
+  const disabledTooltip = !has('feature-notifications') ? $t('common.coming_soon') : undefined
   addToSortedItems($t('notifications.title'), [ {
-    active: onNotificationsPage,
+    active: route.name === 'notifications',
     disabledTooltip,
     label: $t('notifications.title'),
-    route: !onNotificationsPage ? '/notifications' : undefined,
+    route: '/notifications',
```
benji-bitfly, line 104: Please leave the logic like that otherwise the route will be replaced, resulting in a lost `hash value`  eg. `/notifications#clients` -> `/notifications` 

```suggestion
    route: route.name === 'notifications' ? undefined  : '/notifications',
```

marcel-bitfly: good point , will do thx 💪

PR #873, opened 2024-09-19
(BEDS-481) API: dashboards notifications struct adjustments

remoterami:


frontend/utils/wrapWithIdentifier.ts
```diff
@@ -0,0 +1,19 @@
...
+    return
+  }
+  return {
+    data: response.data.map(row => ({
+      ...row,
+      wrapped_identifier: keys.map(key => row[key]).join('-'),
+    })),
+    paging: response.paging,
+  }
+}
```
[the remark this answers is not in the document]
marcel-bitfly: maybe think about generalize this function (is it important for the function that there has to be a row?)

frontend/components/notifications/DashboardsTable.vue
```diff
@@ -61,8 +61,8 @@ const { overview } = useNotificationsDashboardOverviewStore()
       <template #table>
         <ClientOnly fallback-tag="span">
           <BcTable
-            :data="notificationsDashboards"
-            data-key="notification_id"
+            :data="wrapWithIdentifier(notificationsDashboards, 'is_account_dashboard', 'dashboard_id', 'group_id', 'epoch')"
+            data-key="wrapped_identifier"
```
marcel-bitfly, line 65: when using `<BcTable>` and using `wrapedWithIdentifier` function, which has the consequence that `data-key` would be always the `wrapped_identifier` key...

It would be nice if `BcTable` would be refactored that `data-key` is not needed anymore

LuccaBitfly: Do we want to refactor `BcTable` to do that though? As far as the prime vue component goes, `data-key` is optional and only optimizes performance, which may be lost if we wrap every time when using `BcTable`. I would suggest investigating whether `BcTable` truly needs a unique ID first.

frontend/utils/wrapWithIdentifier.ts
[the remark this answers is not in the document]
LuccaBitfly: Pushed the changes, ptal.
As for the generalization, since this func takes `ApiPagingResponse`s, which only concern tables, the function can expect there to be rows. If we want to add identifiers so something other than `ApiPagingResponse`s, we can think about generalizing more.

marcel-bitfly (approved):
👍

frontend/utils/addIdentifier.ts
```diff
@@ -0,0 +1,18 @@
...
+    return
+  }
+  return {
+    data: response.data.map(row => ({
+      ...row,
+      identifier: keys.map(key => row[key]).join('-'),
+    })),
+    paging: response.paging,
+  }
+}
```
marcel-bitfly, line 18: np: this should work too and is a bit cleaner imo
```suggestion
export function addIdentifier<T>(
  pagingResponse: ApiPagingResponse<T> | undefined,
  keys: (keyof T)[],
) {
  if (!pagingResponse) {
    return
  }
  return {
    ...pagingResponse,
    data: pagingResponse.data
      .map(item => ({
        ...item,
        identifier: keys.map(key => item[key]).join('-'),
      })),
  }
}
```

marcel-bitfly: But again, imo this is a `code smell` and could potentially introduce `performance issues` in large data sets.
As this will be done for every row in a table.

I still think `data manipulation` should happen in `backend`.

PR #893, opened 2024-09-27
(BEDS-469) DA: notifications dashboard table

remoterami:


backend/pkg/api/data_access/notifications.go
```diff
@@ -34,8 +42,169 @@ type NotificationsRepository interface {
...
+	}
+
+	// search
+	searchName := regexp.MustCompile(`^[a-zA-Z0-9_\-.\ ]+$`).MatchString(search)
+	if searchName {
+		searchLower := strings.ToLower(strings.Replace(search, "_", "\\_", -1)) + "%"
+		unionQuery = unionQuery.Where(exp.NewExpressionList(
+			exp.OrType,
+			goqu.L("LOWER(?)", goqu.I("dashboard_name")).Like(searchLower),
+			goqu.L("LOWER(?)", goqu.I("group_name")).Like(searchLower),
```
Eisei24, line 158: **suggestion:** You could use `ILIKE` so you don't need to lower the strings.

backend/pkg/api/data_access/notifications.go
```diff
@@ -34,8 +42,169 @@ type NotificationsRepository interface {
...
+		))
+	}
+	unionQuery = unionQuery.Limit(uint(limit + 1))
+
+	query, args, err := unionQuery.ToSQL()
+	if err != nil {
+		return nil, nil, err
+	}
+	//err = d.alloyReader.SelectContext(ctx, &response, query, args...)
```
Eisei24, line 167: **question:** Why not use this? Is it a `pq.Array` conflict in `NULL` cases or something like that?

Eisei24 (changes requested):
Looks good but please have a look at the cursor issue.

frontend/types/api/notifications.ts
```diff
@@ -37,10 +37,10 @@ export type InternalGetUserNotificationsResponse = ApiDataResponse<NotificationO
 export interface NotificationDashboardsTableRow {
   is_account_dashboard: boolean; // if false it's a validator dashboard
   chain_id: number /* uint64 */;
-  timestamp: number /* int64 */;
+  epoch: number /* uint64 */;
   dashboard_id: number /* uint64 */;
+  group_id: number /* uint64 */;
```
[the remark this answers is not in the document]
remoterami: Yes, discussed with API: the groups mapping is only present & cached in validator/account dashboards, not the notifications dashboard, so we need to provide group names. The dashboard names mapping is present though so we don't need them, even though it wouldn't have hurt too much

backend/pkg/api/data_access/general.go
```diff
@@ -46,3 +48,55 @@ func (d *DataAccessService) GetNamesAndEnsForAddresses(ctx context.Context, addr
 	}
 	return nil
 }
+
+// helper function to sort and apply pagination to a query
+// 1st param is the list of all columns necessary to sort the table deterministically; it defines their precedence and sort direction
+// 2nd param is the requested sort column; it may or may not be part of the default columns
+func applySortAndPagination(defaultColumns []types.SortColumn, primary types.SortColumn, cursor types.GenericCursor) ([]exp.OrderedExpression, exp.Expression) {
```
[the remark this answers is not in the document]
remoterami: Yeah thought this would help us going forward, glad you like it! I'm also planning to refactor some queries with this.

Feel free to adjust/extend if useful :+1:

backend/pkg/api/data_access/notifications.go
remoterami: According to my tests a while back `LOWER` seems to be slightly faster, you can also find similar results online (e.g. https://stackoverflow.com/a/20336666). Maybe a preference thing but I'd keep it for that reason

backend/pkg/api/data_access/notifications.go
remoterami: My initial motivation was that I didn't want to use `pq.StringArray` (because `string` won't work with `ARRAY_AGG`) in the struct type definition because I feared it would mess with the ts struct conversion, but turns out that's not the case - it's way cleaner without `rows.Scan` so I've changed it, thanks for pointing out

backend/pkg/api/handlers/common.go
```diff
@@ -687,6 +687,9 @@ func (v *validationError) checkNetworkParameter(param string) uint64 {
 }
 
 func (v *validationError) checkNetworksParameter(param string) []uint64 {
+	if param == "" {
```
peterbitfly, line 690: @LuccaBitfly is this change ok with you?

backend/pkg/api/types/notifications.go
```diff
@@ -29,14 +32,15 @@ type InternalGetUserNotificationsResponse ApiDataResponse[NotificationOverviewDa
...
-	ChainId            uint64   `json:"chain_id"`
-	Epoch              uint64   `json:"epoch"`
-	DashboardId        uint64   `json:"dashboard_id"`
-	GroupId            uint64   `json:"group_id"`
-	GroupName          string   `json:"group_name"`
-	EntityCount        uint64   `json:"entity_count"`
-	EventTypes         []string `json:"event_types" tstype:"('validator_online' | 'validator_offline' | 'group_online' | 'group_offline' | 'attestation_missed' | 'proposal_success' | 'proposal_missed' | 'proposal_upcoming' | 'max_collateral' | 'min_collateral' | 'sync' | 'withdrawal' | 'validator_got_slashed' | 'validator_has_slashed' | 'incoming_tx' | 'outgoing_tx' | 'transfer_erc20' | 'transfer_erc721' | 'transfer_erc1155')[]" faker:"slice_len=2, oneof: validator_online, validator_offline, group_online, group_offline, attestation_missed, proposal_success, proposal_missed, proposal_upcoming, max_collateral, min_collateral, sync, withdrawal, validator_got_slashed, validator_has_slashed, incoming_tx, outgoing_tx, transfer_erc20, transfer_erc721, transfer_erc1155"`
+	IsAccountDashboard bool           `db:"is_account_dashboard" json:"is_account_dashboard"` // if false it's a validator dashboard
```
peterbitfly, line 35: @LuccaBitfly is this change ok with you?

peterbitfly (commented):
waiting for green light regarding the changes in the api package, if thats ok the pr can be merged

backend/pkg/api/types/data_access.go
```diff
@@ -125,6 +132,15 @@ type BlocksCursor struct {
 	Reward   decimal.Decimal
 }
 
+type NotificationsDashboardsCursor struct {
+	GenericCursor
+
+	Epoch         uint64
+	ChainId       uint64
+	DashboardName string
+	GroupName     string
```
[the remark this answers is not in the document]
Eisei24: Yes, in the cursor I would have all 4 of them for an easier query.

backend/pkg/api/data_access/notifications.go
```diff
@@ -146,23 +155,12 @@ func (d *DataAccessService) GetDashboardNotifications(ctx context.Context, userI
 	defaultColumns := []t.SortColumn{
```
Eisei24, line 181: **thought:** Since the default column entries only require those that uniquely identify a row you don't need to add the dashboard and group names to it.
However if you change it this way you would need the `offset` if again.

Both ways work for me but consider that the query could be faster with less WHERE/ORDER BY.

Eisei24 (approved):
utACK, please wait for the confirmation from other reviewers before merging.
I can't really tell what the linter issue here is.

PR #910, opened 2024-10-02
Beds 473/notifications set settings

Eisei24:


backend/pkg/commons/db/migrations/postgres/TBD_add_notifications_column.sql
```diff
@@ -0,0 +1,13 @@
...
+ALTER TABLE users ADD COLUMN IF NOT EXISTS notifications_do_not_disturb_ts TIMESTAMP WITHOUT TIME ZONE;
+ALTER TABLE users_subscriptions ALTER COLUMN event_threshold SET DATA TYPE NUMERIC;
+-- +goose StatementEnd
+
+-- +goose Down
+-- +goose StatementBegin
+SELECT 'down SQL query - remove column notifications_do_not_disturb_ts and revert event_threshold data type';
+ALTER TABLE users DROP COLUMN IF EXISTS notifications_do_not_disturb_ts;
+ALTER TABLE users_subscriptions ALTER COLUMN event_threshold SET DATA TYPE REAL;
+-- +goose StatementEnd
```
remoterami, line 13: reminder to coordinate that this gets applied to user db (if not done yet)

backend/pkg/api/data_access/notifications.go
```diff
@@ -62,19 +69,234 @@ func (d *DataAccessService) GetNotificationSettings(ctx context.Context, userId
...
+	return err
 }
 func (d *DataAccessService) DeleteNotificationSettingsPairedDevice(ctx context.Context, userId uint64, pairedDeviceId string) error {
-	return d.dummy.DeleteNotificationSettingsPairedDevice(ctx, userId, pairedDeviceId)
+	_, err := d.userWriter.ExecContext(ctx, `
+		DELETE FROM users_devices 
+		WHERE user_id = $1 AND device_identifier = $2`,
+		userId, pairedDeviceId)
+	return err
 }
```
remoterami, line 261: Could consider returning an error here for now if no row was updated/deleted as currently there is no proper input parameter validation in api layer (see the TODO there), just for good measure

remoterami (approved):
lgtm (just a small fyi comment/suggestion)

backend/pkg/commons/db/migrations/postgres/TBD_add_notifications_column.sql
Eisei24: Sorry, I forgot to delete this file.
The desired behaviour in the db was discussed in BEDS-483 and it will also implement it.

We will leave the threshold as REAL(float32) and I also handled everything in the code as a float.
The Gas thresholds that we get as Wei (decimal.Decimal) get converted with `.Div(1e9).InexactFloat64()`.

backend/pkg/api/data_access/notifications.go
Eisei24: Sure, I added it with a TODO that it could be deleted once the API layer has further checks.

remoterami (approved):
lgtm

PR #912, opened 2024-10-02
Beds 474/notifications dashboard settings

Eisei24:


backend/pkg/api/data_access/notifications.go
```diff
@@ -77,11 +100,534 @@ func (d *DataAccessService) UpdateNotificationSettingsClients(ctx context.Contex
...
+			case types.ValidatorReceivedWithdrawalEventName:
+				settings.IsWithdrawalProcessedSubscribed = true
+			case types.ValidatorGotSlashedEventName:
+				settings.IsSlashedSubscribed = true
+			case types.RocketpoolCollateralMinReached:
+				settings.IsMinCollateralSubscribed = true
+				settings.MinCollateralThreshold = event.Threshold
+			case types.RocketpoolCollateralMaxReached:
+				settings.IsMaxCollateralSubscribed = true
+				settings.MaxCollateralThreshold = event.Threshold
```
remoterami, line 278: I'm not quite sure if I understand correctly - the goal is to retrieve the current status of user subscriptions, right? Why are we setting everything to true here, I would expect each value to be set according to the value saved in the DB. Is this just default values? Am I missing something?

backend/pkg/api/data_access/notifications.go
```diff
@@ -77,11 +100,534 @@ func (d *DataAccessService) UpdateNotificationSettingsClients(ctx context.Contex
...
+	// 		webhook_target = NULLIF($1, ''),
+	// 		webhook_format = CASE WHEN $2 THEN $3 ELSE NULL END,
+	// 		ignore_spam_transactions = $4,
+	// 		subscribed_chain_ids = $5
+	// 	WHERE dashboard_id = $6 AND id = $7`, settings.WebhookUrl, settings.IsWebhookDiscordEnabled, DiscordWebhookFormat, settings.IsIgnoreSpamTransactionsEnabled, settings.SubscribedChainIds, dashboardId, groupId)
+	// if err != nil {
+	// 	return err
+	// }
+
+	return nil
```
remoterami, line 623:
```suggestion
	return d.dummy.UpdateNotificationSettingsAccountDashboard(ctx, userId, dashboardId, groupId, settings)
```

remoterami (changes requested):
Very clean all in all, just some remarks / questions

also fix linter please^^

backend/pkg/api/data_access/notifications.go
```diff
@@ -77,11 +100,534 @@ func (d *DataAccessService) UpdateNotificationSettingsClients(ctx context.Contex
...
+
+	wg := errgroup.Group{}
+
+	// Initialize the cursor
+	var currentCursor t.NotificationSettingsCursor
+	var err error
+	if cursor != "" {
+		currentCursor, err = utils.StringToCursor[t.NotificationSettingsCursor](cursor)
+		if err != nil {
+			return nil, nil, fmt.Errorf("failed to parse passed cursor as WithdrawalsCursor: %w", err)
```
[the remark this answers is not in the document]
Eisei24: Fixed.

backend/pkg/api/data_access/notifications.go
Eisei24: Right, it should remain as before. Fixed.

backend/pkg/api/data_access/notifications.go
```diff
@@ -77,11 +100,534 @@ func (d *DataAccessService) UpdateNotificationSettingsClients(ctx context.Contex
...
+	moreDataFlag := len(result) > int(limit)
+	if !moreDataFlag && !currentCursor.IsValid() {
+		// No paging required
+		return result, &paging, nil
+	}
+
+	// Remove the last entries from data
+	if moreDataFlag {
+		result = result[:limit]
+	}
```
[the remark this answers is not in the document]
Eisei24: Yes, I always try to get the LIMIT and ORDER BY into the query when possible.
Currently the `users_val_dashboards` could have these since we show all dashboards but it would no longer work if we have the account dashboards query in a separate routine.
Perhaps once these should be implemented and their db structure is defined we can unite them efficiently into one query and add LIMIT/ORDER BY.

However if you want the `users_val_dashboards` query adjusted now I could also do it.

backend/pkg/commons/types/frontend.go
```diff
@@ -89,9 +89,9 @@ const (
 	// Account dashboard events
 	IncomingTransactionEventName  EventName = "incoming_transaction"
 	OutgoingTransactionEventName  EventName = "outgoing_transaction"
-	ERC20TokenTransferEventName   EventName = "erc20_token_transfer"
-	ERC721TokenTransferEventName  EventName = "erc721_token_transfer"
-	ERC1155TokenTransferEventName EventName = "erc1155_token_transfer"
+	ERC20TokenTransferEventName   EventName = "erc20_token_transfer"   // #nosec G101
+	ERC721TokenTransferEventName  EventName = "erc721_token_transfer"  // #nosec G101
+	ERC1155TokenTransferEventName EventName = "erc1155_token_transfer" // #nosec G101
```
Eisei24, line 94: The linter forbade me a lot of string names here because he assumed a password leak.
I set it to ignore it but this would also be a possibility I don't really like:
```	
tokenTransfer                 EventName = "token_transfer"
ERC20TokenTransferEventName   EventName = "erc20_" + tokenTransfer
ERC721TokenTransferEventName  EventName = "erc721_" + tokenTransfer
ERC1155TokenTransferEventName EventName = "erc1155_" + tokenTransfer

File: backend/pkg/api/data_access/notifications.go (hunk 4 above)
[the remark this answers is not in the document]
Comment (remoterami) [reply backend/pkg/api/data_access/notifications.go:1543]: Sounds like a good approach, don't think fully optimizing this particular one is too important for now. So agreed, might do it when we add the account dashboards

File: backend/pkg/commons/types/frontend.go (hunk 5 above)
Comment (remoterami) [reply backend/pkg/commons/types/frontend.go:83]: The ignore is fine

File: backend/pkg/api/data_access/notifications.go [hunk 6]
```diff
@@ -1184,13 +1196,527 @@ func (d *DataAccessService) UpdateNotificationSettingsClients(ctx context.Contex
...
+			accSettings.IsIgnoreSpamTransactionsEnabled = accDashboard.IsIgnoreSpamTransactionsEnabled
+			accSettings.SubscribedChainIds = accDashboard.SubscribedChainIds
+		}
+	}
+
+	// Apply filter
+	if search != "" {
+		lowerSearch := strings.ToLower(search)
+		for key, setting := range settingsMap {
+			if lowerSearch != strings.ToLower(setting.DashboardName) && lowerSearch != strings.ToLower(setting.GroupName) {
```
remoterami, line 1458: here

remoterami (changes requested):
The changes lgtm, just one more thing: According to internal discussion we should apply prefix matching for group/dashboard names instead of full matches, please adapt

backend/pkg/api/data_access/notifications.go
Eisei24: Done.

remoterami (approved):
lgtm

PR #939, opened 2024-10-10
refactor(execution_deposits_exporter): export deposits up to head and…

guybrush:
* export deposits up to head
    * handle reorgs by reexporting blocks since lastExportedFinalizedBlock
    * persist lastExportedFinalizedBlock in redis
    * refactor export-logic from multiple event-based goroutines with canceling to single loop
    * aggregated deposits and cached view are only updated when export is on head
    * avoid fetching old blocks if no deposits are on the chain
    
    BEDS-585

### DIFF
--- a/backend/pkg/exporter/modules/execution_deposits_exporter.go
+++ b/backend/pkg/exporter/modules/execution_deposits_exporter.go
@@ -21,12 +21,14 @@ import (
 	gethrpc "github.com/ethereum/go-ethereum/rpc"
 	"github.com/go-redis/redis/v8"
 	"golang.org/x/exp/maps"
+	"golang.org/x/sync/errgroup"
 
 	"github.com/gobitfly/beaconchain/pkg/commons/contracts/deposit_contract"
 	"github.com/gobitfly/beaconchain/pkg/commons/db"
 	"github.com/gobitfly/beaconchain/pkg/commons/log"
 	"github.com/gobitfly/beaconchain/pkg/commons/metrics"
 	"github.com/gobitfly/beaconchain/pkg/commons/rpc"
+	"github.com/gobitfly/beaconchain/pkg/commons/services"
 	"github.com/gobitfly/beaconchain/pkg/commons/types"
 	"github.com/gobitfly/beaconchain/pkg/commons/utils"
 	constypes "github.com/gobitfly/beaconchain/pkg/consapi/types"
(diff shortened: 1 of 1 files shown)


backend/pkg/exporter/modules/execution_deposits_exporter.go
```diff
@@ -145,89 +158,115 @@ func (d *executionDepositsExporter) OnChainReorg(event *constypes.StandardEventC
...
-	if err != nil {
-		return err
-	}
+	return nil // nop
+}
 
+func (d *executionDepositsExporter) OnHead(event *constypes.StandardEventHeadResponse) (err error) {
 	var nearestELBlock sql.NullInt64
-	err = db.ReaderDb.Get(&nearestELBlock, "select exec_block_number from blocks where slot <= $1 and exec_block_number > 0 order by slot desc limit 1", res.Data.Finalized.Epoch*utils.Config.Chain.ClConfig.SlotsPerEpoch)
+	err = db.ReaderDb.Get(&nearestELBlock, "select exec_block_number from blocks where exec_block_number > 0 order by slot desc limit 1")
```
invis-bitfly, line 167: this is really meh. it made a bit more sense when we were following the finalized checkpoints (we had to resolve the slot to a exec block somehow) but even that wasnt that clean (you can resolve slot => exec block num using [`/eth/v2/beacon/blocks/{block_id}`](https://ethereum.github.io/beacon-APIs/#/Beacon/getBlockV2) `data.message.body.execution_payload.block_number` )

i would rather 
- listen to block events from erigon by adding a new `OnExecutionHead` listener to the modular exporter system
- resolve the slot to a exec block hash using the beacon api as described above
- do this query in the actual exporting loop below (makes no difference doing it here or there, besides having to store it variable when doing the former), though I'm not a big fan of having exporters depend on other exporters because this makes triaging hard and can cause compounding delays

invis-bitfly (changes requested):

backend/pkg/exporter/modules/execution_deposits_exporter.go
```diff
@@ -271,8 +313,13 @@ func (d *executionDepositsExporter) fetchDeposits(fromBlock, toBlock uint64) (de
...
 		}
 
-		depositLog :=
-			depositLogIterator.Event
+		depositLog := depositLogIterator.Event
+
+		if d.exportedTxsCache.Contains(fmt.Sprintf("%x", depositLog.Raw.TxHash.Bytes())) {
+			log.Debugf("skipping already exported deposit-tx: block: %d, tx: %x", depositLog.Raw.BlockNumber, depositLog.Raw.TxHash.Bytes())
+			continue
+		}
```
invis-bitfly, line 321: this is wrong and very dangerous. i meant only caching the validity of the signature, not skipping seen txns

the current solution can result in the following issue:
- there are 3 deposit txns on chain, each with one deposit: `txnA:0`, `txnB:1`, `txnC:2` (`:n` is the merkle tree index)
  - **`txnA:0`, `txnB:1`, `txnC:2`** gets pushed to the db
- the chain gets reorged and a new txn (`txnD`) with 1 deposit get inserted after `txnA`
  - there are now 4 deposits on chain: `txnA:0`, `txnD:1`, `txnB:2`, `txnC:3`
  - the lru cache skips the re-processing of txnB & txnC
  -  **`txnD:1`** gets pushed to the db
- a new txn with a single deposit gets added (`txnE`)
  - **`txnE:4`** gets pushed to the db

the result db state would be
-  `txnA:0`, `txnD:1`, `txnC:2`, `txnE:4`

instead of 
- `txnA:0`, `txnD:1`, **`txnB:2`, `txnC:3`**, `txnE:4`

invis-bitfly (changes requested):

backend/pkg/exporter/modules/execution_deposits_exporter.go
guybrush: removed cache

backend/pkg/exporter/modules/execution_deposits_exporter.go
```diff
@@ -163,59 +165,67 @@ func (d *executionDepositsExporter) OnFinalizedCheckpoint(event *constypes.Stand
...
+		return nil
+	})
+	g.Go(func() error {
+		finSlot, err := d.CL.GetSlot("finalized")
+		if err != nil {
+			return fmt.Errorf("error getting finalized-slot: %w", err)
+		}
+		finBlock = finSlot.Data.Message.Body.ExecutionPayload.BlockNumber
+		return nil
+	})
```
invis-bitfly, line 203: this could be done in `onHead` & `onFinalizedCheckpoint` to follow the event driven approach

but since `export()` is in a fixed loop i would not bother with it rn since it would yield no benefit

invis-bitfly (approved):
lgtm

PR #952, opened 2024-10-15
(BEDS-561) adapt return struct

remoterami:
Added the `dashboard_name`, moved it to the root of the struct and removed all references to `dashboard_id` as the entire request is dashboard specific already


frontend/types/api/notifications.ts
```diff
@@ -45,31 +45,27 @@ export interface NotificationDashboardsTableRow {
...
 export interface NotificationEventValidatorBackOnline {
   index: number /* uint64 */;
   epoch_count: number /* uint64 */;
 }
 export interface NotificationValidatorDashboardDetail {
+  dashboard_name: string;
   validator_offline: number /* uint64 */[]; // validator indices
-  group_offline: NotificationEventGroup[]; // TODO not filled yet
-  proposal_missed: IndexBlocks[];
+  group_offline: string[]; // TODO not filled yet
```
marcel-bitfly, line 59: I do not get why this type is `broader` now?

frontend/types/api/notifications.ts
```diff
@@ -45,31 +45,27 @@ export interface NotificationDashboardsTableRow {
...
+  proposal_missed: IndexSlots[];
   proposal_done: IndexBlocks[];
-  upcoming_proposals: IndexBlocks[];
+  upcoming_proposals: IndexSlots[];
```
marcel-bitfly, line 62: love this thx 💪

marcel-bitfly:
np: `commit messages` could be splitted and follow https://www.conventionalcommits.org/en/v1.0.0/#summary

backend/pkg/api/data_access/notifications.go
```diff
@@ -253,7 +253,7 @@ func (d *DataAccessService) GetNotificationOverview(ctx context.Context, userId
 
 func (d *DataAccessService) GetDashboardNotifications(ctx context.Context, userId uint64, chainIds []uint64, cursor string, colSort t.Sort[enums.NotificationDashboardsColumn], search string, limit uint64) ([]t.NotificationDashboardsTableRow, *t.Paging, error) {
 	// dev hack; TODO remove
```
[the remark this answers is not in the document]
remoterami: Resolved via #956 (requires `is_mock_enabled` to be set)

frontend/types/api/notifications.ts
remoterami: As stated above: because the dashboard name is a static part of the request and doesn't have to be part of each list item (same for group name; also see internal discussion)

backend/pkg/api/data_access/notifications.go
```diff
@@ -419,38 +417,45 @@ func (d *DataAccessService) GetValidatorDashboardNotificationDetails(ctx context
...
+		notificationDetails.GroupName = t.DefaultGroupName
 	}
-
-	buf := bytes.NewBuffer(result)
-	gz, err := gzip.NewReader(buf)
-	if err != nil {
-		return nil, err
+	if notificationDetails.DashboardName == "" {
+		notificationDetails.DashboardName = t.DefaultDashboardName
 	}
```
Eisei24, line 442: **question:** Why do we set the names to default here and don't leave them as empty strings?
It should also be impossible to create a dashboard or group without a name with at least length 1.

Eisei24 (approved):
lgtm, just one question.

backend/pkg/api/data_access/notifications.go
remoterami: You're right, it's more of a debug/testing thing - some internal dashboards don't have a name so I thought it wouldn't hurt

PR #966, opened 2024-10-16
feat(BEDS-536): add cache to raw store

Tangui-Bitfly:


backend/pkg/commons/db2/cache.go
```diff
@@ -0,0 +1,65 @@
...
+		return v.(*FullBlockRawData), nil
+	}
+
+	block, err := c.db.ReadBlockByNumber(chainID, number)
+	if block != nil {
+		c.cache.Store(key, block)
+
+		// retrieve the block hash for caching purpose
+		var mini MinimalBlock
+		_ = json.Unmarshal(block.Block, &mini)
```
Monika-Bitfly, line 42: we should probably check for any errors here

Monika-Bitfly (approved):

PR #976, opened 2024-10-17
chore(dashboard): change percentage threshold

peterbitfly:


frontend/components/bc/format/FormatPercent.vue
```diff
@@ -52,7 +52,7 @@ const data = computed(() => {
   }
   label = formatPercent(percent, config)
   if (props.comparePercent !== undefined) {
-    if (Math.abs(props.comparePercent - percent) <= 0.5) {
+    if (Math.abs(props.comparePercent - percent) <= 0.25) {
       className = 'text-equal'
       leadingIcon = faArrowsLeftRight
       compareResult = 'equal'
```
marcel-bitfly, line 58: this would be easier to understand, but I was a bit too chatty here I guess
```suggestion
    const thresholdToDifferenciateUnderperformerAndOverperformer = 0.25
    if (Math.abs(props.comparePercent - percent) <= thresholdToDifferenciateUnderperformerAndOverperformer) {
      className = 'text-equal'
      leadingIcon = faArrowsLeftRight
      compareResult = 'equal'
```

marcel-bitfly:
np: commit message

feat(BcFormatPercent): change threshold for `efficiency`

See: BEDS-622

marcel-bitfly (approved):
thx looks great.

depending on your workflow, this change should only add `one commit` to the history.
(I would squash everything locally, and do a `rebase and merge`)

PR #986, opened 2024-10-18
test: check if backend-types have been converted to frontend in ci

guybrush:
BEDS-90


marcel-bitfly:
np: commit message

```
ci: check if types for `frontend` have been updated

When `backend` makes changes this script will make sure
that the affected `typescript types` have been updated for
the `frontend`.

See: BEDS-90
```

marcel-bitfly (approved):

.github/workflows/backend-converted-types-check.yml
```diff
@@ -0,0 +1,46 @@
+name: Backend-Converted-Types-Check
+on:
+  push:
+    paths:
+      - 'backend/**'
+      - 'frontend/**'
```
marcel-bitfly, line 6:
```suggestion
```
this might save `compute time` and could be unnecessary if I understood the `usecase` correctly

guybrush: someone could change the types in frontend-dir, but you are right its better to use `frontend/types/api/**`

guybrush: updated

marcel-bitfly: I guess you still missed this line and only updated `pull_requests` 🤔

guybrush: yep, fixed

PR #998, opened 2024-10-21
(BEDS 628) refactor vdb blocks query

remoterami:
Applies goqu to the VDB blocks endpoint to make it more easily understandable. Should hopefully not decrease query performance.

Refactors the sorting/paging helper a bit in order to allow more complex expressions (like table names to avoid conflicts or sum of 2 columns as in this PR instead of column names only). Also adds handling `null` values in sortable columns.


backend/pkg/api/data_access/vdb_blocks.go
```diff
@@ -32,319 +41,272 @@ func (d *DataAccessService) GetValidatorDashboardBlocks(ctx context.Context, das
...
+		)
+	if dashboardId.Validators == nil {
+		filteredValidatorsDs = filteredValidatorsDs.
+			From(validators).
+			Where(validators.Col("dashboard_id").Eq(dashboardId.Id))
+		// apply search filters
 		if searchIndex {
-			params = append(params, search)
-			extraConds = append(extraConds, fmt.Sprintf(`validator_index = $%d`, len(params)))
+			filteredValidatorsDs = filteredValidatorsDs.Where(validators.Col("validator_index").Eq(search))
```
Eisei24, line 72: **issue:** Constructing the search like this doesn't work because the WHERE all get joined by an AND.
So if you search for an index or pubkey the format also qualifies a group name.
Then you only find results where both the index/pubkey and the group name work for the filter.

backend/pkg/api/data_access/vdb_blocks.go
```diff
@@ -32,319 +41,272 @@ func (d *DataAccessService) GetValidatorDashboardBlocks(ctx context.Context, das
...
-		// make sure the distinct clause filters out the correct duplicated row (e.g. block=nil)
-		orderBy += `, exec_block_number NULLS LAST`
-		query = selectStr + from + where + orderBy + limitStr
+		// for cursor only
+		Reward decimal.Decimal
 	}
-
 	startTime := time.Now()
-	err = d.alloyReader.SelectContext(ctx, &proposals, cte+query, params...)
+	query, args, err := blocksDs.Prepared(true).ToSQL()
```
Eisei24, line 295: **issue:** Guest dashboards do not work, you get an error here
`"unsupported type dataaccess.validatorGroup, a struct"`

backend/pkg/api/data_access/vdb_blocks.go
```diff
@@ -32,319 +41,272 @@ func (d *DataAccessService) GetValidatorDashboardBlocks(ctx context.Context, das
...
-	// Will just pass scheduled proposals to query and let db do the sorting etc
-	var scheduledProposers []t.VDBValidator
-	var scheduledEpochs []uint64
-	var scheduledSlots []uint64
-	// don't need to query if requested slots are in the past
+	// 5. Gather and supply scheduled blocks to let db do the sorting etc
 	latestSlot := cache.LatestSlot.Get()
-	if !onlyPrimarySort || !currentCursor.IsValid() ||
+	onlyPrimarySort := colSort.Column == enums.VDBBlockSlot
+	if !(onlyPrimarySort || colSort.Column == enums.VDBBlockBlock) || !currentCursor.IsValid() ||
```
Eisei24, line 197: **question:** As discussed I don't understand how this if works.

Eisei24 (changes requested):
Aside from the comments I had the issues:

* I do not get the correct slots in descending order, I only used `sort=slot:desc` and nothing else.
* I can't reach the second page without an error, I guess because of `WHERE (slot < ?)` so a missing parameter formatting.

remoterami:
Thanks for your feedback, should hopefully all be applied. I also fixed an issue with groups (scheduled blocks were always shown as default group I think) and inverted the default `nulls` sort behaviour (because sorting by Block Rewards DESC feels weird on large dashboards where you have to click through hundreds of missed/scheduled/orphaned blocks until you find your max)

backend/pkg/api/data_access/vdb_blocks.go
```diff
@@ -65,33 +66,38 @@ func (d *DataAccessService) GetValidatorDashboardBlocks(ctx context.Context, das
 		)
 	if dashboardId.Validators == nil {
 		filteredValidatorsDs = filteredValidatorsDs.
-			From(validators).
+			From(goqu.T("users_val_dashboards_validators").As(validators.GetTable())).
```
Eisei24, line 69: **thought:** It does not block anything now I just consider the logic here weird.
We defined `validators` above as a table even though no such table exists and then we say it is now just the alias of the actual table `users_val_dashboards_validators`.
I know that it is this way so that `validators` can be used elsewhere as a table but perhaps we can find a better solution in the future.

This is the same for groups but that one also doesn't have a `.GetTable()`.

backend/pkg/api/data_access/vdb_blocks.go
Eisei24: I am not quite sure whether it is because of my terrible db connection but we now get past this query creation line but then get a panic directly below it when executing the query.

Eisei24 (changes requested):
I still get a group id of 0 for all scheduled blocks.

backend/pkg/api/data_access/vdb_blocks.go
remoterami: Couldn't replicate

backend/pkg/api/data_access/vdb_blocks.go
remoterami: I agree the handling for the different dashboard id types makes it a bit messy and could be improved, but keeping the name as a single definition is still worth it. I've left a code comment :+1:

Eisei24 (approved):
lgtm

PR #1010, opened 2024-10-22
Beds 849/adjust discord webhooks

Eisei24:


backend/pkg/api/data_access/notifications.go
```diff
@@ -1942,10 +1940,13 @@ func (d *DataAccessService) GetNotificationSettingsDashboards(ctx context.Contex
 		resultMap[key].ChainIds = []uint64{valDashboard.Network}
 
 		// Set the settings
-		if valSettings, ok := resultMap[key].Settings.(*t.NotificationSettingsValidatorDashboard); ok {
```
Eisei24, line 1945: I noticed that we had a bug here (and the acc dashboard equivalent below).
The `Settings` are not a pointer so the type check never worked and we never got the correct webhook (& realtime mode) info, just the default `WebhookUrl = ""`, `IsWebhookDiscordEnabled = false`.

peterbitfly (approved):

PR #1014, opened 2024-10-22
(BEDS-848) Move search and cursor parameter validation to API

remoterami:


backend/pkg/api/enums/validator_dashboard_enums.go
```diff
@@ -170,6 +170,35 @@ var VDBBlocksColumns = struct {
...
+	Group     bool
+}
+
+func (s VDBBlocksSearches) GetSearches() []SearchType {
+	return []SearchType{
+		Integer,
+		ValidatorPublicKeyWithPrefix,
+		Name,
+	}
+}
```
LuccaBitfly, line 186: As discussed, we don't really need to provide a list of search types if we always check against every search type and just ignore them in `SetSearchType`

backend/pkg/api/enums/validator_dashboard_enums.go
```diff
@@ -170,6 +170,35 @@ var VDBBlocksColumns = struct {
 	VDBBlockProposerReward,
 }
 
+type VDBBlocksSearches struct {
```
LuccaBitfly, line 173: Don't know if these structs fit in the enum package. But also don't know where else they'd fit.

backend/pkg/api/handlers/input_validation.go
```diff
@@ -416,6 +435,17 @@ func checkSort[T enums.EnumFactory[T]](v *validationError, sortString string) *t
 	return &types.Sort[T]{Column: sortCol, Desc: desc}
 }
 
+func checkSearch[T enums.Searchable](searchString string) (T, error) {
```
LuccaBitfly, line 438: This func should prob never return an error, in worst case the returning searchable should be empty

LuccaBitfly (changes requested):

backend/pkg/api/enums/enums.go
```diff
@@ -16,6 +16,54 @@ func IsInvalidEnum(e Enum) bool {
...
+	Graffiti
+	Cursor
+	Email
+	Password
+	EmailUserToken
+	JsonContentType
+	// Validator Dashboard
+	ValidatorDashboardPublicId
+	ValidatorPublicKeyWithPrefix
+	ValidatorPublicKey
```
LuccaBitfly, line 38: Since const namespace is package-scoped, pls prefix these enums with `SearchType...`, like with the other enums.

LuccaBitfly (changes requested):

backend/pkg/api/enums/validator_dashboard_enums.go
remoterami: It's optional now

backend/pkg/api/enums/validator_dashboard_enums.go
remoterami: I moved it to `types`, I think that's slightly better, ptal

backend/pkg/api/handlers/input_validation.go
remoterami: It's returning a bool now to indicate an `invalid search` (if `false`), aka the search query didn't match any accepted patterns for this search. This may also check extra conditions, e.g. if the `aggregateGroups` flag is set, the group search is not applied even if the regex would otherwise match. See the three current examples, e.g. https://github.com/gobitfly/beaconchain/pull/1014/files#diff-eeb9062bd8258c1d32e584699f9dc786ad81df60981ec50fbc17b3c8962d7581R666-R669

Motivation for this approach was to allow early-outing with empty responses on the api layer already if possible, before even invoking data access at all.

backend/pkg/api/handlers/handler_service.go
```diff
@@ -267,6 +267,10 @@ func writeResponse(w http.ResponseWriter, r *http.Request, statusCode int, respo
 	}
 }
 
+func emptyPagingResponse() interface{} {
```
LuccaBitfly, line 272:
```suggestion
func emptyPagingResponse() types.ApiPagingResponse[any] {
```

backend/pkg/api/types/data_access.go
LuccaBitfly, line 1: I think it would be better if we seperate the types here. Cursors should probably go in their own `cursor.go`, and searches should also probably get their own. We prob have to find some good naming to differentiate between the search endpoint and the search filters on tables.

backend/pkg/api/handlers/public.go
```diff
@@ -657,13 +657,18 @@ func (h *HandlerService) PublicGetValidatorDashboardValidators(w http.ResponseWr
...
+	pagingParams := checkPagingParams[types.ValidatorsCursor](&v, q)
 	sort := checkSort[enums.VDBManageValidatorsColumn](&v, q.Get("sort"))
 	if v.hasErrors() {
 		handleErr(w, r, v)
 		return
 	}
-	data, paging, err := h.getDataAccessor(r).GetValidatorDashboardValidators(r.Context(), *dashboardId, groupId, pagingParams.cursor, *sort, pagingParams.search, pagingParams.limit)
+	search := &types.VDBManageValidatorsSearch{DashboardId: *dashboardId}
+	if !checkSearch(search, pagingParams.search) {
+		returnOk(w, r, emptyPagingResponse())
```
LuccaBitfly, line 708: needs a return statement after this (also for all other `checkSearch` calls)

LuccaBitfly: should maybe also discuss with team whether we want to early return, ignore, or return error (if you did and this was the conclusion please lmk)

backend/pkg/api/types/data_access.go
```diff
@@ -331,3 +352,169 @@ const CtxUserIdKey CtxKey = "user_id"
...
+
+type baseSearchResult[T any] struct {
+	Enabled bool
+	Value   T
+}
+
+type SearchNumber baseSearchResult[uint64]
+type SearchString baseSearchResult[string]
+
+func (bs *basicSearch) AsNumber(st SearchType) SearchNumber {
```
LuccaBitfly: Don't get why this needs to have a `SearchType` argument, feels like it unnecessarily complicates the API of using searchtypes.

backend/pkg/api/types/data_access.go
```diff
@@ -331,3 +352,169 @@ const CtxUserIdKey CtxKey = "user_id"
...
+	if !bs.types[st] {
+		log.Warn("tried accessing invalid search: ", st)
+		return SearchNumber{}
+	}
+
+	switch st {
+	case SearchTypeInteger:
+		number, err := strconv.ParseUint(bs.value, 10, 64)
+		if err != nil {
+			log.Error(err, "error converting search value, check regex parsing", 0)
```
LuccaBitfly: this should definitely return the error and not just log it

backend/pkg/api/types/data_access.go
```diff
@@ -331,3 +352,169 @@ const CtxUserIdKey CtxKey = "user_id"
...
+			return SearchNumber{}
+		}
+		return SearchNumber{true, number}
+	}
+	return SearchNumber{}
+}
+
+func (bs *basicSearch) AsString(st SearchType) SearchString {
```
LuccaBitfly: Again, I think the `SearchTypes` arg makes this more complicated than necessary. I think in practice the data access func would rather just want to have a `.GetValue()` and format the value itself rather than have something funky happening hidden behind a `AsString()`. If it turns out we want to reuse formatting logic, it should happen in another func IMO.

backend/pkg/api/handlers/public.go
```diff
@@ -1029,8 +1035,12 @@ func (h *HandlerService) PublicGetValidatorDashboardSummary(w http.ResponseWrite
 		handleErr(w, r, v)
 		return
 	}
+	search := &types.VDBSummarySearch{DashboardId: *dashboardId}
+	if !checkSearch(search, pagingParams.search) {
+		returnOk(w, r, emptyPagingResponse())
+	}
 
-	data, paging, err := h.getDataAccessor(r).GetValidatorDashboardSummary(r.Context(), *dashboardId, period, pagingParams.cursor, *sort, pagingParams.search, pagingParams.limit, protocolModes)
+	data, paging, err := h.getDataAccessor(r).GetValidatorDashboardSummary(r.Context(), *dashboardId, period, "", *sort, *search, pagingParams.limit, protocolModes)
```
LuccaBitfly, line 1088: Is passing the `""` intentional?

backend/pkg/api/enums/enums.go
```diff
@@ -1,6 +1,8 @@
 package enums
 
-import "time"
+import (
+	"time"
+)
```
LuccaBitfly, line 6: minor: could remove this change

backend/pkg/api/types/data_access.go
```diff
@@ -331,3 +352,169 @@ const CtxUserIdKey CtxKey = "user_id"
...
+	SearchTypeInteger SearchType = iota
+	SearchTypeName
+	SearchTypeEthereumAddress
+	SearchTypeWithdrawalCredential
+	SearchTypeEnsName
+	SearchTypeGraffiti
+	SearchTypeEmail
+	SearchTypePassword
+	SearchTypeEmailUserToken
+	SearchTypeJsonContentType
```
LuccaBitfly, line 364: Why are these valid search types?

backend/pkg/api/data_access/vdb.go
LuccaBitfly, line 1: For all changes within data access you'd probably want to get the direct opinion of data access team members. For now in total this PR adds more lines than it removes, so currently it's achieving the same thing with more lines. BUT it's good and necessary, that the cursor validation happens in handlers, not in data access. I don't know whether you've applied these search type changes everywhere or if it will scale better in future (e.g. with account dashboard). So I can't form a full opinion here.

LuccaBitfly (changes requested):

backend/pkg/api/types/data_access.go
remoterami: It was meant to enable increased but unified flexibility in case that different types should be returned in different formats, e.g. to scale ETH -> wei values or something. Atm I also can't think of many examples (removing/adding an optional `0x` prefix could be one though), so it's true that this looks like unneeded overhead, didn't think it's a big issue as this is encapsulated away from every-day data access workflow. But I'll disable it for now until a real usecase comes up

backend/pkg/api/types/data_access.go
remoterami: I tried to explicitly design the entire validation process in a way that api handles errors early so that data access doesn't have to deal with catching anymore at the point of using the result, as this is kind of the point of everything and greatly improves readability. Converting the raw string result to the expected format should happen earlier though, I'll try to revert back to using struct fields instead of methods

backend/pkg/api/handlers/public.go
remoterami: Yes basically, atleast until we decided what to do here: https://github.com/gobitfly/beaconchain/pull/1014/files#diff-bc69fc9b7d7d8ba0514ef4cf9cb7f09971cb7860fd7460d0dc6e3ebeb76e7b1bR31

backend/pkg/api/types/data_access.go
remoterami: They are not, just copied all regexes for now

backend/pkg/api/types/data_access.go
remoterami: Okay I've removed these methods and gone a step in that direction, ptal. Wouldn't remove the conversion entirely as doing it separately requires error handling in data access which would be a form of input validation.

backend/pkg/api/data_access/vdb.go
remoterami: Yeah you're right, I'll also ask other members. Types haven't been applied everywhere yet, will do so next; imo it simplifies things but the changes also need thorough testing

PR #1022, opened 2024-10-23
fix: implement proper auth check for editing paired devices

LuccaBitfly:
See: BEDS-863

### DIFF
--- a/backend/pkg/api/data_access/notifications.go
+++ b/backend/pkg/api/data_access/notifications.go
@@ -51,8 +51,9 @@ type NotificationsRepository interface {
 	GetNotificationSettingsDefaultValues(ctx context.Context) (*t.NotificationSettingsDefaultValues, error)
 	UpdateNotificationSettingsGeneral(ctx context.Context, userId uint64, settings t.NotificationSettingsGeneral) error
 	UpdateNotificationSettingsNetworks(ctx context.Context, userId uint64, chainId uint64, settings t.NotificationSettingsNetwork) error
-	UpdateNotificationSettingsPairedDevice(ctx context.Context, userId uint64, pairedDeviceId uint64, name string, IsNotificationsEnabled bool) error
-	DeleteNotificationSettingsPairedDevice(ctx context.Context, userId uint64, pairedDeviceId uint64) error
+	GetPairedDeviceUserId(ctx context.Context, pairedDeviceId uint64) (uint64, error)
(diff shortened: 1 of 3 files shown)


backend/pkg/api/data_access/notifications.go
```diff
@@ -1639,47 +1640,47 @@ func (d *DataAccessService) UpdateNotificationSettingsNetworks(ctx context.Conte
...
-	// TODO: This can be deleted when the API layer has an improved check for the device id
-	rowsAffected, err := result.RowsAffected()
-	if err != nil {
-		return err
-	}
-	if rowsAffected == 0 {
-		return fmt.Errorf("device with id %v to update notification settings not found", pairedDeviceId)
+		WHERE id = $3`,
+		name, IsNotificationsEnabled, pairedDeviceId)
+	if err == sql.ErrNoRows {
```
Eisei24, line 1667: **issue:** `ExecContext` does not return a `sql.ErrNoRows` if no rows were affected by either UPDATE or DELETE.

backend/pkg/api/data_access/notifications.go
```diff
@@ -1639,47 +1640,47 @@ func (d *DataAccessService) UpdateNotificationSettingsNetworks(ctx context.Conte
...
-	// TODO: This can be deleted when the API layer has an improved check for the device id
-	rowsAffected, err := result.RowsAffected()
-	if err != nil {
-		return err
+func (d *DataAccessService) DeleteNotificationSettingsPairedDevice(ctx context.Context, pairedDeviceId uint64) error {
+	_, err := d.userWriter.ExecContext(ctx, `
+		DELETE FROM users_devices
+		WHERE id = $1`,
+		pairedDeviceId)
+	if err == sql.ErrNoRows {
```
Eisei24, line 1678: **issue:** `ExecContext` does not return a `sql.ErrNoRows` if no rows were affected by either UPDATE or DELETE.

Eisei24 (changes requested):
This does fix the issue of the BEDS by not even entering `UpdateNotificationSettingsPairedDevice` or `DeleteNotificationSettingsPairedDevice` if the id does not exist but the methods themselves don't work like that.

backend/pkg/api/data_access/notifications.go
LuccaBitfly: fixed, ptal

backend/pkg/api/data_access/notifications.go
LuccaBitfly: fixed, ptal

Eisei24 (approved):
lgtm

PR #1053, opened 2024-10-31
feat: add validator count field to group summary

LuccaBitfly:
See: BEDS-626

### DIFF
--- a/frontend/types/api/validator_dashboard.ts
+++ b/frontend/types/api/validator_dashboard.ts
@@ -54,7 +54,8 @@ export interface VDBSummaryTableRow {
 export type GetValidatorDashboardSummaryResponse = ApiPagingResponse<VDBSummaryTableRow>;
 export interface VDBGroupSummaryColumnItem {
   status_count: StatusCount;
-  validators?: number /* uint64 */[];
+  validators?: number /* uint64 */[]; // fill with up to 3 validator indexes
+  validator_count: number /* uint64 */; // number of distinct validators
 }
 export interface VDBGroupSummarySyncCount {
   current_validators: number /* uint64 */;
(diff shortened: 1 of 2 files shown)


frontend/types/api/validator_dashboard.ts
```diff
@@ -54,7 +54,8 @@ export interface VDBSummaryTableRow {
 export type GetValidatorDashboardSummaryResponse = ApiPagingResponse<VDBSummaryTableRow>;
 export interface VDBGroupSummaryColumnItem {
   status_count: StatusCount;
-  validators?: number /* uint64 */[];
+  validators?: number /* uint64 */[]; // fill with up to 3 validator indexes
+  validator_count: number /* uint64 */; // number of distinct validators
```
LuccaBitfly, line 58: This will contain the count for sync and slash validators, within the `sync` and `slashings` field of a `VDBGroupSummarySyncCount`.

frontend/types/api/validator_dashboard.ts
```diff
@@ -75,7 +76,8 @@ export interface VDBGroupSummaryData {
   sync: VDBGroupSummaryColumnItem;
   sync_count: VDBGroupSummarySyncCount;
   slashings: VDBGroupSummaryColumnItem; // Failed slashings are count of validators in the group that were slashed
-  proposal_validators: number /* uint64 */[];
+  proposal_validators: number /* uint64 */[]; // fill with up to 3 validator indexes
+  proposal_validator_count: number /* uint64 */; // number of distinct validators
```
LuccaBitfly, line 80: This will contain the proposal validators count. Keep in mind that, unlike sync and slash, this does not get a `VDBGroupSummaryColumnItem`, since proposal data is already available earlier in the table call, as to avoid redundant data.

marcel-bitfly (approved):

PR #1054, opened 2024-10-31
fix(NotificationsManagementNetwork): `currentNetworkId` by using `environment variable`

marcel-bitfly:
As the `multichain feature` is not implemented yet, we are using `environment variables` to identify the current network

See: BEDS-898

### DIFF
--- a/frontend/components/notifications/management/NotificationsManagementNetwork.vue
+++ b/frontend/components/notifications/management/NotificationsManagementNetwork.vue
@@ -3,8 +3,13 @@ const { t: $t } = useTranslation()
 
 const notificationsManagementStore = useNotificationsManagementStore()
 
+const { chainIdByDefault } = useRuntimeConfig().public
+
 const networks = computed(() => notificationsManagementStore.settings.networks ?? [])
-const currentNetworkId = computed(() => notificationsManagementStore.settings.networks[0]?.chain_id ?? 0)
+const currentNetworkId = computed(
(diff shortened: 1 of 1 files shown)


frontend/components/notifications/management/NotificationsManagementNetwork.vue
```diff
@@ -3,8 +3,13 @@ const { t: $t } = useTranslation()
 
 const notificationsManagementStore = useNotificationsManagementStore()
 
+const { chainIdByDefault } = useRuntimeConfig().public
+
 const networks = computed(() => notificationsManagementStore.settings.networks ?? [])
-const currentNetworkId = computed(() => notificationsManagementStore.settings.networks[0]?.chain_id ?? 0)
+const currentNetworkId = computed(
+  () => notificationsManagementStore.settings.networks
+    .find(network => network.chain_id === Number(chainIdByDefault)) ?? 0,
```
marcel-bitfly, line 11: if `NUXT_PUBLIC_CHAIN_ID_BY_DEFAULT` is not available, setting `chain id` to `0` is might not be a good idea....would `1` be a better choice?

peterbitfly: using chain id 0 will throw an error on the api layer, setting it to 1 is not a good idea, as it will ultimately show wrong data.

peterbitfly (approved):

PR #1091, opened 2024-11-05
refactor(DashboardTableSummaryDetails): remove `missed rewards`

benji-bitfly:
This is a `quick-fix`, temporarily remove `missed rewards` per request.

See: BEDS-923

### DIFF
--- a/frontend/.vscode/settings.json
+++ b/frontend/.vscode/settings.json
@@ -9,6 +9,7 @@
     "BcToggle",
     "DashboardChartSummaryChartFilter",
     "DashboardGroupManagementModal",
+    "DashboardTableSummaryDetails",
     "DashboardTableValidators",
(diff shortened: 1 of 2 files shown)


frontend/.vscode/settings.json
```diff
@@ -22,6 +22,7 @@
     "NotificationsNetworkTable",
     "NotificationsOverview",
     "NotificationsTableEmpty",
+    "SummaryDetails",
```
marcel-bitfly, line 25: we should never add wrong ScopeNames -> `DashboardSummaryDetails` 

would be nice to rename the component then

benji-bitfly: definitely, thanks for the catch :pray:

marcel-bitfly (approved):
🤷

PR #1098, opened 2024-11-06
fix(DashboardValidatorManagmentModal): check if user has `bulk adding`

benji-bitfly:
See: BEDS-885


marcel-bitfly (approved):
💪 nice thx for the fix

PR #1100, opened 2024-11-06
Fixed wrong type of EventType in the cursor

Eisei24:


backend/pkg/api/types/data_access.go
```diff
@@ -139,20 +140,12 @@ type NotificationClientsCursor struct {
 	Ts     time.Time
 }
 
-type NotificationRocketPoolsCursor struct {
```
Eisei24, line 142: An obsolete cursor.

peterbitfly (approved):

PR #1109, opened 2024-11-08
refactor(DashboardCreationController): update `placeholder`

benji-bitfly:
See: BEDS-570

### DIFF
--- a/frontend/components/dashboard/creation/DashboardCreationController.vue
+++ b/frontend/components/dashboard/creation/DashboardCreationController.vue
@@ -78,7 +78,7 @@ function show(
   }
   network.value = forcedNetwork || currentNetwork.value
   state.value = 'type'
-  name.value = isLoggedIn.value ? '' : 'cookie'
+  name.value = isLoggedIn.value ? '' : ''
 }
 
 defineExpose({ show })
--- a/frontend/locales/en.json
+++ b/frontend/locales/en.json
@@ -117,7 +117,7 @@
       "title": "Add a new dashboard",
       "type": {
         "accounts": "Accounts",
-        "placeholder": "Name",
+        "placeholder": "Enter dashboard name",
         "subtitle": "What would you like to monitor?",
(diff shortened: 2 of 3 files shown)


marcel-bitfly:
np: `commit msg` is kinda misleading atm

frontend/components/dashboard/creation/DashboardCreationController.vue
```diff
@@ -78,7 +78,7 @@ function show(
   }
   network.value = forcedNetwork || currentNetwork.value
   state.value = 'type'
-  name.value = isLoggedIn.value ? '' : 'cookie'
+  name.value = isLoggedIn.value ? '' : ''
```
marcel-bitfly, line 81: I am not entirely sure on my own what this `line` does...which means that the whole `feature` needs refactoring...but plz never do things like this, because this line no is `max-confusing`

```suggestion
```
(would exactly be the same)

frontend/locales/en.json
```diff
@@ -117,7 +117,7 @@
       "title": "Add a new dashboard",
       "type": {
         "accounts": "Accounts",
-        "placeholder": "Name",
+        "placeholder": "Enter dashboard name",
```
marcel-bitfly, line 120: nice that you improve things along the way 💪.

But we are using `placeholder` as `input labels`.

See: https://www.deque.com/blog/accessible-forms-the-problem-with-placeholders/

Placeholder value should be an `example` of the things `a user should input`.
Maybe you come up with a different idea.

marcel-bitfly:
np: commit `msg`s: plz try to describe your changes as simple as possible, also double check your grammar to prevent confusion

refactor(DashboardCreationController): remove `cookie` for `dashboard name`

In the `dashboard creation process` the `input field` field for the `dashboard name`
was prefilled with `cookie`.

See: BEDS-570

---

feat(DashboardCreationController): update placeholder for `dashboard name`

PR #1116, opened 2024-11-11
Added conditions to the subqueries that use union

Eisei24:
The `blocksDs` and `scheduledDs` subqueries now also directly apply the limitations from pagination and limit with ordering.

### DIFF
--- a/backend/pkg/api/data_access/vdb_blocks.go
+++ b/backend/pkg/api/data_access/vdb_blocks.go
@@ -198,8 +198,7 @@ func (d *DataAccessService) GetValidatorDashboardBlocks(ctx context.Context, das
 	if err != nil {
 		return nil, nil, err
 	}
-	blocksDs = goqu.Dialect("postgres").From(goqu.T("past_blocks_cte")).
-		With("past_blocks_cte", blocksDs). // encapsulate so we can use selected fields
+	blocksDs = blocksDs.
 		Order(order...)
 	if directions != nil {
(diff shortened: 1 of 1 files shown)


backend/pkg/api/data_access/vdb_blocks.go
```diff
@@ -267,12 +266,17 @@ func (d *DataAccessService) GetValidatorDashboardBlocks(ctx context.Context, das
...
 					groupId,
 					goqu.V("0").As("status"),
-					goqu.V(nil).As("exec_block_number"),
-					goqu.V(nil).As("graffiti_text"),
-					goqu.V(nil).As("fee_recipient"),
-					goqu.V(nil).As("el_reward"),
+					goqu.L("NULL::INTEGER").As("exec_block_number"),
+					goqu.L("NULL::TEXT").As("graffiti_text"),
+					goqu.L("NULL::BYTEA").As("fee_recipient"),
+					goqu.L("NULL::NUMERIC").As("el_reward"),
```
Eisei24, line 272: This got quite complicated here.
As soon as `scheduledDs` gets a `LIMIT` added to it the `Union` goqu method determines that it now needs to wrap itself, meaning `SELECT ...` => `SELECT * FROM ( SELECT ... LIMIT ...) AS t1`
As a result of that it could no longer correctly identify the types of our NULL values.
The only solution I found is adding such casts.

peterbitfly (approved):

PR #1117, opened 2024-11-12
docs(decision):add conventional commits

marcel-bitfly:
Using https://www.ietf.org/rfc/rfc2119.txt


frontend/decision-record.md
marcel-bitfly, line 1: This file could also be removed.

peterbitfly (approved):

PR #1123, opened 2024-11-12
Beds 880/adding many validators to dashboard freezes

Eisei24:


LuccaBitfly (approved):
type changes lgtm, haven't looked at data access though

backend/pkg/api/data_access/vdb_management.go
```diff
@@ -836,83 +835,68 @@ func (d *DataAccessService) GetValidatorDashboardGroupExists(ctx context.Context
...
-		ON CONFLICT (dashboard_id, validator_index) DO UPDATE SET
-			dashboard_id = EXCLUDED.dashboard_id,
-			group_id = EXCLUDED.group_id,
-			validator_index = EXCLUDED.validator_index
-		RETURNING validator_index, group_id
-	`
-
-	// Find all the pubkeys
-	err := d.alloyReader.SelectContext(ctx, &pubkeys, pubkeysQuery, pq.Array(validators))
+	tx, err := d.userWriter.BeginTxx(ctx, nil)
```
peterbitfly, line 845: this should be `writerDb` not `userWriter`.

backend/pkg/api/data_access/vdb_management.go
```diff
@@ -836,83 +835,68 @@ func (d *DataAccessService) GetValidatorDashboardGroupExists(ctx context.Context
...
-	for _, validatorIndex := range validators {
-		addValidatorsArgsIntf = append(addValidatorsArgsIntf, validatorIndex)
-	}
-	err = d.alloyWriter.SelectContext(ctx, &addedValidators, addValidatorsQuery, addValidatorsArgsIntf...)
-	if err != nil {
-		return nil, err
-	}
+	numArgs := 3
+	batchSize := 65535 / numArgs // max 65535 params per batch, since postgres uses int16 for binding input params
+	batchIdx, allIdx := 0, 0
```
peterbitfly, line 853: goqu has a built in batch insert feature, see https://github.com/gobitfly/beaconchain/blob/staging/backend/pkg/notification/queuing.go#L448 for its usage

peterbitfly (changes requested):

backend/pkg/api/data_access/vdb_management.go
Eisei24: Right, copy&paste error. Good find.

backend/pkg/api/data_access/vdb_management.go
Eisei24: As discussed the linked code part does not avoid the parameter limit.

The only way I have found is e.g. [here](https://github.com/gobitfly/beaconchain/blob/743a86528b0930da74e63d6a5420dfe5a86bd368/backend/pkg/exporter/modules/ethstore.go#L190) with the use of `pgx` and `CopyFrom`/`CopyFromSlice`. If a high email notification count has to work for the linked part you could use that. I just can't use it here because it does not support `ON CONFLICT` handling.

peterbitfly (approved):

PR #1124, opened 2024-11-13
refactor: remove `network average`

benji-bitfly:
See: BEDS-641


marcel-bitfly (approved):
thx for aligning our `component names` 👍

(but would have been 💯 to have a separat commit for this)

PR #1147, opened 2024-11-21
Rocket Pool Smoothing Pool Export & Cheaper Minipool Data Access

manuelsc:


backend/pkg/exporter/modules/rocketpool.go
```diff
@@ -731,6 +794,21 @@ func (rp *RocketpoolExporter) SaveMinipools() error {
 		}
 	}
 
+	// updating index column after writing minipools for cheaper access later
+	_, err = tx.Exec(`
+		WITH mapping AS (
```
peterbitfly: I would prefer if this could use a direct join on the validators table instead of loading the whole validators table?
```
UPDATE A
SET foo = B.bar
FROM TableA A
JOIN TableB B
    ON A.col1 = B.colx
WHERE ...
```

peterbitfly (changes requested):

backend/pkg/exporter/modules/rocketpool.go
manuelsc: Fair enough, I use the validators table directly. Afaik postgres will treat the FROM WHERE as an implicit join anyway

peterbitfly (approved):

PR #1149, opened 2024-11-21
feat(DashboardTableEfficiency): update `UI` for relative values in th…

benji-bitfly:
…e `validator summary`

Display `-` when relative values are equal to `0`.

See: BEDS-844

### DIFF
--- a/frontend/components/dashboard/table/DashboardTableEfficiency.vue
+++ b/frontend/components/dashboard/table/DashboardTableEfficiency.vue
@@ -49,13 +49,20 @@ const data = computed(() => {
       />
     </span>
 
-    <BcFormatPercent
-      v-else
-      class="percent"
+    <BaseFormatPercent
+      v-else-if="props.success"
+      v-slot="{ result, ratio }"
       :base="data.sum"
       :value="props.success"
-      :color-break-point="80"
-      :full-on-empty-base="true"
-    />
+    >
+      <BcColor
+        :color="ratio >= 0.8 ? 'green' : 'red'"
+      >
+        {{ result }}
+      </BcColor>
+    </BaseFormatPercent>
+    <span v-else>
+      -
+    </span>
   </BcTooltip>
 </template>
--- a/frontend/components/base/BaseFormatPercent.vue
+++ b/frontend/components/base/BaseFormatPercent.vue
@@ -19,18 +19,24 @@ const ratio = computed(() => {
   assertIsNotZero(base)
   return (Number(value) / Number(base))
 })
+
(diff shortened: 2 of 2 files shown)


marcel-bitfly:
np: `commit-msgs`

```
feat(DashboardTableEfficiency): display `-` when relative values are equal to `0`

See: BEDS-844
```

frontend/components/dashboard/table/DashboardTableEfficiency.vue
```diff
@@ -32,30 +32,36 @@ const data = computed(() => {
...
     <span v-if="absolute">
-      <BcFormatNumber
-        class="positive"
-        :value="props.success"
-      />
-      <span class="slash"> / </span>
-      <BcFormatNumber
-        :class="data.failedClass"
-        :value="props.failed"
+      <span v-if="props.success === 0 && props.failed === 0">0 / 0</span>
```
marcel-bitfly: how would `50/70` be displayed? 
1. `50/50` or
2. `50 / 50`

frontend/components/dashboard/table/DashboardTableEfficiency.vue
```diff
@@ -49,13 +49,20 @@ const data = computed(() => {
...
       :value="props.success"
-      :color-break-point="80"
-      :full-on-empty-base="true"
-    />
+    >
+      <BcColor
+        :color="ratio >= 0.8 ? 'green' : 'red'"
+      >
+        {{ result }}
+      </BcColor>
```
enzo-bitfly, line 62: question: why do we need `BcColor` in the slot instead of passing the `color` prop to `BaseFormatPercent`?

marcel-bitfly: I need `ratio` an information which comes from the `child component` (v-slot -> scoped slot prop).
(as I didn't find a prop name which expresses that the color will change from `red` to `green` at a certain point of `threshold` [color-break-point before])

enzo-bitfly: Maybe `statusColorThreshold` 🤔 your decision to keep it like this or not.

enzo-bitfly (approved):

frontend/components/dashboard/table/DashboardTableEfficiency.vue
marcel-bitfly: We have to revisit this for sure when taking a look at the other instances where percentages or numbers (with or without icons) change their colours

PR #1161, opened 2024-11-25
Rocket Pool Dashboard List

manuelsc:


backend/pkg/api/data_access/mobile.go
```diff
@@ -408,20 +398,20 @@ func (d *DataAccessService) GetValidatorDashboardMobileValidators(ctx context.Co
...
-			rn.smoothing_pool_opted_in
+			rn.smoothing_pool_opted_in,
+			validator_index
 		FROM rocketpool_minipools
 		LEFT JOIN rocketpool_nodes rn ON rocketpool_minipools.node_address = rn.address
-		WHERE pubkey = ANY($1)
+		WHERE validator_index = ANY($1)
 		`
-		err := d.alloyReader.SelectContext(ctx, &rocketPoolResults, validatorsQuery, pq.ByteaArray(pubKeys))
+		err := d.alloyReader.SelectContext(ctx, &rocketPoolResults, validatorsQuery, indices)
```
remoterami, line 407: This query could probably be merged with the call to `GetValidatorDashboardValidators`, I'm not the biggest fan of passing (potentially large) validator arrays around via `ANY()`.
But I can see how that's a lot of effort, so prob only worth doing once we hit perf issues

backend/pkg/api/data_access/vdb_rocket_pool.go
```diff
@@ -2,14 +2,251 @@ package dataaccess
...
-	return d.dummy.GetValidatorDashboardRocketPool(ctx, dashboardId, cursor, colSort, search, limit)
+	// Initialize the cursor
+	var currentCursor t.RocketPoolCursor
+	var err error
+	var paging t.Paging
+
+	if cursor != "" {
+		currentCursor, err = utils.StringToCursor[t.RocketPoolCursor](cursor)
+		if err != nil {
+			return nil, &paging, fmt.Errorf("failed to parse passed cursor as ValidatorsCursor: %w", err)
```
remoterami:
```suggestion
			return nil, &paging, fmt.Errorf("failed to parse passed cursor as RocketPoolCursor: %w", err)
```

backend/pkg/api/data_access/vdb_rocket_pool.go
```diff
@@ -2,14 +2,251 @@ package dataaccess
...
+	ds := goqu.Dialect("postgres").
+		From(goqu.T("rocketpool_minipools").As("mp")).
+		Select(
+			goqu.T("n").Col("address").As("naddress"),
+			goqu.L("SUM(mp.node_deposit_balance) AS staked_eth"),
+			goqu.T("n").Col("rpl_stake"),
+			goqu.L("COUNT(mp.address) as minipools_total"),
+			goqu.L("SUM(CASE WHEN mp.node_deposit_balance = 8e18 THEN 1 ELSE 0 END) AS minipools_leb8"),
+			goqu.L("SUM(CASE WHEN mp.node_deposit_balance = 16e18 THEN 1 ELSE 0 END) AS minipools_leb16"),
+			goqu.L("CASE WHEN COUNT(mp.address) > 0 THEN SUM(mp.node_fee) / COUNT(mp.address) ELSE 0 END AS avg_commission"),
```
remoterami: **check:** Should the average commission only factor in `Staking` minipools? Depending on how the data is used. Also this could be weighted by the amount of borrowed ETH per MP, as the user is prob interested on the avg additional rewards.

remoterami (commented):
Some minor comments/thoughts, ptal

backend/pkg/api/data_access/mobile.go
manuelsc: Good point, they way I handled it is that it will use the result table of GetValidatorDashboardValidators to get the indices, not the full list of all indices. Basically this is already a truncated version of the indices involved and should never be more than whatever max rows `GetValidatorDashboardValidators` returns

backend/pkg/api/data_access/vdb_rocket_pool.go
manuelsc: Great catch, only counting staking minipools now and weighted based on deposit size

backend/pkg/api/data_access/mobile.go
remoterami: Ah right it's limited because you're passing the cursor/limit along, should be minimal impact then :+1:

backend/pkg/api/data_access/vdb_rocket_pool.go
remoterami: > weighted based on deposit size

Not sure what it's doing now.. am I reading this right, basically it's `fee * node_deposit_balance / node_deposit_balance`? That'd not achieve a lot, but I could be missing something :smile:
I was thinking about `fee * user_deposit_balance / node_deposit_balance`. But I'm pretty sure there are other places where this has not been applied, should prob also check that and create tickets maybe.

remoterami (approved):
lgtm, just fix CI tasks please (FE type conversion)

PR #1163, opened 2024-11-26
Beds 982/login tests

tanya-bitfly:


frontend/tests/.prettierrc
marcel-bitfly, line 1: for now we are not using prettier, but [eslint stylistic](https://eslint.style/packages/default), if you want to add some rules you can add them to [eslint.config.mjs](https://github.com/gobitfly/beaconchain/blob/cdffa1ef8f5c23a19200c6ea1e2108494f1d2ff5/frontend/eslint.config.mjs#L0-L1)

If you need help to configure your `ide` feel free to reach out

frontend/package.json
```diff
@@ -56,6 +58,7 @@
     "lint:fix": "npm run lint -- --fix",
     "postinstall": "nuxt prepare",
     "preview": "nuxt preview",
+    "test": "npx playwright test --config=tests/playwright.config.ts",
```
marcel-bitfly: we will add more `tests` like `unit` and `componets` soon, so this might describe your `tests`
 better
```suggestion
    "test:e2e": "playwright test --config=tests/playwright.config.ts",
```

frontend/tests/page-object/base.page.ts
```diff
@@ -0,0 +1,8 @@
+import type { Page } from "@playwright/test";
+
+export const BasePage = {
+    goto: async (page: Page, endpoint: string) => {
+        await page.goto("https://v2-staging-mainnet.beaconcha.in" + endpoint);
+        await page.waitForTimeout(2000);
+    },
```
marcel-bitfly, line 7:
```suggestion
export const basePage = {
    goto: async (page: Page, path: string) => {
        await page.goto("https://v2-staging-mainnet.beaconcha.in" + path);
        await page.waitForTimeout(2000);
    },
```

🤔 don't know if we should `stick` to your `PascalcCase convention` for `classes` or go with the `default` in `js`

frontend/package.json
tanya-bitfly: good note, updated to "test:e2e": "playwright test --config=tests/playwright.config.ts",

frontend/tests/page-object/base.page.ts
tanya-bitfly: I would rather use PascalCase, just because in tests it will give more visability that this is for the specific page like await BasePage.goto(page)

PR #1167, opened 2024-11-27
Beds 151/add rp mode test

Eisei24:
When reviewing this the important step to keep in mind is that I had to adjust the queries to give me the results for each validator individually instead of aggregates because I then need to multiply the reward with the validator specific rocketpool factor.

EL rewards for minipools in the smoothing pool should be ignored in the query (except in the blocks view), instead the rewards should be taken from the "smoothing pool reward table" split up for each minipool.

### DIFF
--- a/backend/pkg/api/data_access/vdb_helpers.go
+++ b/backend/pkg/api/data_access/vdb_helpers.go
@@ -135,41 +135,38 @@ func (d *DataAccessService) getTimeToNextWithdrawal(distance uint64) time.Time {
 }
 
 func (d *DataAccessService) getRocketPoolInfos(ctx context.Context, dashboardId t.VDBId, groupId int64) (*t.RPInfo, error) {
+	var rpInfo t.RPInfo
 	wg := errgroup.Group{}
 
 	queryResult := []struct {
-		ValidatorIndex       uint64           `db:"validatorindex"`
-		NodeAddress          []byte           `db:"node_address"`
-		NodeFee              float64          `db:"node_fee"`
-		NodeDepositBalance   decimal.Decimal  `db:"node_deposit_balance"`
-		UserDepositBalance   decimal.Decimal  `db:"user_deposit_balance"`
-		EndTime              sql.NullTime     `db:"end_time"`
-		SmoothingPoolAddress []byte           `db:"smoothing_pool_address"`
-		SmoothingPoolEth     *decimal.Decimal `db:"smoothing_pool_eth"`
+		ValidatorIndex     uint64           `db:"validator_index"`
+		NodeAddress        []byte           `db:"node_address"`
+		NodeFee            float64          `db:"node_fee"`
+		NodeDepositBalance decimal.Decimal  `db:"node_deposit_balance"`
+		UserDepositBalance decimal.Decimal  `db:"user_deposit_balance"`
+		EndTime            sql.NullTime     `db:"end_time"`
+		SmoothingPoolEth   *decimal.Decimal `db:"smoothing_pool_eth"`
 	}{}
 
 	wg.Go(func() error {
 		ds := goqu.Dialect("postgres").
 			Select(
-				goqu.L("v.validatorindex"),
+				goqu.L("rplm.validator_index"),
 				goqu.L("rplm.node_address"),
 				goqu.L("rplm.node_fee"),
 				goqu.L("rplm.node_deposit_balance"),
 				goqu.L("rplm.user_deposit_balance"),
 				goqu.L("rplrs.end_time"),
-				goqu.L("rploc.smoothing_pool_address"),
 				goqu.L("rplrs.smoothing_pool_eth"),
 			).
 			From(goqu.L("rocketpool_minipools AS rplm")).
-			LeftJoin(goqu.L("validators AS v"), goqu.On(goqu.L("rplm.pubkey = v.pubkey"))).
 			LeftJoin(goqu.L("rocketpool_rewards_summary AS rplrs"), goqu.On(goqu.L("rplm.node_address = rplrs.node_address"))).
-			LeftJoin(goqu.L("rocketpool_onchain_configs AS rploc"), goqu.On(goqu.L("rplm.rocketpool_storage_address = rploc.rocketpool_storage_address"))).
 			Where(goqu.L("rplm.node_deposit_balance IS NOT NULL")).
 			Where(goqu.L("rplm.user_deposit_balance IS NOT NULL"))
 
 		if len(dashboardId.Validators) == 0 {
 			ds = ds.
-				LeftJoin(goqu.L("users_val_dashboards_validators uvdv"), goqu.On(goqu.L("uvdv.validator_index = v.validatorindex"))).
+				LeftJoin(goqu.L("users_val_dashboards_validators uvdv"), goqu.On(goqu.L("uvdv.validator_index = rplm.validator_index"))).
 				Where(goqu.L("uvdv.dashboard_id = ?", dashboardId.Id))
 
 			if groupId != t.AllGroups {
@@ -178,7 +175,7 @@ func (d *DataAccessService) getRocketPoolInfos(ctx context.Context, dashboardId 
 			}
 		} else {
 			ds = ds.
-				Where(goqu.L("v.validatorindex = ANY(?)", pq.Array(dashboardId.Validators)))
+				Where(goqu.L("rplm.validator_index = ANY(?)", pq.Array(dashboardId.Validators)))
 		}
 
 		query, args, err := ds.Prepared(true).ToSQL()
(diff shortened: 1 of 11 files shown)


backend/pkg/api/data_access/vdb_helpers.go
```diff
@@ -130,3 +133,136 @@ func (d *DataAccessService) getTimeToNextWithdrawal(distance uint64) time.Time {
...
+			}
+		}
+
+		node := hexutil.Encode(res.NodeAddress)
+		if res.EndTime.Valid && res.SmoothingPoolEth != nil {
+			epoch := uint64(utils.TimeToEpoch(res.EndTime.Time))
+			splitReward := res.SmoothingPoolEth.Div(decimal.NewFromUint64(nodeMinipoolCount[node]))
+			rpInfo.Minipool[res.ValidatorIndex].SmoothingPoolRewards[epoch] =
+				rpInfo.Minipool[res.ValidatorIndex].SmoothingPoolRewards[epoch].Add(splitReward)
+		}
```
remoterami, line 253: This looks quite complicated, I don't think we need to handle multiple result rows with the same validator index, do we? Should be unique (unless we're planning for reusable indices maybe, but then the math would be wrong).

But I could be missing something, so as it's also not wrong I'll keep it as is

backend/pkg/api/data_access/vdb_helpers.go
```diff
@@ -130,3 +133,136 @@ func (d *DataAccessService) getTimeToNextWithdrawal(distance uint64) time.Time {
...
+	}
+
+	if len(queryResult) == 0 {
+		return nil, nil
+	}
+
+	rpInfo := t.RPInfo{
+		Minipool: make(map[uint64]t.RPMinipoolInfo),
+		// Smoothing pool address is the same for all nodes on the network so take the first result
+		SmoothingPoolAddress: queryResult[0].SmoothingPoolAddress,
```
remoterami: Should use the other query for this information. As its result will have less rows, we'll also return less duplicates. Also makes the first query simpler which is the slower query here.

backend/pkg/api/data_access/vdb_helpers.go
```diff
@@ -130,3 +133,136 @@ func (d *DataAccessService) getTimeToNextWithdrawal(distance uint64) time.Time {
...
+				goqu.L("rplm.node_address"),
+				goqu.L("rplm.node_fee"),
+				goqu.L("rplm.node_deposit_balance"),
+				goqu.L("rplm.user_deposit_balance"),
+				goqu.L("rplrs.end_time"),
+				goqu.L("rploc.smoothing_pool_address"),
+				goqu.L("rplrs.smoothing_pool_eth"),
+			).
+			From(goqu.L("rocketpool_minipools AS rplm")).
+			LeftJoin(goqu.L("validators AS v"), goqu.On(goqu.L("rplm.pubkey = v.pubkey"))).
```
remoterami: This join is not necessary anymore, the minipools table got a `validator_index` column recently.

backend/pkg/api/data_access/vdb_management.go
```diff
@@ -362,61 +370,16 @@ func (d *DataAccessService) GetValidatorDashboardOverview(ctx context.Context, d
...
-			metadata := validatorMapping.ValidatorMetadata[validator]
+			// Balance
 			validatorBalance := utils.GWeiToWei(big.NewInt(int64(metadata.Balance)))
 			effectiveBalance := utils.GWeiToWei(big.NewInt(int64(metadata.EffectiveBalance)))
 
-			if rpValidator, ok := rpValidators[validator]; ok {
+			if rpInfos == nil {
+				data.Balances.Total = data.Balances.Total.Add(validatorBalance)
+
+				nonRpDashboardId.Validators = append(nonRpDashboardId.Validators, validator)
```
remoterami: This does the same as the `else` block, cleaned this up slightly

backend/pkg/api/data_access/vdb_rewards.go
```diff
@@ -661,57 +792,70 @@ func (d *DataAccessService) GetValidatorDashboardGroupRewards(ctx context.Contex
...
 			ret.Inactivity.StatusCount.Success++
 		}
 
-		ret.Proposal.Income = ret.Proposal.Income.Add(entry.BlocksClReward.Mul(gWei))
+		elReward := elRewards[entry.ValidatorIndex].Mul(rpFactor)
+		if rpInfos != nil && protocolModes.RocketPool {
+			if _, ok := rpInfos.Minipool[entry.ValidatorIndex]; ok {
+				if _, ok := rpInfos.Minipool[entry.ValidatorIndex].SmoothingPoolRewards[epoch]; ok {
+					elReward = elReward.Add(rpInfos.Minipool[entry.ValidatorIndex].SmoothingPoolRewards[epoch])
+				}
```
remoterami: Accessing an non-existing key doesn't crash in go, it just returns the empty object. This would make this nested if more readable imo

backend/pkg/api/data_access/vdb_rewards.go
```diff
@@ -759,54 +923,95 @@ func (d *DataAccessService) GetValidatorDashboardRewardsChart(ctx context.Contex
...
+
+			validatorGroupMap[entry.ValidatorIndex] = entry.GroupId
+
+			current := &queryResultSum[len(queryResultSum)-1]
+			reward := utils.GWeiToWei(big.NewInt(entry.ClRewards))
+			if rpInfos != nil && protocolModes.RocketPool {
+				if rpValidator, ok := rpInfos.Minipool[entry.ValidatorIndex]; ok {
+					reward = reward.Mul(d.getRocketPoolOperatorFactor(rpValidator))
+				}
+			}
```
remoterami, line 996: This code section repeats quite often, thinking about generalizing this - how about defining a method on the `RPInfo` struct?

backend/pkg/api/data_access/vdb_rewards.go
```diff
@@ -759,54 +923,95 @@ func (d *DataAccessService) GetValidatorDashboardRewardsChart(ctx context.Contex
...
+		// Has to be done here in the cl and not el part because here we have the list of all relevant validators
+		if rpInfos != nil && protocolModes.RocketPool {
+			for validatorIndex, groupId := range validatorGroupMap {
+				for epoch, reward := range rpInfos.Minipool[validatorIndex].SmoothingPoolRewards {
+					if _, ok := smoothingPoolRewards[epoch]; !ok {
+						smoothingPoolRewards[epoch] = make(map[uint64]decimal.Decimal)
+					}
+					smoothingPoolRewards[epoch][groupId] = smoothingPoolRewards[epoch][groupId].Add(reward)
+				}
+			}
```
remoterami, line 1010: this and the start of the loop are copy pasted, could try to unify

backend/pkg/api/data_access/vdb_rewards.go
```diff
@@ -1041,37 +1284,91 @@ func (d *DataAccessService) GetValidatorDashboardDuties(ctx context.Context, das
...
+	type QueryResultAdjusted struct {
+		QueryResultBase
+		AttestationsSourceReward    decimal.Decimal
+		AttestationsTargetReward    decimal.Decimal
+		AttestationsHeadReward      decimal.Decimal
+		SyncRewards                 decimal.Decimal
+		BlocksClSlasherReward       decimal.Decimal
+		BlocksClAttestationsReward  decimal.Decimal
+		BlocksClSyncAggregateReward decimal.Decimal
+	}
```
remoterami: CHECK: can these be combined? Is there a clickhouse driver which can fill decimals?

backend/pkg/api/data_access/vdb_rewards.go
```diff
@@ -1041,37 +1284,91 @@ func (d *DataAccessService) GetValidatorDashboardDuties(ctx context.Context, das
...
+					AttestationsSourceExecuted: entry.AttestationsSourceExecuted,
+					AttestationsTargetExecuted: entry.AttestationsTargetExecuted,
+					AttestationsHeadExecuted:   entry.AttestationsHeadExecuted,
+					SyncScheduled:              entry.SyncScheduled,
+					SyncExecuted:               entry.SyncExecuted,
+					Slashed:                    entry.Slashed,
+					BlocksSlashingCount:        entry.BlocksSlashingCount,
+					BlocksScheduled:            entry.BlocksScheduled,
+					BlocksProposed:             entry.BlocksProposed,
+				},
```
remoterami: Can be simplified

backend/pkg/api/data_access/vdb_summary.go
```diff
@@ -172,22 +170,80 @@ func (d *DataAccessService) GetValidatorDashboardSummary(ctx context.Context, da
...
+		groupSum.SyncScheduled += row.SyncScheduled
+
+		clRewardWei := utils.GWeiToWei(big.NewInt(row.ClRewards))
+		if rpInfos != nil && protocolModes.RocketPool {
+			if rpValidator, ok := rpInfos.Minipool[row.ValidatorIndex]; ok {
+				clRewardWei = clRewardWei.Mul(d.getRocketPoolOperatorFactor(rpValidator))
+			}
 		}
+		groupSum.ClRewards = groupSum.ClRewards.Add(clRewardWei)
+		queryResultSumMap[row.GroupId] = groupSum
```
remoterami: Not sure if splitting all those sum queries into individual rows in order to apply the correct RP scale factor and summing up in code is the best approach. Mainly because of efficiency reasons:
- It increases the total amount of data that needs to be transferred (e.g. from 11 rows to 2.75 million for the query in `GetValidatorDashboardRewards` for megatron)
- The DB is likely more efficient in batch processing like sums as its specialized for such tasks
- Readability/Maintainablity: All that manual summing is a lot more code to maintain
- All of the above are only done to handle RP Minipool rewards correctly which likely is just a minor subset of all validators, but the mechanism is applied to _all_ validators

If there's no better solution that has to be put up with of course. Also depends how far we want to push for optimization; But shouldn't it be possible to let the DB do it? Either directly (for EL/alloy) or by feeding the list of Minipool indices + NO shares if it's a cross-db query.

Just thoughts for now and maybe it can't be applied to all queries, but even some could help. Could also go for a middleground where this is only applied if the RP mode is set, but stick to the old db-summing otherwise (would help with efficiency but prob not readability)

remoterami (changes requested):
Left some thoughts; will implement pending changes

backend/pkg/api/data_access/vdb_rewards.go
remoterami: Answer: yes should be possible. Not aware of any precision loss or something so I merged these

PR #1207, opened 2024-12-12
Beds 1056/postman ci cd

sasha-bitfly:


sasha-bitfly:
We will use [api_test.go](https://github.com/gobitfly/beaconchain/blob/staging/backend/pkg/api/api_test.go) for running e2e tests

PR #1214, opened 2024-12-16
BEDS-1061 / rounding blocks cl rewards for mainnet

sasha-bitfly:


remoterami (approved):
right missed this, lgtm

PR #1220, opened 2024-12-18
(BEDS-875) Add rewards chart history based on premium perks

remoterami:


backend/pkg/api/types/user.go
```diff
@@ -125,6 +125,7 @@ type PremiumPerks struct {
 	ManageDashboardViaApi                          bool                `json:"manage_dashboard_via_api"`
 	BulkAdding                                     bool                `json:"bulk_adding"`
 	ChartHistorySeconds                            ChartHistorySeconds `json:"chart_history_seconds"`
+	RewardsChartHistorySeconds                     ChartHistorySeconds `json:"rewards_chart_history_seconds"`
```
LuccaBitfly, line 131: Thought: If we add more chart types than these 2 in future, we should think of having them in an array. For now this is fine.

backend/pkg/api/handlers/handler_service.go
```diff
@@ -184,7 +193,7 @@ func (h *HandlerService) getCurrentChartTimeLimitsForDashboard(ctx context.Conte
 		return limits, err
 	}
 	limits.MinAllowedTs = limits.LatestExportedTs - min(maxAge, limits.LatestExportedTs)                        // min to prevent underflow
-	secondsPerEpoch := uint64(12 * 32)                                                                          // TODO: fetch dashboards chain id and use correct value for network once available
+	secondsPerEpoch := utils.Config.Chain.ClConfig.SlotsPerEpoch * utils.Config.Chain.ClConfig.SecondsPerSlot   // TODO: fetch dashboards chain id and use correct value for network once available
```
LuccaBitfly: Issue: I've been actively trying to avoid relying on global states in v2. If you insist on this change, pls add the config as a field of the `HandlerService` struct and pass it in the constructor in `api/router.go`.

backend/pkg/api/handlers/public.go
```diff
@@ -610,12 +610,7 @@ func (h *HandlerService) PublicPostValidatorDashboardValidators(w http.ResponseW
 		handleErr(w, r, err)
 		return
 	}
-	var limit uint64
-	if isUserAdmin(userInfo) {
-		limit = math.MaxUint32 // no limit for admins
-	} else if dashboardLimit >= existingValidatorCount {
-		limit = dashboardLimit - existingValidatorCount
-	}
+	limit := dashboardLimit - existingValidatorCount
```
LuccaBitfly: Question: Why did you remove the limit for admins?

backend/pkg/api/handlers/public.go
```diff
@@ -1323,26 +1318,40 @@ func (h *HandlerService) PublicGetValidatorDashboardGroupRewards(w http.Response
 //	@Tags			Validator Dashboard
 //	@Produce		json
 //	@Param			dashboard_id	path		string	true	"The ID of the dashboard."
+//	@Param			group_ids		query		string	false	"Provide a comma separated list of group IDs to filter the results by."
+//	@Param			aggregation		query		string	false	"Aggregation type to get data for."	Enums(epoch, hourly, daily, weekly)	Default(hourly)
 //	@Param			modes			query		string	false	"Provide a comma separated list of protocol modes which should be respected for validator calculations. Possible values are `rocket_pool``."
+//	@Param			after_ts		query		string	false	"Return data after this timestamp."
+//	@Param			before_ts		query		string	false	"Return data before this timestamp."
```
LuccaBitfly: Question: IIRC this interval should be half open, i.e. `after_ts` is inclusive whereas `before_ts` is exclusive (this would be my interpretation from the doc rn). I believe this is the case for the summary chart. Have you confirmed this for the rewards?

LuccaBitfly (changes requested):

backend/pkg/api/handlers/handler_service.go
remoterami: Well no I don't, but I don't think using magic constants is better either. With gnosis coming up this should already lead to wrong results I think.

I agree that this is also not a permanent solution (so I didn't touch the comment), but for now I'd say it's a better temporary one. I can apply your suggestion with the config param though

backend/pkg/api/handlers/public.go
remoterami: Because I think it's redundant. Realized that this should already be set to `maxJsInt` via https://github.com/gobitfly/beaconchain/blob/BEDS-875/rewards-chart/backend/pkg/commons/db/user.go#L268-L272

backend/pkg/api/handlers/public.go
remoterami: Good point, this should probably be documented more explicitly and handled carefully. Currently it's inclusive on both ends, but it's the same for the summary charts endpoint (https://github.com/gobitfly/beaconchain/blob/BEDS-875/rewards-chart/backend/pkg/api/data_access/vdb_summary.go#L1049). I don't really mind which way it should be, do you have a preference? Otherwise could ask frontend/consumers

backend/pkg/api/enums/enums.go
```diff
@@ -202,6 +202,20 @@ var ValidatorDuties = struct {
...
+// Chart Types
+
+type ChartType int
+
+const (
+	ChartDefault ChartType = iota
+	ChartRewards
+)
+
+func (d ChartType) Int() int {
```
sasha-bitfly: What is the int value that we pass is invalid?

Should we have a function that checks the valid number
```
func (d ChartType) IsValid() bool {
    return d >= ChartDefault && d <= ChartRewards
}
```

and maybe a function that stringify for debugging and logging purposes?
```
func (d ChartType) String() string {
    switch d {
    case ChartDefault:
        return "Default"
    case ChartRewards:
        return "Rewards"
    default:
        return "Unknown"
    }
}
```

sasha-bitfly (approved):

sasha-bitfly:
@remoterami Ideally to cover the changes with tests. Unit tests are nice, but api [tests](https://github.com/gobitfly/beaconchain/blob/staging/backend/pkg/api/api_test.go) can be fine.

backend/pkg/api/enums/enums.go
remoterami: >What is the int value that we pass is invalid?

This should be covered here already: https://github.com/gobitfly/beaconchain/blob/2ca36e5156bb3f5078064934ee80cbb2daf35c72/backend/pkg/api/enums/enums.go#L15-L17

>and maybe a function that stringify for debugging and logging purposes?

It's meant as an internal type only, thought it's cleaner than passing strings or numbers. It doesn't depend on user params and isn't logged anywhere, so I think that's not needed.

LuccaBitfly (approved):
lgtm

backend/pkg/api/handlers/public.go
LuccaBitfly: I also have no strong opinion on this, I think both being inclusive prob leads to least confusion.

remoterami:
> @remoterami Ideally to cover the changes with tests. Unit tests are nice, but api [tests](https://github.com/gobitfly/beaconchain/blob/staging/backend/pkg/api/api_test.go) can be fine.

I've added some tests and did some refactoring. Could probably further solidify and generalize the testing setup, but that depends on our coverage goals and I didn't want to over-engineer for now. Open to suggestions though

sasha-bitfly (approved):

LuccaBitfly:
rebased with staging

backend/pkg/api/enums/enums.go
LuccaBitfly: removed this from the PR since it is unused

backend/pkg/api/data_access/vdb_rewards.go
```diff
@@ -712,22 +702,12 @@ func (d *DataAccessService) GetValidatorDashboardRewardsChart(ctx context.Contex
...
-	// ------------------------------------------------------------------------------------------------------------------
-	// Build the query that serves as base for both the main and EL rewards queries
-	// CL
-	rewardsDs := goqu.Dialect("postgres").
+	clDs := goqu.Dialect("postgres").
 		Select(
 			goqu.L(`SUM(e.attestations_reward + e.blocks_cl_reward + e.sync_reward) AS cl_rewards`),
 			timeColumn.As("timestamp"),
+			goqu.L(fmt.Sprintf("min(%s)", epochStartCol)).As("epoch_start"),
+			goqu.L(fmt.Sprintf("max(%s)", epochEndCol)).As("epoch_end"),
```
remoterami:
```suggestion
			goqu.MIN(epochStartCol).As("epoch_start"),
			goqu.MAX(epochEndCol).As("epoch_end"),
```

backend/pkg/api/data_access/vdb_rewards.go
```diff
@@ -712,22 +702,12 @@ func (d *DataAccessService) GetValidatorDashboardRewardsChart(ctx context.Contex
...
 		).
 		From(dataTable).
 		With("validators", goqu.Dialect("postgres").
```
remoterami: **suggestion**: Might not be changed by your changes, but this CTE is only used if `dashboardId.Validators == nil`. Please move it to that section below (l. 735)

backend/pkg/api/data_access/vdb_rewards.go
```diff
@@ -712,22 +702,12 @@ func (d *DataAccessService) GetValidatorDashboardRewardsChart(ctx context.Contex
 		return nil, fmt.Errorf("unexpected aggregation type: %v", aggregation)
```
remoterami: **suggestion**: With the aggregation enum this default  branch can't be reached. I suggest removing it and changing the func signature to only return `*goqu.SelectDataset`, no error

backend/pkg/api/data_access/vdb_rewards.go
```diff
@@ -740,206 +720,198 @@ func (d *DataAccessService) GetValidatorDashboardRewardsChart(ctx context.Contex
...
 		ClRewards  int64     `db:"cl_rewards"`
-	}{}
-
-	query, args, err := rewardsDs.Prepared(true).ToSQL()
-	if err != nil {
-		return nil, fmt.Errorf("error preparing query: %w", err)
 	}
 
-	err = d.clickhouseReader.SelectContext(ctx, &queryResult, query, args...)
+	queryResult, err := runQueryRows[[]clResult](ctx, d.clickhouseReader, clDs)
```
remoterami: Pls do analog to `elQueryResult`:
```suggestion
	clQueryResult, err := runQueryRows[[]clResult](ctx, d.clickhouseReader, clDs)
```

remoterami (commented):
very nice refactoring, looks way better now. ptal, left some comments

backend/pkg/api/data_access/vdb_rewards.go
```diff
@@ -680,193 +683,238 @@ func (d *DataAccessService) GetValidatorDashboardGroupRewards(ctx context.Contex
...
+	case enums.IntervalEpoch:
+		dataTable = goqu.L("validator_dashboard_data_epoch AS e")
+		timeColumn = goqu.C("epoch_timestamp")
+		epochStartCol, epochEndCol = "e.epoch", "e.epoch"
+	case enums.IntervalHourly:
+		dataTable = goqu.L("validator_dashboard_data_hourly AS e FINAL")
+	case enums.IntervalDaily:
+		dataTable = goqu.L("validator_dashboard_data_daily AS e FINAL")
+	case enums.IntervalWeekly:
+		dataTable = goqu.L("validator_dashboard_data_weekly AS e FINAL")
```
invis-bitfly: please do not use `FINAL`, grouping on `timeColumn` is enough.

invis-bitfly (dismissed):

backend/pkg/api/data_access/vdb_rewards.go
```diff
@@ -680,193 +683,238 @@ func (d *DataAccessService) GetValidatorDashboardGroupRewards(ctx context.Contex
...
-	for _, groupId := range groupList {
-		for _, propertyName := range propertyNames {
-			result.Series = append(result.Series, t.ChartSeries[int, decimal.Decimal]{
-				Id:       int(groupId),
-				Property: propertyName,
-			})
-		}
+	// create the chart data
+	for _, epoch := range epochStarts {
+		ret.Categories = append(ret.Categories, uint64(utils.EpochToTime(epoch).Unix()))
```
invis-bitfly, line 889: question: why not use `clResult.Timestamp`?

backend/pkg/api/data_access/vdb_rewards.go
LuccaBitfly: Fixed, ptal

backend/pkg/api/data_access/vdb_rewards.go
LuccaBitfly: Fixed, ptal

backend/pkg/api/data_access/vdb_rewards.go
LuccaBitfly: Fixed, ptal

backend/pkg/api/data_access/vdb_rewards.go
LuccaBitfly: Fixed, ptal

backend/pkg/api/data_access/vdb_rewards.go
LuccaBitfly: Because using `epoch_start` IS accurate, right? It's not the same as `clResult.Timestamp`.

backend/pkg/api/data_access/vdb_rewards.go
LuccaBitfly: Fixed, ptal

remoterami:
> all my comments have been addressed

same, lgtm

PR #1228, opened 2025-01-02
BEDS 1078/block proposal notifc

LuccaBitfly:
- **refactor: split api block proposal success & miss notification events**
- **refactor: split frontend block proposal success & miss notification events**


marcel-bitfly (approved):
after resolving the conflict it looks good thx

PR #1229, opened 2025-01-02
(BEDS-1052) VDB: validator search by hex graffiti

remoterami:


backend/pkg/api/handlers/input_validation.go
```diff
@@ -34,8 +34,9 @@ var (
 	reEthereumAddress              = regexp.MustCompile(`^(0x)?[0-9a-fA-F]{40}$`)
 	reWithdrawalCredential         = regexp.MustCompile(`^(0x0[01])?[0-9a-fA-F]{62}$`)
 	reEnsName                      = regexp.MustCompile(`^.+\.eth$`)
-	reGraffiti                     = regexp.MustCompile(`^.{2,}$`)          // at least 2 characters, so that queries won't time out
-	reCursor                       = regexp.MustCompile(`^[A-Za-z0-9-_]+$`) // has to be base64
+	reGraffiti                     = regexp.MustCompile(`^.{2,}$`)                     // at least 2 characters, so that queries won't time out
+	reGraffitiHex                  = regexp.MustCompile(`^(0x)?([0-9a-fA-F]{2}){2,}$`) // at least 2 bytes, so that queries won't time out
```
LuccaBitfly: Question: Should this be a valid hex input?
```bash
> curl --location 'http://0.0.0.0:8081/api/i/search' \
                          --header 'Content-Type: application/json' \
                          --data '{
                          "input":"0x0000000000000000000000000000000000000000000000000000000000000000",
                          "networks":["holesky"]
                      }'
                      
{"data":[{"type":"validators_by_graffiti","chain_id":17000,"value":{"graffiti":"","hex":"0x0000000000000000000000000000000000000000000000000000000000000000","count":72496}}]}
```

LuccaBitfly (changes requested):

backend/pkg/api/handlers/input_validation.go
remoterami: Good question, clarified and the answer is "yes can do that for now". Can exclude empty graffitis later if we run into issues.

remoterami:
I added some validation checks for maximum graffiti size of 32 bytes (which actually is a chain parameter but I doubt anyone would be messing with this..).
Also since we're doing full match search only, the hex graffiti should only be accepted and returned as a fixed size string

backend/pkg/api/handlers/input_validation.go
LuccaBitfly: Issue: Frontend would pass the graffiti string, not hex, to the endpoint for adding validators to a dashboard by their graffiti and passing an empty graffiti counts as not passing it at all. See: https://github.com/gobitfly/beaconchain/blob/f698e5576bcda5c0e2163b86310e423f314deb01/backend/pkg/api/handlers/public.go#L567
image

I believe we have to forbid empty graffitis

LuccaBitfly (changes requested):

backend/pkg/api/handlers/input_validation.go
remoterami: Bummer! Applied, needs explicit handling though since look-arounds aren't a thing in go's `regexp`

LuccaBitfly (approved):
lgtm!

PR #1230, opened 2025-01-02
feat(db): clickhouse failover support

invis-bitfly:
also includes a switch to a more sane env lib so i dont have to define everything 45 times
first commit might not build, need to remove all the now useless casting everywhere

### DIFF
--- a/backend/cmd/eth1indexer/main.go
+++ b/backend/cmd/eth1indexer/main.go
@@ -120,25 +120,7 @@ func Run() {
 		}()
 	}
 
-	db.WriterDb, db.ReaderDb = db.MustInitDB(&types.DatabaseConfig{
-		Username:     cfg.WriterDatabase.Username,
-		Password:     cfg.WriterDatabase.Password,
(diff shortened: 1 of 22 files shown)


invis-bitfly:
seems to work with some basic testing:
- invalid host as primary, valid host as fallback: connection succeeds after a small delay (~8s)
- valid host as primary, invalid host as fallback: connection succeeds instantly, no attempt made to invalid host

should note that the logic only gets applied when a new connection gets opened. this means that a faulty host that accepts/holds connections but fails to execute any queries will continue to cause issues until the faulty host gets stopped/fixed.
a potential fix is setting the max connection lifetime to something low (5m?) and adding a [custom dial function](https://github.com/ClickHouse/clickhouse-go/pull/855) that does a sanity check each time to confirm that the host is safe to use. sounds annoying tho as you would only have the raw dial to do that.
at some point it makes more sense to just write our own wrapper tho which would allow for much more flexibility.

also this pr is breaking as it makes the ssl flag configurable instead of always forcing ssl with clickhouse.

PR #1233, opened 2025-01-06
(BEDS-1100) API: Incorporate dashboard archival status into notification endpoints

remoterami:


backend/pkg/api/router.go
```diff
@@ -353,10 +353,10 @@ func addNotificationRoutes(hs *handlers.HandlerService, publicRouter, internalRo
 
 	publicDashboardNotificationSettingsRouter := publicNotificationRouter.NewRoute().Subrouter()
 	internalDashboardNotificationSettingsRouter := internalNotificationRouter.NewRoute().Subrouter()
-	// TODO add adb auth middleware to account dashboard endpoints once they are implemented
+	// TODO add adb auth and archivedCheck middleware to account dashboard endpoints once they are implemented
 	if !debug {
-		publicDashboardNotificationSettingsRouter.Use(hs.VDBAuthMiddleware)
-		internalDashboardNotificationSettingsRouter.Use(hs.VDBAuthMiddleware)
+		publicDashboardNotificationSettingsRouter.Use(hs.VDBAuthMiddleware, hs.VDBArchivedCheckMiddleware)
+		internalDashboardNotificationSettingsRouter.Use(hs.VDBAuthMiddleware, hs.VDBArchivedCheckMiddleware)
```
LuccaBitfly: Issue: Preventing users from seeing the details modal for archived dashboards (`GET /users/me/notifications/validator-dashboards/{dashboard_id}/groups/{group_id}/epochs/{epoch}`) might be problematic if the notification for that dashboard still shows up in the table (`GET /users/me/notifications/dashboards`). We should probably do one of the following:
1. Make sure archived dashboards don't show up in the table
2. Make sure archived dashboards have a flag in the table, so that they can be greyed out
3. Simply let the details endpoint still be accessible for archived dashboards

backend/pkg/api/router.go
LuccaBitfly, line 1: Issue: Not your fault  but it seems the error msg returned is a bit broken, if you could fix that. Also, accessing the archived dashboard through a shared ID leaks the internal ID, if you could fix that 🙏
image

LuccaBitfly (changes requested):

LuccaBitfly:
Also if you're not adding integration tests now, pls create medium prio ticket for adding them.

backend/pkg/api/router.go
remoterami: You're right good find, I missed this. 3 should be the way to go imo, past notification events should still be accessible even if the dashboard is archived now

remoterami:
> Also if you're not adding integration tests now, pls create medium prio ticket for adding them.

I think that should be covered by 579 already

backend/pkg/api/router.go
LuccaBitfly: Issue: Still getting archived error. Keep in mind that you have to turn off the debug flag for the middlewares.
image

backend/pkg/api/handlers/middlewares.go
```diff
@@ -156,7 +156,7 @@ func (h *HandlerService) VDBArchivedCheckMiddleware(next http.Handler) http.Hand
 			return
 		}
 		if dashboard.IsArchived {
-			handleErr(w, r, newForbiddenErr("dashboard with id %v is archived", dashboardId))
+			handleErr(w, r, newForbiddenErr("dashboard with id %s is archived", mux.Vars(r)["dashboard_id"]))
```
LuccaBitfly, line 159: Thought (non-blocking): Accessing a dashboard through a shared ID will "leak" the info that the dashboard owner has archived it. Alternatively we could return a 404 and pretend the dashboard doesn't exist. Neither option seems better than the other though IMO.

LuccaBitfly (changes requested):

backend/pkg/api/handlers/middlewares.go
remoterami: I think at this point the harm of leaking information is smaller than the usability gained, but also no strong opinion

backend/pkg/api/router.go
remoterami: Of course, sorry fixed

LuccaBitfly (approved):
lgtm!

PR #1234, opened 2025-01-07
feat(exporter): use real timestamps in bigtable

peterbitfly:
use real timestamps reading part

fix various bugs

refractor, remove obsolete code

remove unused constant

fix wrong return parameter order


Bukksie:
/

guybrush (approved):

PR #1239, opened 2025-01-08
BEDS 1065/apps christmas miracle

LuccaBitfly:
Adds efficiency, balances and rewards field to validator dashboard group summary.
Rule-of-thumb performance estimates:
- Balances: adds 500ms - 2.5s per request depending on dashboard size
- Rewards: data is already there so no changes
- Efficiency: data is mostly already there so no changes

### DIFF
--- a/backend/pkg/api/data_access/vdb_helpers.go
+++ b/backend/pkg/api/data_access/vdb_helpers.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"database/sql"
 	"fmt"
+	"math"
 	"time"
 
 	"github.com/doug-martin/goqu/v9"
@@ -13,6 +14,7 @@ import (
 	"github.com/gobitfly/beaconchain/pkg/commons/utils"
 	"github.com/lib/pq"
 	"github.com/pkg/errors"
+	"github.com/shopspring/decimal"
 )
 
 //////////////////// 		Helper functions (must be used by more than one VDB endpoint!)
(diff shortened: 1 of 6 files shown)


backend/pkg/api/data_access/vdb_helpers.go
LuccaBitfly, line 1: if the changes here don't fit the intended usage for this file please lmk

backend/pkg/api/data_access/vdb_summary.go
```diff
@@ -825,11 +824,39 @@ func (d *DataAccessService) GetValidatorDashboardGroupSummary(ctx context.Contex
...
+	if err != nil {
+		return nil, err
+	}
+	ret.Balances = balances
 
 	return ret, nil
 }
 
-func (d *DataAccessService) internal_getElClAPR(ctx context.Context, dashboardId t.VDBId, groupId int64, hours int) (elIncome decimal.Decimal, elAPR float64, clIncome decimal.Decimal, clAPR float64, err error) {
+func (d *DataAccessService) getElClAPR(ctx context.Context, dashboardId t.VDBId, groupId int64, hours int) (elIncome decimal.Decimal, elAPR float64, clIncome decimal.Decimal, clAPR float64, err error) {
```
remoterami: That function could probably also be moved to `vdb_helpers.go`, it's used a few times in different locations :pray:

backend/pkg/api/data_access/vdb_summary.go
```diff
@@ -825,11 +824,39 @@ func (d *DataAccessService) GetValidatorDashboardGroupSummary(ctx context.Contex
 			ret.SyncCommittee.Validators = ret.SyncCommittee.Validators[:3]
 		}
 	}
+	var attestationEfficiency, proposerEfficiency, syncEfficiency sql.NullFloat64
+	if totalIdealAttestationRewards > 0 {
+		attestationEfficiency.Float64 = decimal.NewFromInt(totalAttestationRewards).Div(decimal.NewFromInt(totalIdealAttestationRewards)).InexactFloat64()
+		attestationEfficiency.Valid = true
+	}
```
remoterami: This was calculated above already, see `ret.AttestationEfficiency`

backend/pkg/api/data_access/vdb_summary.go
```diff
@@ -825,11 +824,39 @@ func (d *DataAccessService) GetValidatorDashboardGroupSummary(ctx context.Contex
...
+	validatorMapping, err := d.services.GetCurrentValidatorMapping()
+	if err != nil {
+		return nil, err
+	}
+	balances, err := d.calculateValidatorDashboardBalance(ctx, rpOperatorInfo, validators, validatorMapping, protocolModes)
+	if err != nil {
+		return nil, err
+	}
+	ret.Balances = balances
```
remoterami: **check**: This would factor in the `protocolModes` for the new balances field, but the rest of the function does not yet use the parameter. I'm not 100% certain which other fields would also be affected, but apr might be a candidate.
- If there are any, I would prefer to not return "mixed" data as that could be confusing. Rather either apply the calculation to all relevant fields, or (temporarily) apply it to none (hardcoding `protocolModes.RocketPool` to `false`)
- If there are none (which might actually be the case, could be that the parameter is just a leftover from when we were briefly planning to add RP stats to the group summaries or something), the comment about it at the top of the method should be removed / clarified.

remoterami (changes requested):
Looks good overall, left a few comments. ptal

backend/pkg/api/data_access/vdb_summary.go
LuccaBitfly: done, ptal

backend/pkg/api/data_access/vdb_summary.go
LuccaBitfly: adapted, ptal

backend/pkg/api/data_access/vdb_summary.go
LuccaBitfly: You're completely right. If you allow it, I'll create another ticket and keep this in scope for the minimum required app changes :pray:

remoterami: Sure, sounds good :+1:

remoterami (approved):
lgtm :gift: (just create ticket pls, see comment)

PR #1247, opened 2025-01-09
BEDS-1106/fix percent formatting

enzo-bitfly:
We were using two different percent formatting functions, which caused inconsistencies in the resulting values.
This MR replaces the uses of the older by the newer one and removes it.


marcel-bitfly (approved):
nice work thx for cleaning up our stuff 💪

PR #1258, opened 2025-01-13
BEDS 1124/prices

LuccaBitfly:
- **refactor: generalize util slice dedupe**
- **fix: show correct exchange rates for gnosis**

### DIFF
--- a/backend/pkg/api/data_access/header.go
+++ b/backend/pkg/api/data_access/header.go
@@ -72,13 +72,9 @@ func (d *DataAccessService) GetLatestBlockHeightsForEpoch(ctx context.Context, e
 func (d *DataAccessService) GetLatestExchangeRates(ctx context.Context) ([]t.EthConversionRate, error) {
 	result := []t.EthConversionRate{}
 
-	availableCurrencies := price.GetAvailableCurrencies()
+	availableCurrencies := utils.Deduplicate(price.GetAvailableCurrencies())
 	for _, code := range availableCurrencies {
-		if code == "ETH" {
-			// Don't return ETH/ETH info
-			continue
-		}
-		rate := price.GetPrice("ETH", code)
+		rate := price.GetPrice(d.config.Frontend.MainCurrency, code)
 		result = append(result, t.EthConversionRate{
 			Currency: price.GetCurrencyLabel(code),
 			Code:     code,
--- a/backend/pkg/api/services/service.go
+++ b/backend/pkg/api/services/service.go
@@ -46,7 +46,7 @@ func (s *Services) InitServices() {
 	go s.startEmailSenderService(wg)
 
 	log.Infof("initializing prices...")
-	price.Init(utils.Config.Chain.ClConfig.DepositChainID, utils.Config.Eth1ErigonEndpoint, utils.Config.Frontend.ClCurrency, utils.Config.Frontend.ElCurrency)
+	price.Init(utils.Config.Chain.ClConfig.DepositChainID, utils.Config.Eth1ErigonEndpoint, utils.Config.Frontend.MainCurrency, utils.Config.Frontend.ClCurrency, utils.Config.Frontend.ElCurrency)
 	log.Infof("...prices initialized")
 
 	wg.Wait()
--- a/backend/pkg/commons/price/price.go
+++ b/backend/pkg/commons/price/price.go
@@ -26,9 +26,9 @@ var prices = map[string]float64{}
 var pricesMu = &sync.Mutex{}
 var didInit = uint64(0)
 var feeds = map[string]*chainlink_feed.Feed{}
-var calcPairs = map[string]bool{}
-var clCurrency = "ETH"
-var elCurrency = "ETH"
+var mainCurrency = "ETH" // currency in which all exchange rates are denominated, e.g. ETH or GNO
+var clCurrency = "ETH"   // currency in which all CL values are denominated, e.g. ETH or mGNO
+var elCurrency = "ETH"   // currency in which all EL values are denominated, e.g. ETH or xDai
 
 var currencies = map[string]struct {
 	Symbol string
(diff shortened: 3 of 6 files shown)


backend/pkg/api/data_access/header.go
```diff
@@ -72,13 +72,9 @@ func (d *DataAccessService) GetLatestBlockHeightsForEpoch(ctx context.Context, e
...
 	result := []t.EthConversionRate{}
 
-	availableCurrencies := price.GetAvailableCurrencies()
+	availableCurrencies := utils.Deduplicate(price.GetAvailableCurrencies())
 	for _, code := range availableCurrencies {
-		if code == "ETH" {
-			// Don't return ETH/ETH info
-			continue
-		}
-		rate := price.GetPrice("ETH", code)
```
remoterami, line 81: Shouldn't we keep that? Skip if `code == d.mainCurrency`

backend/pkg/api/services/service.go
```diff
@@ -46,7 +46,7 @@ func (s *Services) InitServices() {
 	go s.startEmailSenderService(wg)
 
 	log.Infof("initializing prices...")
-	price.Init(utils.Config.Chain.ClConfig.DepositChainID, utils.Config.Eth1ErigonEndpoint, utils.Config.Frontend.ClCurrency, utils.Config.Frontend.ElCurrency)
+	price.Init(utils.Config.Chain.ClConfig.DepositChainID, utils.Config.Eth1ErigonEndpoint, utils.Config.Frontend.MainCurrency, utils.Config.Frontend.ClCurrency, utils.Config.Frontend.ElCurrency)
```
remoterami, line 49: Suggestion: At some point we could probably start passing around a `Currencies` struct

backend/pkg/commons/price/price.go
```diff
@@ -104,34 +99,16 @@ func Init(chainId uint64, eth1Endpoint, clCurrencyParam, elCurrencyParam string)
...
 		availableCurrencies = []string{"ETH", "USD", "EUR", "GBP", "CNY", "CAD", "AUD", "JPY"}
-	case 11155111:
-		// see: https://docs.chain.link/data-feeds/price-feeds/addresses/
-		feedAddrs["ETH/USD"] = "0x694AA1769357215DE4FAC081bf1f309aDC325306"
-		feedAddrs["EUR/USD"] = "0x1a81afB8146aeFfCFc5E50e8479e826E7D55b910"
-		feedAddrs["JPY/USD"] = "0x8A6af2B75F23831ADc973ce6288e5329F63D86c6"
-		feedAddrs["GBP/USD"] = "0x91FAB41F5f3bE955963a986366edAcff1aaeaa83"
-		feedAddrs["AUD/USD"] = "0xB0C712f98daE15264c8E26132BCC91C40aD4d5F9"
-
-		availableCurrencies = []string{"ETH", "USD", "EUR", "GBP", "AUD", "JPY"}
```
remoterami, line 117: question: Why is the sepolia price feed  retrieval removed? Couldn't it be useful for testing

backend/pkg/commons/price/price.go
```diff
@@ -177,19 +151,21 @@ func updatePrices() {
...
+
 	setPrice(elCurrency, elCurrency, 1)
 	setPrice(clCurrency, clCurrency, 1)
 
 	runOnce.Do(func() { runOnceWg.Done() })
 }
 
-func calcPricePairs(currency string) error {
+// calculates all prices in a given currency
+func addPricePairs(currency string) error {
```
remoterami: nitpick: `addPricePairs` is not that much better as we're not adding new pairs most of the time (just update). Since that internal function is only used once I'd probably merge it into the caller (which also has a very related purpose), but could also rename to something like `convertPricesToBase`

remoterami (changes requested):
prices more accurate than Binance :+1: Some minor comments, ptal

backend/pkg/api/data_access/header.go
LuccaBitfly: With the introduction of the "main currency" concept our API now must always provide the exchange rate of Main/CL and Main/EL, which on mainnet + other Ethereum testnets means providing the ETH/ETH rate, so we need that. In case of Gnosis, IMO it can be also nice to see at a glance that the exchange rates are denominated in GNO and not ETH. If you insist I can remove it.

backend/pkg/api/services/service.go
LuccaBitfly: True, in the next task I'm planning I'll have to introduce that anyways. I think thematically it fits better in that PR.

backend/pkg/commons/price/price.go
LuccaBitfly: Currently all networks except mainnet and Gnosis get an early return at the beginning of the function, which makes this unreachable code. https://github.com/gobitfly/beaconchain/blob/f48e30e1052f2a3a9e039b130a16d0bb7f21e47b/backend/pkg/commons/price/price.go#L61-L69. We can easily add it again at a later time, if we want to.

backend/pkg/commons/price/price.go
LuccaBitfly: good idea :) merged it to the caller, ptal

backend/pkg/api/data_access/header.go
remoterami: Oh you're right, I see now, makes total sense

remoterami (approved):
lgtm!

Unrelated to your changes, but in `addPricePairs` (which you merged) I think that `pricesCopy` doesn't really do anything (bcus maps are ref types in go, can't create a deep copy like that). If you agree you could clean that up, but shouldn't cause any harm if not. See https://github.com/gobitfly/beaconchain/pull/1258/files#diff-b341d664d42e9927a41a4d294042b57ffd5461d62357ed14e6022352b5f840b0R158

PR #1260, opened 2025-01-13
BEDS-1104/fix adding validators via withdrawal address

enzo-bitfly:
**Info:**
The `ValidatorUpdateBody` type definition fixed in this PR is currently declared in the frontend, as there's no way to access it from the `api type definitions` right now. Making these types available was requested from the backend and should be added soon.


marcel-bitfly (approved):
thx 💪

PR #1269, opened 2025-01-14
refactor: replace hardcoded with a calculated `network finality` info

marcel-bitfly:


frontend/stores/useNetworkStore.ts
```diff
@@ -29,9 +34,14 @@ export function useNetworkStore() {
   }
 
   const getTimestampFromEpoch = (epoch: number) => {
-    return networkInfo.value.timeStampSlot0 + epoch * networkInfo.value.slotsPerEpoch * networkInfo.value.secondsPerSlot
+    return timeStampSlot0 + epoch * slotsPerEpoch * secondsPerSlot
   }
 
+  const numberOfEpochsTheNetworkIsConsideredToBeFinalized = 3
```
enzo-bitfly, line 40: Somehting like `amountOfEpochsToFinality` should be specific enough. Mentioning 'network' can be avoided, since the concept of `finality` is only applied to networks, no?

frontend/stores/useNetworkStore.ts
```diff
@@ -29,9 +34,14 @@ export function useNetworkStore() {
...
+  const secondsUntilNetworkFinality = computed(
+    () => slotsPerEpoch * secondsPerSlot * numberOfEpochsTheNetworkIsConsideredToBeFinalized,
```
enzo-bitfly: As a human linter, I would order by hierarchy here, so: `seconds-slots-epochs` :D

enzo-bitfly (commented):
Good job 💪

frontend/stores/useNetworkStore.ts
marcel-bitfly: Get it...I am also struggling with my wording, but I do want to state here that it is an `arbitrary value`.
`amountOfEpochsToFinality` still sounds like that is a thing that comes by design of the `EVM`

frontend/stores/useNetworkStore.ts
marcel-bitfly: I like that...thx

frontend/stores/useNetworkStore.ts
enzo-bitfly: `numberOfEpochsToAsumeFinality`?

marcel-bitfly: will keep it like that for now I think it gives more context

enzo-bitfly (approved):

PR #1272, opened 2025-01-14
fix: correctly calculate APR for gnosis

LuccaBitfly:
See: BEDS-1138


LuccaBitfly:
Values look similar enough on mainnet:
image

manuelsc (approved):
utACK

PR #1273, opened 2025-01-14
feat(raw): validate raw block data

Tangui-Bitfly:
* introduce raw store
* raw store add func validate the block data

I would encourage to check per commit:
* 311314ee244b4576a3654e53f98c0cb47d7661a6 introduce the raw store, it's quite chunky but it's code that I already use on other dev branch so it's well tested
* b03c9144b480f31c28d11383fc3219f17252baf2 is the real fix of this pr and is really small compare to the other commit


backend/pkg/commons/chain/chainID.go
```diff
@@ -0,0 +1,25 @@
...
+import (
+	"math/big"
+)
+
+type IDGetter struct {
+	Mainnet    *big.Int
+	Sepolia    *big.Int
+	Gnosis     *big.Int
+	Optimistic *big.Int
+	Arbitrum   *big.Int
```
Monika-Bitfly, line 12: Maybe we can also add `Holesky` to the list?

backend/pkg/commons/db2/database/remote.go
```diff
@@ -0,0 +1,315 @@
...
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+)
+
+const (
+	routeBulkAdd         = "/bulkAdd"
+	routeGetRowsRange    = "/rowRange"
```
Monika-Bitfly:
```suggestion
	routeGetRowsRange    = "/rowsRange"
```

Monika-Bitfly (approved):

PR #1274, opened 2025-01-15
fix: don't allow user deletion with active subscriptions

LuccaBitfly:
See: BEDS-924


remoterami (approved):
lgtm

LuccaBitfly:
tested on staging-2 with a temporary PR (#1289) as a dry run version of this, got `409` with an active subscription accound and `204` with a free tier account, as expected.

PR #1280, opened 2025-01-15
BEDS-175 / missed EL rewards for group summary

sasha-bitfly:


backend/pkg/api/data_access/vdb_summary.go
```diff
@@ -521,17 +521,76 @@ func (d *DataAccessService) GetValidatorDashboardGroupSummary(ctx context.Contex
...
+		// Define the `targets` CTE
+		targets := goqu.
+			From("blocks").
+			Join(
+				goqu.T("users_val_dashboards_validators").As("uvdv"),
+				goqu.On(goqu.I("blocks.proposer").Eq(goqu.I("uvdv.validator_index"))),
+			).
+			Where(
+				goqu.I("blocks.status").Neq("1"),
+				goqu.I("uvdv.dashboard_id").Eq(dashboardId.Id),
```
invis-bitfly: needs handling for when we dont have a dashboard id and instead have a list of validators
rest looks pretty good

invis-bitfly (changes requested):

backend/pkg/api/data_access/vdb_summary.go
```diff
@@ -521,17 +521,93 @@ func (d *DataAccessService) GetValidatorDashboardGroupSummary(ctx context.Contex
...
+		}
+
+		// Define the `res` CTE
+		res := goqu.
+			From("targets").
+			LeftJoin(
+				goqu.T("execution_rewards_finalized").As("b"),
+				goqu.On(
+					goqu.I("b.slot").Gte(goqu.L("targets.slot - 16")),
+					goqu.I("b.slot").Lt(goqu.L("targets.slot + 16")),
```
invis-bitfly: please replace this hardcoded 16 with `utils.Config.Chain.ClConfig.SlotsPerEpoch / 2` - this is needed for correct gnosis support

backend/pkg/api/data_access/vdb_summary.go
```diff
@@ -521,17 +521,93 @@ func (d *DataAccessService) GetValidatorDashboardGroupSummary(ctx context.Contex
...
+			GroupBy(goqu.I("targets.slot"))
+
+		// Build the final query
+		query := goqu.From("res").
+			With("targets", targets).
+			With("res", res).
+			Select(goqu.L("COALESCE(SUM(v), 0)"))
+
+		// Generate SQL and arguments
+		sql, args, err := query.ToSQL()
```
invis-bitfly:
```suggestion
		sql, args, err := query.Prepared(true).ToSQL()
```
please do this instead so goqu doesn't try to interpolate args into the query string and instead puts them into the args array. https://doug-martin.github.io/goqu/docs/interpolation.html

backend/pkg/api/data_access/vdb_summary.go
```diff
@@ -521,17 +521,93 @@ func (d *DataAccessService) GetValidatorDashboardGroupSummary(ctx context.Contex
 		return nil, err
 	}
 
-	getLastScheduledBlockAndSyncDate := func() (time.Time, time.Time, error) {
+	getMissedELRewards := func(epochStart, epochEnd uint64) (float64, error) {
+		// Initialize the result variable
+		var totalMissedRewardsEl float64
+
+		// Define the `targets` CTE
+		targets := goqu.
```
invis-bitfly: can you confirm that not specifying the dialect as postgres works correctly? https://doug-martin.github.io/goqu/docs/dialect.html

invis-bitfly (changes requested):
some minor things

backend/pkg/api/data_access/vdb_summary.go
sasha-bitfly: it works without goqu.dialect("postgres"), but I have added it

invis-bitfly (approved):
lgtm

PR #1282, opened 2025-01-16
(BEDS 1117) payment issue notification service

remoterami:
Based on #1263

- Needs migration of user db [backend/pkg/commons/db/migrations/postgres/20250115144049_users_subscription_payment_issue_notification_ts.sql](https://github.com/gobitfly/beaconchain/pull/1282/files#diff-724b385d4f50f6d99a3f18de4bddcf3a8f8eb0150dc9dfbe00dd46d096a56d9e)
- requires restart of `user-service`
  - assumes config value for `frontend.siteDomain` is `v2-beta-mainnet.beaconcha.in` (which likely requires a solution for BEDS-1032)

### DIFF
--- a/backend/pkg/userservice/subscription_end_reminder.go
+++ b/backend/pkg/userservice/subscription_end_reminder.go
@@ -1,0 +1,281 @@
+package userservice
+
+import (
+	"fmt"
+	"html/template"
+	"time"
+
+	"github.com/doug-martin/goqu/v9"
+	t "github.com/gobitfly/beaconchain/pkg/api/types"
+	"github.com/gobitfly/beaconchain/pkg/commons/db"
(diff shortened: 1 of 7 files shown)


backend/pkg/userservice/subscription_end_reminder.go
```diff
@@ -0,0 +1,281 @@
...
+}
+
+const reminderFrequency = utils.Week
+const gracePeriod = utils.Week * 2
+
+func SubscriptionEndReminder() {
+	for {
+		start := time.Now()
+
+		// get all subscriptions running on grace period which have'nt been warned recently
```
LuccaBitfly: typo:
```suggestion
		// get all subscriptions running on grace period which haven't been warned recently
```

LuccaBitfly (changes requested):

PR #1287, opened 2025-01-20
(BEDS-1162) Fix: don't increase mailsSent count if limit reached

remoterami:
The mail sender does check the `maxEmailsPerDay` limit, but it increases the count anyways which results in wrong numbers being shown.

Fixed & renamed the count setter + added a getter method for clarity.

### DIFF
--- a/backend/pkg/commons/mail/mail.go
+++ b/backend/pkg/commons/mail/mail.go
@@ -78,26 +78,21 @@ func createTextMessage(msg types.Email) string {
 // It will return a ratelimit-error if the configured ratelimit is exceeded.
 func SendMailRateLimited(content types.TransitEmailContent, maxEmailsPerDay int64, bucket string) error {
 	sendThresholdReachedMail := false
-	count, err := db.GetSentMessagesCount(context.Background(), bucket, content.UserId)
+	incr, newCount, err := db.IncrSentMessagesCount(bucket, content.UserId, 1, maxEmailsPerDay)
 	if err != nil {
-		return err
+		log.Error(err, "error incrementing sent email count", 0)
 	}
 	timeLeft := time.Until(time.Now().Add(utils.Day).Truncate(utils.Day))
 
-	log.Debugf("user %d has sent %d of %d emails today, time left is %v", content.UserId, count, maxEmailsPerDay, timeLeft)
-	if count >= maxEmailsPerDay {
+	log.Debugf("user %d has sent %d of %d emails today, time left is %v", content.UserId, newCount, maxEmailsPerDay, timeLeft)
+	if incr == 0 {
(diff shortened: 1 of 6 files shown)


backend/pkg/notification/queuing.go
```diff
@@ -424,6 +424,7 @@ func RenderEmailsForUserEvents(epoch uint64, notificationsByUserID types.Notific
 
 func QueueEmailNotifications(epoch uint64, notificationsByUserID types.NotificationsPerUserId, tx *sqlx.Tx) error {
 	// for emails multiple notifications will be rendered to one email per user for each run
+	// TODO filter out ratelimited users, don't need to queue or even render emails
 	emails, err := RenderEmailsForUserEvents(epoch, notificationsByUserID)
 	if err != nil {
```
remoterami, line 429: This would be the clean thing to do. Instead of storing rendered emails to db first and let the sender throw them away again later, doing it here with batch querying of user premium perks + email counts from redis would be a nice early out.

backend/pkg/commons/mail/mail.go
```diff
@@ -78,14 +78,14 @@ func createTextMessage(msg types.Email) string {
 // It will return a ratelimit-error if the configured ratelimit is exceeded.
 func SendMailRateLimited(content types.TransitEmailContent, maxEmailsPerDay int64, bucket string) error {
 	sendThresholdReachedMail := false
-	count, err := db.CountSentMessage(bucket, content.UserId)
+	count, err := db.GetSentMessagesCount(context.Background(), bucket, content.UserId)
```
peterbitfly: I don't think this is save, as multiple senders run at the same time, so some kind of locking needs to be implemented here.

peterbitfly (changes requested):
Please also include a section on how those changes were tested / validated.

remoterami:
> Please also include a section on how those changes were tested / validated.

I ran some manual tests, isolating the affected methods and pointing to a local redis instance. But I could also try to setup unit testing for the notification sender if preferred.
(I'm not sure what the performance requirements are though, the best implementation could depend on the amount of users, networks, notifications etc)

backend/pkg/commons/mail/mail.go
remoterami: You're right. I went back to the original approach of increasing the count in redis before sending the mail, but added an upper limit using optimistic locking (see [cas](https://redis.io/docs/latest/develop/interact/transactions/#cas)). I assume this should solve the issue for us.

If actual locking should be required, I would suggest integrating redis' [distributed-locks](https://redis.io/docs/latest/develop/use/patterns/distributed-locks/) (instead of some psql locking) as it's close to the data in question here and prob faster; however that'd add another dependency. ptal

peterbitfly (approved):

PR #1292, opened 2025-01-22
Beds 536/port v1 changes

Tangui-Bitfly:
Replicate changes from this v1 PR https://github.com/gobitfly/eth2-beaconchain-explorer/pull/2965


backend/pkg/executionlayer/transformer.go
```diff
@@ -264,10 +272,27 @@ func TransformContract(chainID string, block *types.Eth1Block, res *db2.IndexedB
 	return nil
 }
 
+func isReverted(internal *types.Eth1InternalTransaction, revertSource *string) bool {
+	var reverted bool
+	if internal.ErrorMsg != "" {
```
mccreedy-bitfly, line 277: Is a non-empty error message always a revert?

Tangui-Bitfly: Yes. We need to save the "source" of the revert because this internal tx can have some "children" that were successful but since their parent is reverted, they also are reverted.
For example let say we have this set of internal tx
```
0		// first internal tx
0 0		// child of the first internal tx
1		// second internal tx
1 0		// child of the second internal tx
```
In this example, let say internal [1] revert after the execution of [1 0]. In this case if you look at the error message, only internal [1] will have one. But its child [1 0] was reverted too. That way we extract the highest revert source, here [1], and check if the other internals are a child of this source.
Moreover in this example if [0] was successful, the transaction is not reverted as a whole. It's considerate as partially executed / reverted.
If you want to understand it better I can provide solidity example that will showcase those cases.

Monika-Bitfly (approved):

PR #1293, opened 2025-01-24
(BEDS-1173) using correct table for api keys

remoterami:


backend/pkg/api/data_access/user.go
```diff
@@ -47,21 +47,38 @@ func (d *DataAccessService) GetUserByEmail(ctx context.Context, email string) (u
 
 func (d *DataAccessService) CreateUser(ctx context.Context, email, password string) (uint64, error) {
 	// (password is already hashed)
-	var result uint64
-
-	apiKey, err := utils.GenerateRandomAPIKey()
+	var userId uint64
+	err := d.userWriter.GetContext(ctx, &userId, `
+    	INSERT INTO users (password, email, register_ts)
```
sasha-bitfly: Does it make sense to wrap SQL query in `goqu` same as we did for [test](https://github.com/gobitfly/beaconchain/pull/1293/files#diff-7c33b74599122872fbfed6ec69b4755324b9cfdef1f3674c622e3d57bd103736)

remoterami: For simpler queries like this it's usually not required as it's not really worth it, with the additional error handling etc. readability isn't really increased. I'd consider it optional in such cases but if you insist, I can change it

sasha-bitfly: LGTM as soon as we added test

sasha-bitfly (approved):

backend/pkg/api/data_access/user.go
sasha-bitfly: I think the benefit of goqu is to validate the query before executing it. Wrapping or not is up to you.

PR #1296, opened 2025-01-27
Beds 1177/pass correct value to name argument when updating notifications paired device

enzo-bitfly:


frontend/stores/notifications/useNotificationsManagementStore.ts
```diff
@@ -55,22 +55,24 @@ export const useNotificationsManagementStore = defineStore('notifications-manage
...
+    name?: string,
     value: boolean,
   }) => {
     await fetch<PutUserNotificationSettingsPairedDevicesResponse>(
       'NOTIFICATIONS_MANAGEMENT_PAIRED_DEVICES_SET_NOTIFICATION',
       {
         body: {
           is_notifications_enabled: value,
-          name: id,
+          ...(name && { name }),
```
marcel-bitfly: 🤔 this might be too smart...can you think of an alternative?

enzo-bitfly: It's a pretty common syntax to add a key only if the value exists.
If not using this, I'd have to first create an object only with `is_notifications_enabled: value`, and then add `name` only if it exists, and then pass that object in the `request body` - It would be more verbose and not necessarily clearer.

enzo-bitfly: Something like:
```
    const requestBody = {
      is_notifications_enabled: value,
    }
    if (deviceName) requestBody.name = deviceName
 ```

Comment (marcel-bitfly) [inline frontend/stores/notifications/useNotificationsManagementStore.ts]:
```suggestion
          name,
```
I think this should work too, as `ofetchs serializer` is smart enough to omit `name: undefined`

enzo-bitfly: Good to know 👌

PR #1300, opened 2025-01-28
BEDS 1140/vali group

LuccaBitfly:
- if a validator was part of a dashboard and that dashboard had reached its limit, the validators group couldn't be changed when providing an index list
- fixed by this PR

### DIFF
--- a/backend/pkg/api/data_access/vdb_management.go
+++ b/backend/pkg/api/data_access/vdb_management.go
@@ -827,6 +827,28 @@ func (d *DataAccessService) AddValidatorDashboardValidators(ctx context.Context,
 	return result, nil
 }
 
+func (d *DataAccessService) GetValidatorDashboardValidatorsOfList(ctx context.Context, dashboardId t.VDBIdPrimary, validators []t.VDBValidator) ([]t.VDBValidator, error) {
+	var result []uint64
+
+	ds := goqu.Dialect("postgres").
+		Select(goqu.L("DISTINCT uvdv.validator_index")).
+		From(goqu.I("users_val_dashboards_validators").As("uvdv")).
+		Where(goqu.L("uvdv.dashboard_id = ?", dashboardId)).
+		Where(goqu.L("uvdv.validator_index = ANY(?)", pq.Array(validators)))
+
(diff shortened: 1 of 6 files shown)


backend/pkg/commons/utils/utils.go
```diff
@@ -241,10 +241,10 @@ func ConstantTimeDelay(start time.Time, intendedMinWait time.Duration) {
 	}
 }
 
-func SliceToMap[T comparable](s []T) map[T]bool {
-	m := make(map[T]bool)
+func SliceToMap[T comparable](s []T) map[T]struct{} {
```
sasha-bitfly, line 244: @LuccaBitfly Please, write a unit test for this function

backend/pkg/api/data_access/vdb_management.go
```diff
@@ -827,6 +827,21 @@ func (d *DataAccessService) AddValidatorDashboardValidators(ctx context.Context,
 	return result, nil
 }
 
+func (d *DataAccessService) GetValidatorDashboardValidatorsOfList(ctx context.Context, dashboardId t.VDBIdPrimary, validators []t.VDBValidator) ([]t.VDBValidator, error) {
+	var result []uint64
+	q := `SELECT DISTINCT uvdv.validator_index
```
sasha-bitfly: Should we wrap sql query in goqu?

sasha-bitfly (changes requested):

sasha-bitfly:
Added minor comments, otherwise LGFM :+1:

backend/pkg/commons/utils/utils.go
sasha-bitfly: @remoterami @LuccaBitfly 
utils is the best file for us to start covering with the unit tests
especially I've seen places in a code where / 0 is possible

LuccaBitfly: done

backend/pkg/api/data_access/vdb_management.go
LuccaBitfly: done

sasha-bitfly (approved):

PR #1301, opened 2025-01-29
refactor: separate validation, response and business logic

LuccaBitfly:
Adds a pattern for splitting the HTTP logic from the business logic in the API handler funcs. Adds a few example test cases. Changes the test workflow to include these tests.

### DIFF
--- a/backend/pkg/api/handlers/handler_service.go
+++ b/backend/pkg/api/handlers/handler_service.go
@@ -12,6 +12,7 @@ import (
 	"strings"
 
 	"github.com/gobitfly/beaconchain/pkg/commons/log"
+	"github.com/gorilla/mux"
 	"github.com/invopop/jsonschema"
 
 	"github.com/alexedwards/scs/v2"
@@ -50,7 +51,9 @@ func NewHandlerService(dataAccessor dataaccess.DataAccessor, dummy dataaccess.Da
 // if the request is mocked, the data access dummy is returned; otherwise the data access service.
 // should only be used if getting mocked data for the endpoint is appropriate
 func (h *HandlerService) getDataAccessor(ctx context.Context) dataaccess.DataAccessor {
-	if isMocked, ok := ctx.Value(types.CtxIsMockedKey).(bool); ok && isMocked {
+	isMocked, isMockedOk := ctx.Value(types.CtxIsMockedKey).(bool)                         // set in StoreIsMockedFlagMiddleware
+	isMockingAllowed, isMockingAllowedOk := ctx.Value(types.CtxIsMockingAllowedKey).(bool) // set in Handle function
+	if isMockedOk && isMocked && isMockingAllowedOk && isMockingAllowed {
 		return h.daDummy
 	}
 	return h.daService
@@ -58,6 +61,45 @@ func (h *HandlerService) getDataAccessor(ctx context.Context) dataaccess.DataAcc
 
 // all networks available in the system, filled on startup in NewHandlerService
 var allNetworks []types.NetworkInfo
+
+type InputValidator[T any] interface {
+	Validate(params map[string]string, payload io.ReadCloser) (T, error)
+}
+
+type BusinessLogicFunc[Input any, Response any] func(ctx context.Context, input Input) (Response, error)
+
+func Handle[Input InputValidator[Input], Response any](defaultCode int, logicFunc BusinessLogicFunc[Input, Response], isMockingAllowed bool) func(w http.ResponseWriter, r *http.Request) {
+	return func(w http.ResponseWriter, r *http.Request) {
+		// prepare input
+		vars := mux.Vars(r)
+		q := r.URL.Query()
+		for k, v := range q {
+			if _, ok := vars[k]; ok || len(v) == 0 {
+				continue
+			}
+			vars[k] = v[0]
+		}
+		// input validation
+		var i Input
+		input, err := i.Validate(vars, r.Body)
+		if err != nil {
+			handleErr(w, r, err)
+			return
+		}
+		ctx := r.Context()
+		if isMockingAllowed {
+			ctx = context.WithValue(ctx, types.CtxIsMockingAllowedKey, true)
+		}
+		// business logic
+		response, err := logicFunc(ctx, input)
+		if err != nil {
+			handleErr(w, r, err)
+			return
+		}
+
+		writeResponse(w, r, defaultCode, response)
+	}
+}
 
 // --------------------------------------
 // errors
--- a/backend/pkg/api/handlers/handler_service.go
+++ b/backend/pkg/api/handlers/handler_service.go
@@ -2,7 +2,6 @@ package handlers
 
 import (
 	"context"
-	"encoding/base64"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -123,33 +122,6 @@ type validatorSet struct {
 	PublicKeys []string
 }
 
-// parseDashboardId is a helper function to validate the string dashboard id param.
-func parseDashboardId(id string) (interface{}, error) {
-	var v validationError
-	if reInteger.MatchString(id) {
-		// given id is a normal id
-		id := v.checkUint(id, "dashboard_id")
-		if v.hasErrors() {
-			return nil, v
-		}
-		return types.VDBIdPrimary(id), nil
-	}
-	if reValidatorDashboardPublicId.MatchString(id) {
-		// given id is a public id
-		return types.VDBIdPublic(id), nil
-	}
-	// given id must be an encoded set of validators
-	decodedId, err := base64.RawURLEncoding.DecodeString(id)
-	if err != nil {
-		return nil, newBadRequestErr("given value '%s' is not a valid dashboard id", id)
-	}
-	indexes, publicKeys := v.checkValidatorList(string(decodedId), forbidEmpty)
-	if v.hasErrors() {
-		return nil, newBadRequestErr("given value '%s' is not a valid dashboard id", id)
-	}
-	return validatorSet{Indexes: indexes, PublicKeys: publicKeys}, nil
-}
-
 // getDashboardId is a helper function to convert the dashboard id param to a VDBId.
 // precondition: dashboardIdParam must be a valid dashboard id and either a primary id, public id, or list of validators.
 func (h *HandlerService) getDashboardId(ctx context.Context, dashboardIdParam interface{}) (*types.VDBId, error) {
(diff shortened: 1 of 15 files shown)


sasha-bitfly (commented):
@LuccaBitfly 

Linter job fails:

```

Check failure on line 100 in backend/pkg/api/handlers/input_validation.go
printf: non-constant format string in call to
github.com/gobitfly/beaconchain/pkg/api/handlers.newBadRequestErr 

```

```

Check failure on line 156 in backend/pkg/api/handlers/input_validation.go
unnecessary leading newline (whitespace)

```

backend/pkg/api/handlers/handler_service.go
```diff
@@ -59,6 +60,41 @@ func (h *HandlerService) getDataAccessor(ctx context.Context) dataaccess.DataAcc
 // all networks available in the system, filled on startup in NewHandlerService
 var allNetworks []types.NetworkInfo
 
+type Checker[T any] interface {
+	Check(params map[string]string, payload io.ReadCloser) (T, error)
+}
```
remoterami: Imo it's worth re-thinking naming for the interface and method name here, `Check(er)` is a bit vague. It's only supposed to be used in the context of input validation, right? Maybe `InputValidator` and `Validate()` or so

backend/pkg/api/handlers/input_validation.go
```diff
@@ -142,14 +149,8 @@ func (v *validationError) checkUserEmailToken(token string) string {
 
 // check request structure (body contains valid json and all required parameters are present)
 // return error only if internal error occurs, otherwise add error to validationError and/or return nil
-func (v *validationError) checkBody(data interface{}, r *http.Request) error {
-	// check if content type is application/json
-	if contentType := r.Header.Get("Content-Type"); !reJsonContentType.MatchString(contentType) {
-		v.add("request body", "'Content-Type' header must be 'application/json'")
-	}
```
remoterami, line 149: Couldn't this check be kept somehow? Maybe move it to middleware, like if there is a body present it should be `application/json`

backend/pkg/api/handlers/input_validation.go
```diff
@@ -142,14 +149,8 @@ func (v *validationError) checkUserEmailToken(token string) string {
...
-
-	bodyBytes, err := io.ReadAll(r.Body)
-	r.Body = io.NopCloser(bytes.NewReader(bodyBytes)) // unconsume body for error logging
+func (v *validationError) checkBody(data interface{}, requestBody io.ReadCloser) error { // <- should be checkBody after refactoring
```
remoterami:
```suggestion
func (v *validationError) checkBody(data interface{}, requestBody io.ReadCloser) error {
```

backend/pkg/api/handlers/input_validation.go
```diff
@@ -560,3 +561,28 @@ func (v *validationError) checkTimestamps(r *http.Request, chartLimits ChartTime
 		return afterTs, beforeTs
 	}
 }
+
+func (v *validationError) checkDashboardId(id string) interface{} {
```
remoterami, line 565: This looks essentially like the same as https://github.com/gobitfly/beaconchain/blob/staging/backend/pkg/api/handlers/handler_service.go#L85-L109 to me, what's the reason for this? Would be cool if it was possible to merge them

backend/pkg/api/handlers/validator_dashboard.go
```diff
@@ -0,0 +1,112 @@
...
+
+import (
+	"context"
+	"io"
+
+	"github.com/gobitfly/beaconchain/pkg/api/enums"
+	"github.com/gobitfly/beaconchain/pkg/api/types"
+)
+
+// PublicPostValidatorDashboardGroups godoc
```
remoterami: check: I think doc generation doesn't work anymore since the method name changed

backend/pkg/api/handlers/validator_dashboard.go
```diff
@@ -0,0 +1,112 @@
...
+	if err := v.checkBody(&req, body); err != nil {
+		return i, err
+	}
+	i.dashboardId = v.checkPrimaryDashboardId(params["dashboard_id"])
+	i.name = v.checkNameNotEmpty(req.Name)
+	return i, v.AsError()
+}
+
+func (h *HandlerService) PostValidatorDashboardGroups(ctx context.Context, input inputPostValidatorDashboardGroups) (*types.ApiDataResponse[types.VDBPostCreateGroupData], error) {
+	dataAccessor := h.getDataAccessor(ctx)
```
remoterami, line 42: This step will exist in virtually every business logic func, right? Could you think of a clean way to also inject this via `handler_service.Handle()`, I think that would be cool

backend/pkg/api/handlers/validator_dashboard_test.go
```diff
@@ -0,0 +1,132 @@
...
+		assert.NotNil(t, err)
+		assert.True(t, errors.Is(err, errConflict))
+	})
+}
+
+// ------------------------------------------------------------
+
+func stringAsBody(s string) io.ReadCloser {
+	return io.NopCloser(strings.NewReader(s))
+}
```
remoterami: I guess this will be used in other unit test files, too. Maybe such helpers should be placed in a new/existing common file (prob same for `handlerTestSetup()`)

backend/pkg/api/handlers/validator_dashboard_test.go
```diff
@@ -0,0 +1,132 @@
...
+
+// ------------------------------------------------------------
+
+func handlerTestSetup() (context.Context, *HandlerService) {
+	ctx := context.WithValue(context.Background(), types.CtxUserIdKey, uint64(1))
+	da := &dataAccessStub{}
+	return ctx, NewHandlerService(da, da, nil, false)
+}
+
+func TestPostValidatorDashboardGroups(t *testing.T) {
```
remoterami, line 69: Looking at the split of tests into input validation and business logic: Maybe both could be grouped under one test method which is responsible for the entire api call, and then either use subsubtests or just do input first and business logic after.

What do you think? Collapsing entire units at once could be nice dev qol

remoterami (changes requested):
If I got it right, `handlers/public.go` can be removed entirely once your suggestion is applied to all functions, right? That would be so great

backend/pkg/api/handlers/input_validation.go
LuccaBitfly: In essence only `checkDashboardId` uses the validation error method receiver like all other check funcs. The funcs inside `handler_service.go` should get removed once all handlers have been converted.

LuccaBitfly: On second thought I might as well remove it right now. done

backend/pkg/api/handlers/validator_dashboard.go
LuccaBitfly: No, this will only be used for functions we want to be able to be mocked, currently just validator dashboard and notification endpoints. Auth sensitive endpoints should still use use `h.daService`

backend/pkg/api/handlers/validator_dashboard_test.go
LuccaBitfly: Yes, I'd do this once we have multiple test files.

backend/pkg/api/handlers/validator_dashboard_test.go
LuccaBitfly: This kinda goes against the thought of splitting the handlers into smaller isolated functions so that they can be unit tested. I don't even know if that would count as unit tests anymore :thinking:

backend/pkg/api/handlers/handler_service.go
LuccaBitfly: done

backend/pkg/api/handlers/input_validation.go
LuccaBitfly: done

backend/pkg/api/handlers/validator_dashboard.go
LuccaBitfly: Unfortunately the doc comments must be directly above a function and aren't read if they're above a struct :roll_eyes:   fixed

backend/pkg/api/handlers/input_validation.go
LuccaBitfly: fixed

backend/pkg/api/handlers/validator_dashboard_test.go
remoterami: Wouldn't change test definition itself, just their semantic grouping into logical units. You can still run the individual unit tests via e.g. `go test -run="FunctionName/InputValidation/success"` if desired, nothing changing about that and it would still make use of the split handlers

backend/pkg/api/handlers/validator_dashboard.go
remoterami: Okay I see. I would still prefer more explicit handling of that then, using `h.getDataAccessor(ctx)` if mockable and `h.daService` if not is just not immediately obvious to new contributors. E.g. add a `mockable` flag to `Handle()` (or a generalized `properties` stringset param) and then do the injection, or at least rename `getDataAccessor` to signal its mocking-relation.

But if you insist I could also live with it, just trying to give as much input as possible for this foundational testing rework

frontend/types/api/validator_dashboard.ts
```diff
@@ -331,3 +331,6 @@ export interface PostValidatorDashboardValidatorsRequest {
   withdrawal_credential?: string;
   graffiti?: string;
 }
+export interface PostValidatorDashboardGroupsRequest {
+  name: string;
+}
```
remoterami, line 336: That's a nice thing to add, should really do this with all request structs over time (not in scope for this PR of course) :+1:

remoterami (approved):
Changes lgtm, thanks. I would still adjust minor things (see comments above), but those might just be personal preference

frontend/types/api/validator_dashboard.ts
LuccaBitfly: Will move all request schemas to their own types as I convert endpoint handlers :eyes:  (@enzo-bitfly)

backend/pkg/api/handlers/validator_dashboard_test.go
LuccaBitfly: after internal discussion, leaving tests for one func only

backend/pkg/api/handlers/validator_dashboard.go
LuccaBitfly: done

PR #1302, opened 2025-01-29
Refactor indexing

Tangui-Bitfly:
The main goal of this PR is to remove functionalities from `db` package and put them into there own object "easily" testable. Those changes aim at removing the basic indexing flow.
⚠️ Please do not focus on the code quality, I tried to change as little as possible to reduce risk of breaking something. The goal here is not to refactor everything but to separate things, remove coupling, introduce dependancies injections and a tests.

Overview of the new architecture
```
pkg
    commons
        db2
            database            // wrappers around database interractions
                databasetest    // test utility that provide an in mem database that database wrappers can use
            data                // handles all the interactions with the store `data` (indexing resulted objects)
            metadaupdates       // handles all the interactions with the store `metadaupdates` (balances, contracts, ...)
    executionlayer              // contains the indexer and the transformers      
```
Those are the main package to focus your review on.
Lot of the changed lines come from generated code around contract bindings. You can find those in `internal/contracts`, and I would suggest to not spend too much time on those beside maybe having a look at the readme here `internal/contracts/README.md`.


backend/cmd/store/main.go
```diff
@@ -0,0 +1,39 @@
...
+	table := args[2]
+	port := args[3]
+
+	bt, err := database.NewBigTable(project, instance, nil)
+	if err != nil {
+		panic(err)
+	}
+	remote := database.NewRemote(database.Wrap(bt, table))
+	go func() {
+		log.Info("starting remote raw store on port 8087")
```
Monika-Bitfly: To avoid confusion we should read the port number from the args here

backend/pkg/commons/db2/data/keys.go
```diff
@@ -0,0 +1,596 @@
...
+	Indexed *types.Eth1BlobTransactionIndexed
+	TxIndex int
+}
+
+func BlobToItems(chainID string, blobs []BlobWithIndex) (map[string][]database.Item, error) {
+	items := make(map[string][]database.Item)
+	for _, blob := range blobs {
+		if blob.TxIndex > txPerBlockLimit {
+			return nil, fmt.Errorf("unexpected number of transactions in block expected at most %d but got: %v, tx: %x", txPerBlockLimit, blob.TxIndex, blob.Indexed.GetHash())
+		}
```
Monika-Bitfly, line 108: Probably good to add validation for the transaction hash length

backend/pkg/executionlayer/transformer.go
```diff
@@ -0,0 +1,823 @@
...
+	}
+	return update, nil, nil
+}
+
+func (t *Transformer) Uncle(chainID string, block *types.Eth1Block) (map[string][]database.Item, map[string][]database.Item, error) {
+	updateMetadata := make(map[string][]database.Item)
+
+	var uncles []data.UncleWithIndexes
+	for i, uncle := range block.Uncles {
+		if i > 99 {
```
Monika-Bitfly: This value should probably be 10 instead of 99 since we’ve set the max uncle value to 10

backend/pkg/executionlayer/transformer.go
```diff
@@ -0,0 +1,823 @@
...
+	update, err := data.UnclesToItems(chainID, uncles)
+	if err != nil {
+		return nil, nil, err
+	}
+	return update, updateMetadata, nil
+}
+
+func (t *Transformer) Withdrawal(chainID string, block *types.Eth1Block) (map[string][]database.Item, map[string][]database.Item, error) {
+	updateMetadata := make(map[string][]database.Item)
+
```
Monika-Bitfly: I think we should add back the validation to throw an error if the number of block withdrawals exceeds MaxWithdrawalsPerPayload

backend/pkg/commons/db2/data/data.go
```diff
@@ -0,0 +1,222 @@
...
+		Value:   transfer.Value,
+		Asset:   hex.EncodeToString(transfer.TokenAddress),
+		From:    hex.EncodeToString(transfer.From),
+		To:      hex.EncodeToString(transfer.To),
+	}
+}
+
+func parseTx(tx *types.Eth1TransactionIndexed) *Interaction {
+	return &Interaction{
+		ChainID: "",
```
Monika-Bitfly, line 281:
```suggestion
		ChainID: "", // TODO parse chain ID
```

backend/pkg/commons/db2/data/data.go
```diff
@@ -0,0 +1,222 @@
...
+		Hash:    tx.Hash,
+		Method:  tx.MethodId,
+		Time:    tx.Time.AsTime(),
+		Type:    "Transaction",
+		Value:   tx.Value,
+		Asset:   "ETH",
```
Monika-Bitfly, line 287:
```suggestion
		Asset:   "ETH", // TODO parse asset from tx
```

backend/pkg/executionlayer/transformer_test.go
```diff
@@ -0,0 +1,1213 @@
...
+					Operator:     aliceAddress.Bytes(),
+					From:         aliceAddress.Bytes(),
+					To:           bobAddress.Bytes(),
+					TokenId:      []byte{1},
+					Value:        []byte{2},
+				},
+			},
+		},
+		// TODO add transfer bulk when bug fixed
+	}
```
mccreedy-bitfly, line 319: Whats this?

backend/pkg/commons/db2/data/data_external_test.go
```diff
@@ -0,0 +1,109 @@
...
+	/*
+		list of filter
+		data.ByMethod(method)
+		data.ByAsset(asset)
+		data.OnlyReceived()
+		data.OnlySent()
+		data.IgnoreTransactions()
+		data.IgnoreTransfers()
+		data.WithTimeRange(from, to)
+	*/
```
mccreedy-bitfly, line 58: Why commented out code?

backend/pkg/executionlayer/transformer.go
```diff
@@ -0,0 +1,708 @@
...
+		Number:     block.GetNumber(),
+		GasLimit:   block.GetGasLimit(),
+		GasUsed:    block.GetGasUsed(),
+		Time:       block.GetTime(),
+		BaseFee:    block.GetBaseFee(),
+		// Duration:               uint64(block.GetTime().AsTime().Unix() - previous.GetTime().AsTime().Unix()),
+		UncleCount:       uint64(len(block.GetUncles())),
+		TransactionCount: uint64(len(block.GetTransactions())),
+		// BaseFeeChange:          new(big.Int).Sub(new(big.Int).SetBytes(block.GetBaseFee()), new(big.Int).SetBytes(previous.GetBaseFee())).Bytes(),
+		// BlockUtilizationChange: new(big.Int).Sub(new(big.Int).Div(big.NewInt(int64(block.GetGasUsed())), big.NewInt(int64(block.GetGasLimit()))), new(big.Int).Div(big.NewInt(int64(previous.GetGasUsed())), big.NewInt(int64(previous.GetGasLimit())))).Bytes(),
```
mccreedy-bitfly, line 171: ?

backend/pkg/executionlayer/transformer.go
```diff
@@ -0,0 +1,708 @@
...
+		ParentHash: block.GetParentHash(),
+		UncleHash:  block.GetUncleHash(),
+		Coinbase:   block.GetCoinbase(),
+		Difficulty: block.GetDifficulty(),
+		Number:     block.GetNumber(),
+		GasLimit:   block.GetGasLimit(),
+		GasUsed:    block.GetGasUsed(),
+		Time:       block.GetTime(),
+		BaseFee:    block.GetBaseFee(),
+		// Duration:               uint64(block.GetTime().AsTime().Unix() - previous.GetTime().AsTime().Unix()),
```
mccreedy-bitfly, line 167: ?

mccreedy-bitfly (approved):
Generally looks OK. I would like to note that this CR is enormous, and so couldn't review it for functional changes. Since this review concentrates mostly on testing, I focused my efforts there, specifically on the structure and ensuring we were creating a good model that others could use moving forward.

Your suggestion to use table-based tests is a good one, and raises the quality and our confidence of the code as a whole. I have concerns however about the flexibility of the testing code if we were to for example change the underlying database model, and how much of these tests we would have to re-write. However, as we had 0 tests prior to this CR and because this is clearly already a measurable improvement, I believe that is out of scope and we can tackle that problem later.

Please avoid having "TODOs" in the code, as it is very unlikely we will go back and revisit them. Consider any TODOs as essentially final. Also, next time it would make sense to split such a large CR up in to smaller pieces, so that they are reviewable and can be iterated on more quickly. This saves time for both the author of the changes and the reviewers.

backend/pkg/executionlayer/transformer_test.go
Tangui-Bitfly: It's related to a bug I found but did not fix.
https://github.com/gobitfly/beaconchain/blob/refactor-indexing/backend/pkg/executionlayer/transformer.go#L398
To sum up, erc1155 can be transferred in batch, in that case we only save the last batched transfer because we overwrite our indexed log for each batch elem.
There is no easy fix to that because we store those indexed log with key `<tx_index>:<log_index>` but in this case all the batched transfer have the same log index. So even if we fix our indexing, it will be overwritten in the db.

backend/pkg/commons/db2/data/data_external_test.go
Tangui-Bitfly: it was more a test to showcase how to interact with the `data.Store`, I can remove it bring confusion.

backend/pkg/executionlayer/transformer.go
Tangui-Bitfly: copy pasted previous code

backend/pkg/executionlayer/transformer.go
Tangui-Bitfly: copy pasted previous code

PR #1303, opened 2025-01-29
Beds 1037/large dashboards should not load slot viz icons and absolute values

enzo-bitfly:
## Changes:
**SlotViz:**
- Add initial state and persistance to `SlotViz` icons
- Restructure component hierarchy for `slot-viz` section (duties toggle, group selector and slots)
- Prevent extra `slot-viz fetch` that happened between changing dashboard and loading group-selector
- Prevent horizontal scrollbar when the page content is at it's maximum width by making the horizontal paddings smaller (this was introduced when adapting for Gnosis, but it's a design error anyway, as the slot-viz horizontal padding is too small in the designs themselves)

**Summary table:**
- Add initial state and persist absolute/relative values

**ValidatorDashboardOverviewStore**
- Join store and composable that was placed in the same file, and adapt all usages to keep reactivity

### Restructure of `SlotViz` section:
Screenshot 2025-01-29 at 17 39 41


frontend/components/dashboard/ValidatorSlotViz.vue
```diff
@@ -36,7 +36,7 @@ watch(
       && (newValue[0] !== oldValue[0]
         || (newValue[1] as number[]).length !== (oldValue[1] as number[]).length)
     ) {
-      resetTick()
+    resetTick()
```
marcel-bitfly: fyi: you might not have configured your IDE properly...this should have been autofixed (so that such whitespace issues cannot even be committed by accident)

```json
# settings.json
   "editor.codeActionsOnSave": {
        "source.fixAll.eslint": "always"
    },
    "editor.formatOnSave": false,
```

frontend/stores/dashboard/useValidatorDashboardOverviewStore.ts
```diff
@@ -6,22 +6,17 @@ import type {
 } from '~/types/api/validator_dashboard'
 import type { DashboardKey } from '~/types/dashboard'
 
-const validatorOverviewStore = defineStore('validator_overview_store', () => {
-  const data = ref<null | undefined | VDBOverviewData>()
-  return { data }
-})
-
-export function useValidatorDashboardOverviewStore() {
+export const useValidatorDashboardOverviewStore = defineStore('validator_overview_store', () => {
```
marcel-bitfly: np: let's remove `store-suffix` for every store `one by one`

```...defineStore('validator-overview', () => ...)```

File: frontend/stores/dashboard/useValidatorDashboardOverviewStore.ts [hunk 3]
```diff
@@ -6,22 +6,17 @@ import type {
...
+  const overview = ref<null | undefined | VDBOverviewData>()
```
marcel-bitfly: plz have a look if this can really be `null | undefined` (whenever you encounter this pattern)....I doubt it to be `null` so I think this can be simplified to 
```const overview = ref<undefined | VDBOverviewData>()```

File: frontend/stores/dashboard/useValidatorDashboardOverviewStore.ts [hunk 4]
```diff
@@ -6,22 +6,17 @@ import type {
...
+  const loading = ref(false)
   const { fetch } = useCustomFetch()
-  const { data } = storeToRefs(validatorOverviewStore())
   const { clearCache: clearRewardDetails }
     = useAllValidatorDashboardRewardsDetailsStore()
 
-  const overview = computed(() => data.value)
-
   async function refreshOverview(key: DashboardKey) {
+    loading.value = true
```
marcel-bitfly: we do not need to manually handle loading state, plz use `useAsyncData` for this in the component you are using `refreshOverview()` (see: https://github.com/gobitfly/beaconchain/blob/BEDS-1037/large-dashboards-should-not-load-slot-viz-icons-and-absolute-values/frontend/components/notifications/NotificationsDashboardDialogEntity.vue#L28)

frontend/composables/useValidatorDashboardGroups.ts
marcel-bitfly, line 1: plz remove this composable and expose groups from `useValidatorDashboardOverviewStore()`.
There is no reason for this `composable` to exist.

Plz adapt `commit msg` accordingly. If you need support handling `interactive rebase` feel free to reach out.

frontend/.vscode/settings.json
```diff
@@ -13,6 +13,7 @@
         "DashboardChartSummary",
         "DashboardChartSummaryFilter",
         "DashboardGroupManagementModal",
+        "DashboardSlotViz",
```
marcel-bitfly, line 16: np: commit msg
(refactor(DashboardSlotViz): improve component structure and namings)

We should use the `scope` as a hint that this commit's changes are limited to the named `slot` 👉`DashboardSlotViz`

frontend/components/dashboard/slot-viz/DashboardSlotViz.vue
```diff
@@ -0,0 +1,223 @@
+<script setup lang="ts">
+import { faInfoCircle } from '@fortawesome/pro-regular-svg-icons'
+import { FontAwesomeIcon } from '@fortawesome/vue-fontawesome'
+import { formatNumber } from '~/utils/format'
+import { useNetworkStore } from '~/stores/useNetworkStore'
```
marcel-bitfly: plz remove `stores` are `autoimported` (this often happens with the `IDE autoimport feature`)

frontend/pages/playground.vue
```diff
@@ -103,12 +103,7 @@ const tabs: HashTabs = [
       default-tab="components"
       :use-route-hash="true"
     >
-      <template #tab-panel-slotviz>
-        <SlotVizViewer
-          v-if="slotVizData"
-          :data="slotVizData"
-        />
-      </template>
+      <template #tab-panel-slotviz />
```
marcel-bitfly: plz remove this

frontend/components/dashboard/slot-viz/DashboardSlotViz.vue
```diff
@@ -55,30 +55,34 @@ const currentSlotId = computed(() => {
...
-      oldValue
-      && (newValue[0] !== oldValue[0]
-        || (newValue[1] as number[]).length !== (oldValue[1] as number[]).length)
-    ) {
+  () =>
+    activeValidatorGroups.value,
+  () => {
+    selectedGroupIds.value = activeValidatorGroups.value.length > 1
+      ? activeValidatorGroups.value.map(group => group.id)
+      : []
```
marcel-bitfly, line 66: this can be simplified: 
```
 selectedGroupIds.value = activeValidatorGroups.value.map(group => group.id)
```

frontend/components/dashboard/slot-viz/DashboardSlotViz.vue
```diff
@@ -55,30 +55,34 @@ const currentSlotId = computed(() => {
...
   { immediate: true },
 )
+watch(
+  () => tick.value,
+  async () => {
+    refetchingSlotVizStoreData.value = true
+    await refreshSlotViz(dashboardKey.value, selectedGroupIds.value)
+    refetchingSlotVizStoreData.value = false
+  },
+)
```
marcel-bitfly: I think you can simplify these two watchers with something like this:

```
const {data, status} = useAsyncData(`validator-dashboard-slotviz`, 
() => refreshSlotViz(dashboardKey.value, selectedGroupIds.value),
{
  immediate: true,
  watch: tick,
}
)
```

frontend/components/dashboard/slot-viz/DashboardSlotVizGroupSelector.vue
```diff
@@ -62,7 +58,8 @@ watch(
 )
 watch(() => selectedGroupIds.value, () => {
   emit('updateSelectedGroupIds', selectedGroupIds.value)
-})
+},
+)
```
marcel-bitfly: plz double check for unnecessary changes like that in the future

marcel-bitfly:
np: commit message
(feat(DashboardSlotViz): calculate and persist slotviz visible duties
- default visible duties depend on the dashbnoard being `shared` and over or under 64 validators
- loading/intermediate states are added to prevent flashes on the visible duties when changing dashboards)

if this is the commit which solves the issue plz put a `reference` in:

```
feat(...

See: BEDS-1037
```

also I think all these commits kind of handle the same thing (reason for the code change) and should be one commit
image

frontend/components/dashboard/slot-viz/DashboardSlotViz.vue
```diff
@@ -103,6 +105,7 @@ watch(
       </BcTooltip>
 
       <DashboardSlotVizDutyVisibilityToggle
+        :initially-hide-visible="isSharedDashboard && !!isLargeDashboard"
```
marcel-bitfly: np: this is a very confusing `prop name`, that may be linked to the implementation, which I guess would be hard to refactor now, but maybe you can get getting rid of this or find a better name.

frontend/components/dashboard/slot-viz/DashboardSlotViz.vue
```diff
@@ -116,12 +119,23 @@ watch(
 
       <DashboardSlotVizGroupSelector
         v-if="activeValidatorGroups.length > 1"
+        :validator-groups="activeValidatorGroups"
```
marcel-bitfly, line 121: could this be named the same?

`: activeValidatorGroups`

frontend/components/dashboard/slot-viz/DashboardSlotViz.vue
```diff
@@ -116,12 +119,23 @@ watch(
...
         class="dashboard-slot-viz-group-selector"
         @update-selected-group-ids="(newGroupIdSelection) => selectedGroupIds = newGroupIdSelection"
       />
     </div>
 
     <div
+      v-if="(loadingSlotViz && !refetchingSlotViz) || (loadingOverview && !!dashboardKey)"
```
marcel-bitfly: We talked already how this refactoring worked out for you, so I understand that such things happen sometimes.

I just have to point out that this is too complex and can for sure be avoided in the future.

frontend/components/dashboard/slot-viz/DashboardSlotVizDutyVisibilityToggle.vue
```diff
@@ -133,4 +124,27 @@ watch(
...
+  animation: pulse 1.2s infinite;
+
+  &-content {
+    width: 100%;
+    height: 100%;
+    background-color: var(--container-border-color);
+    border-radius: var(--border-radius);
+  }
+
+  // this keyframe imitates Tailwind's 'pulse' animation until we install it.
```
marcel-bitfly, line 144: plz remove this comment, as it says what the code does which is superfluous
leave the reference to tailwind.css

frontend/stores/dashboard/useValidatorDashboardOverviewStore.ts
```diff
@@ -6,6 +6,8 @@ import type {
 } from '~/types/api/validator_dashboard'
 import type { DashboardKey } from '~/types/dashboard'
 
+const VALIDATOR_DASHBOARD_SIZE_THRESHOLD = 64
```
marcel-bitfly: maybe you can quickly mention the reasoning behind `64` ideally in the `variable name`,  in a comment or in yr commit message.

The reasoning could also be something like: `NUMBER_OF_VALIDATORS_A_DASHBOARD_IS_CONSIDERED_BIG`or `BIG_DASHBOARD_VALIDATOR_THRESHOLD`

(the important thing here is that we make it crystal clear, that this is set arbitrarily by us / a product decision and this does not come from any mechanism which is connected to our domain -> e.g.`EVM`)

frontend/stores/dashboard/useValidatorDashboardOverviewStore.ts
```diff
@@ -80,9 +82,16 @@ export const useValidatorDashboardOverviewStore = defineStore('validator_overvie
     weekly: (overview.value?.chart_history_seconds?.weekly ?? 0) > 0,
   }))
 
+  const isLargeDashboard = computed(() => {
+    if (!validatorCount.value) return
```
marcel-bitfly: if possible the `return types` should be stay `stable`
```    if (!validatorCount.value) return false```

Comment (marcel-bitfly) [reply frontend/stores/dashboard/useValidatorDashboardOverviewStore.ts]: if the `VALIDATOR_DASHBOARD_SIZE_THRESHOLD` is not referenced somewhere else, it should be defined inside `computed` (cohesiveness). As there will also be no noticeable `memory` or performance gains

**Review by marcel-bitfly — CHANGES_REQUESTED** [review]
I can imagine this was a very challenging task, although it sounded very easy.
Nice how deep you digged into the `architecture` of the concerned components and that you took the chance to simplify things and create a cleaner data flow.

Some things grew in complexity, that happens sometimes, especially when you are unfamiliar with the code base and everything is tightly coupled.

In general we should not add more complexity, but abort refactoring (thats also one benefit of a clean commit structure, that we could take commits out, e.g. into another branch). But this time it does not hurt us to stay with your current implementation.

Good work in respect to the unforeseeable complexity 💪

File: frontend/.vscode/settings.json (hunk 5 above)
Comment (enzo-bitfly) [reply frontend/.vscode/settings.json:16]: I rephrased the commit message and removed the `scope`

File: frontend/components/dashboard/slot-viz/DashboardSlotViz.vue (hunk 8 above)
Comment (enzo-bitfly) [reply frontend/components/dashboard/slot-viz/DashboardSlotViz.vue:66]: Not really, that would be if `...length > 0`, but here I don't want to have any selected `group id` if we only have 1 group containing validators - this is to keep the `GET request` cleaner, as when having only one group we don't need to pass `group ids` and just request all validators.

File: frontend/components/dashboard/slot-viz/DashboardSlotViz.vue (hunk 11 above)
Comment (enzo-bitfly) [reply frontend/components/dashboard/slot-viz/DashboardSlotViz.vue]: Yeah it's ugly. I removed it and used the values directly in `DashboardSlotVizDutyVisibilityToggle`

File: frontend/components/dashboard/slot-viz/DashboardSlotViz.vue (hunk 12 above)
Comment (enzo-bitfly) [reply frontend/components/dashboard/slot-viz/DashboardSlotViz.vue:121]: I prefer to keep it like this. The `DashboardSlotVizGroupSelector` only cares about getting `validator-groups`, but in the context of the parent (`ValidatorSlotViz`), it's not necessarily passing all `validator-groups`, but those that contain validators (groups can also be empty).

File: frontend/components/dashboard/slot-viz/DashboardSlotViz.vue (hunk 13 above)
Comment (enzo-bitfly) [reply frontend/components/dashboard/slot-viz/DashboardSlotViz.vue]: This brings up a good point: I was setting `loading = true` in the `validator_overview_store` too early, and it remained true even if the fetch wasn't performed by lack of `dashboardKey` in the case of guest dashboards. I can now remove the `!!dashboardKey` at least.

File: frontend/components/dashboard/slot-viz/DashboardSlotVizDutyVisibilityToggle.vue (hunk 14 above)
Comment (enzo-bitfly) [reply frontend/components/dashboard/slot-viz/DashboardSlotVizDutyVisibilityToggle.vue:144]: Done

File: frontend/composables/useValidatorDashboardGroups.ts
Comment (enzo-bitfly) [reply frontend/composables/useValidatorDashboardGroups.ts:1]: That would open a new rabbit hole and it's out of scope. As you see, the composable adds a beautiful `translation key`, which I don't want to have in the store, for the `name` in case there are no groups. If I refactor it, it would take me some work to check for a correct behaviour in all it's 15 uses.

File: frontend/pages/playground.vue (hunk 7 above)
Comment (enzo-bitfly) [reply frontend/pages/playground.vue]: Done

File: frontend/stores/dashboard/useValidatorDashboardOverviewStore.ts (hunk 15 above)
Comment (enzo-bitfly) [reply frontend/stores/dashboard/useValidatorDashboardOverviewStore.ts]: I added a comment.

File: frontend/stores/dashboard/useValidatorDashboardOverviewStore.ts (hunk 2 above)
Comment (enzo-bitfly) [reply frontend/stores/dashboard/useValidatorDashboardOverviewStore.ts]: Renamed to `validator-dashboard-overview`

File: frontend/stores/dashboard/useValidatorDashboardOverviewStore.ts (hunk 3 above)
Comment (enzo-bitfly) [reply frontend/stores/dashboard/useValidatorDashboardOverviewStore.ts]: Right, I couldn't find any case where it can be `null`, so I removed it.

File: frontend/stores/dashboard/useValidatorDashboardOverviewStore.ts (hunk 4 above)
Comment (enzo-bitfly) [reply frontend/stores/dashboard/useValidatorDashboardOverviewStore.ts]: The problem is that the fetch happens in the parent (`pages/dashboard/index.vue`) - in order for `DashboardSlotViz` to get that value, I could pass it as a prop, but then it gets uglier, as it only is `loading-overview`, and internally the component is also be fetching `slot-viz`. 

I think we need to redefine how fetches are made and not have to wrap them with `useAsyncData` on every use we need it - gut feeling because I haven't gotten into those details.

File: frontend/stores/dashboard/useValidatorDashboardOverviewStore.ts (hunk 16 above)
Comment (enzo-bitfly) [reply frontend/stores/dashboard/useValidatorDashboardOverviewStore.ts]: Done

**Review by enzo-bitfly — COMMENTED** [review]
Great review, thanks 👍

File: frontend/stores/dashboard/useValidatorDashboardOverviewStore.ts [hunk 17]
```diff
@@ -84,11 +80,23 @@ export function useValidatorDashboardOverviewStore() {
     weekly: (overview.value?.chart_history_seconds?.weekly ?? 0) > 0,
   }))
 
+  const isLargeDashboard = computed(() => {
+    if (!validatorCount.value) return false
+
+    // This value is a product decision from Bitfly, and doesn't stem from any mechanism in the blockchains we handle.
```
marcel-bitfly:
```suggestion
    // this amount is a product decision
```

frontend/components/dashboard/slot-viz/DashboardSlotViz.vue
marcel-bitfly: true 😅, but confusing

marcel-bitfly (approved):
thx

PR #1309, opened 2025-02-05
refactor(dashboard): added unit tests for dashboard helpers

sasha-bitfly:
Before we cover functions like `GetValidatorDashboardSummary` or `GetValidatorDashboardGroupSummary` with unit tests we need to cleanup and refactor a codebase.

Helper functions to be moved to` vdb_helpers.go` to avoid mess in `vdb_dashboard.go`

Every function is split into build query, execute query, process results

Generic `executeQuery` function is added

### DIFF
--- a/backend/pkg/api/data_access/general.go
+++ b/backend/pkg/api/data_access/general.go
@@ -9,6 +9,7 @@ import (
 	"github.com/ethereum/go-ethereum/common/hexutil"
 	"github.com/gobitfly/beaconchain/pkg/api/types"
 	"github.com/gobitfly/beaconchain/pkg/commons/db"
+	"github.com/jmoiron/sqlx"
 )
 
 // retrieve (primary) ens name and optional name (=label) maintained by beaconcha.in, if present
@@ -130,4 +131,45 @@ func applySortAndPagination(defaultColumns []types.SortColumn, primary types.Sor
 	}
 
 	return queryOrder, queryWhere, nil
+}
+
+// Generic function to execute a query
+// Ensures T is a slice in compile-time
+// Retrieves multiple rows and stores them in a slice
+// Used when the query returns multiple results (e.g., SELECT * FROM users)
+// The destination must be a slice ([]T)
+func runQueryRows[T ~[]E, E any](ctx context.Context, db *sqlx.DB, ds *goqu.SelectDataset) (T, error) {
+	query, args, err := ds.Prepared(true).ToSQL()
+	if err != nil {
+		var zero T
+		return zero, fmt.Errorf("error preparing query: %w", err)
+	}
+
+	var result T // slice of type T
+	err = db.SelectContext(ctx, &result, query, args...)
+	if err != nil {
+		return result, fmt.Errorf("error executing query: %w", err)
+	}
+
+	return result, nil
+}
+
+// Generic function to execute a query
+// Retrieves a single row and stores it in a struct or variable
+// Used when the query is expected to return only one row (e.g., SELECT * FROM users WHERE id = ?)
+// The destination must be a single struct or variable (T)
+func runQuery[T any](ctx context.Context, db *sqlx.DB, ds *goqu.SelectDataset) (T, error) {
+	query, args, err := ds.Prepared(true).ToSQL()
+	if err != nil {
+		var zero T
+		return zero, fmt.Errorf("error preparing query: %w", err)
+	}
+
+	var result T
+	err = db.GetContext(ctx, &result, query, args...)
+	if err != nil {
+		return result, fmt.Errorf("error executing query: %w", err)
+	}
+
+	return result, nil
 }
--- a/backend/pkg/api/data_access/vdb_helpers.go
+++ b/backend/pkg/api/data_access/vdb_helpers.go
@@ -400,3 +400,217 @@ func (d *DataAccessService) calculateValidatorDashboardBalance(ctx context.Conte
 	}
 	return balances, nil
 }
+
+// Builds a query to select the maximum timestamp from the validator_dashboard_data view for an interval (epoch, hourly, daily, weekly)
+func buildLatestExportedChartQuery(view, dateColumn string) *goqu.SelectDataset {
+	return goqu.Dialect("postgres").From(view).Select(goqu.MAX(dateColumn))
+}
+
+// Converts a time.Time value into a Unix timestamp (uint64).
+func processLastExportedChartResult(ts time.Time) (uint64, error) {
+	return uint64(ts.Unix()), nil
+}
+
+// Retrieves the latest exported chart timestamp for a given aggregation level
+// 1. Determines the appropriate view and date column based on the aggregation type
+// 2. Constructs and executes the query to fetch the latest timestamp
+// 3. Converts the result into a Unix timestamp and returns it
+func (d *DataAccessService) GetLatestExportedChartTs(ctx context.Context, aggregation enums.ChartAggregation) (uint64, error) {
+	view, dateColumn, err := getViewAndDateColumn(aggregation)
+	if err != nil {
+		return 0, err
+	}
+
+	ds := buildLatestExportedChartQuery(view, dateColumn)
+
+	ts, err := runQuery[time.Time](ctx, d.clickhouseReader, ds)
+	if err != nil {
+		return 0, err
+	}
+
+	return processLastExportedChartResult(ts)
+}
+
+type SyncCommitteeResult struct {
+	ValidatorIndex uint64 `db:"validatorindex"`
+	Period         uint64 `db:"period"`
+}
+
+// Pass the latest (most recently finalized epoch)
+// to determine validators current and upcoming sync committees
+func (d *DataAccessService) getCurrentAndUpcomingSyncCommittees(ctx context.Context, latestEpoch uint64) (map[uint64]bool, map[uint64]bool, error) {
+	currentSyncPeriod := utils.SyncPeriodOfEpoch(latestEpoch)
+	ds := buildSyncCommitteeQuery(currentSyncPeriod)
+	queryResult, err := runQueryRows[[]SyncCommitteeResult](ctx, d.readerDb, ds)
+	if err != nil {
+		return nil, nil, err
+	}
+	return processSyncCommitteeResults(queryResult, utils.SyncPeriodOfEpoch(latestEpoch))
+}
+
+// Builds a query to fetch sync committee validators for the current and next period
+func buildSyncCommitteeQuery(currentSyncPeriod uint64) *goqu.SelectDataset {
+	return goqu.Dialect("postgres").
+		Select(
+			goqu.L("validatorindex"),
+			goqu.L("period"),
+		).
+		From("sync_committees").
+		Where(goqu.L("period IN (?, ?)", currentSyncPeriod, currentSyncPeriod+1))
+}
+
+// Processes query results and categorizes validators into current and upcoming sync committees
+func processSyncCommitteeResults(queryResult []SyncCommitteeResult, currentSyncPeriod uint64) (map[uint64]bool, map[uint64]bool, error) {
+	currentSyncCommitteeValidators := make(map[uint64]bool)
+	upcomingSyncCommitteeValidators := make(map[uint64]bool)
+
+	for _, entry := range queryResult {
+		if entry.Period == currentSyncPeriod {
+			currentSyncCommitteeValidators[entry.ValidatorIndex] = true
+		} else {
+			upcomingSyncCommitteeValidators[entry.ValidatorIndex] = true
+		}
+	}
+
+	return currentSyncCommitteeValidators, upcomingSyncCommitteeValidators, nil
+}
+
+// Retrieves the start epoch for a given time period (last 1h, 24h, 7d, 30d)
+func (d *DataAccessService) getEpochStart(ctx context.Context, period enums.TimePeriod) (uint64, error) {
+	clickhouseTable, _, err := getTablesForPeriod(period)
+	if err != nil {
+		return 0, err
+	}
+
+	ds := buildEpochStartQuery(clickhouseTable)
+	epochStart, err := runQuery[uint64](ctx, d.clickhouseReader, ds)
+	if err != nil {
+		return 0, err
+	}
+
+	return epochStart, nil
+}
+
+// Builds a query to retrieve the earliest epoch start value
+// Orders results in ascending order and limits to the first row.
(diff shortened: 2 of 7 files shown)


backend/pkg/api/data_access/vdb_helpers.go
```diff
@@ -400,3 +402,220 @@ func (d *DataAccessService) calculateValidatorDashboardBalance(ctx context.Conte
...
+	}
+
+	var result T
+	resultType := reflect.TypeOf(result)
+
+	if resultType != nil && resultType.Kind() == reflect.Slice {
+		err = db.SelectContext(ctx, &result, query, args...)
+	} else {
+		err = db.GetContext(ctx, &result, query, args...)
+	}
```
LuccaBitfly: Issue: I'd do 2 separate function for select and get rather than doing a runtime type check for every query, since reflecting does have a performance impact.

LuccaBitfly: Also we're gonna need something similar for `ExecContext`, right? `executeQuery` as a func name might be a bit misleading then.

backend/pkg/api/data_access/vdb_helpers.go
```diff
@@ -400,3 +402,220 @@ func (d *DataAccessService) calculateValidatorDashboardBalance(ctx context.Conte
...
 	return balances, nil
 }
+
+func (d *DataAccessService) GetLatestExportedChartTs(ctx context.Context, aggregation enums.ChartAggregation) (uint64, error) {
+	view, dateColumn, err := d.getViewAndDateColumn(aggregation)
+	if err != nil {
+		return 0, err
+	}
+
+	query := fmt.Sprintf(`SELECT max(%s) FROM %s`, dateColumn, view)
```
LuccaBitfly: Question: shouldn't we use goqu for all queries?

backend/pkg/api/data_access/vdb_helpers.go
```diff
@@ -400,3 +402,220 @@ func (d *DataAccessService) calculateValidatorDashboardBalance(ctx context.Conte
...
+// --- Past Sync Committees Functions ---
+func (d *DataAccessService) getPastSyncCommittees(ctx context.Context, indices []uint64, epochStart uint64, latestEpoch uint64) (map[uint64]uint64, error) {
+	ds := buildPastSyncCommitteesQuery(indices, epochStart, latestEpoch)
+	validatorIndices, err := executeQuery[[]uint64](ctx, d.alloyReader, ds)
+	if err != nil {
+		return nil, err
+	}
+
+	validatorCountMap := processPastSyncCommitteesResults(validatorIndices)
+	return validatorCountMap, nil
```
LuccaBitfly: Nitpick (non-blocking):
```suggestion
	return processPastSyncCommitteesResults(validatorIndices)
, nil
```

backend/pkg/api/data_access/vdb_helpers.go
```diff
@@ -400,3 +402,220 @@ func (d *DataAccessService) calculateValidatorDashboardBalance(ctx context.Conte
...
+}
+
+func buildPastSyncCommitteesQuery(indices []uint64, epochStart, latestEpoch uint64) *goqu.SelectDataset {
+	pastSyncPeriodCutoff := utils.SyncPeriodOfEpoch(epochStart)
+	currentSyncPeriod := utils.SyncPeriodOfEpoch(latestEpoch)
```
LuccaBitfly: Question: You have a bit of logic in your build query func? Do we want this? Or should we maybe pass this to the builder func?

backend/pkg/api/data_access/vdb_helpers_test.go
```diff
@@ -0,0 +1,568 @@
...
+
+	cfg := types.Config{
+		Chain: types.Chain{
+			ClConfig: types.ClChainConfig{
+				EpochsPerSyncCommitteePeriod: 32,
+				AltairForkEpoch:              0,
+			},
+		},
+	}
+	utils.Config = &cfg
```
LuccaBitfly, line 29: Issue: We should not rely on global states in our tests. The data access service struct already has a specific field for its config and should ONLY read from that. We should get rid of global state reliance ASAP

backend/pkg/api/data_access/vdb_helpers_test.go
```diff
@@ -0,0 +1,568 @@
...
+
+	sqlxDB := sqlx.NewDb(mockDB, "sqlmock")
+	dataService := &DataAccessService{
+		alloyReader: sqlxDB,
+	}
```
LuccaBitfly: Issue: All this test setup should be in it's own function, otherwise we'll end up with a lot of repeated code across our tests. Keep in mind that the `DataAccessService` struct already has a `Close()` function implemented, which closes all of it's databases in a nil safe manner. I imagine that, ideally, we'd end up with a test setup like 
```go
func dataAccesesTestSetup(/* whatever params you need */) (service *DataAccessService, closeFunc func()) {
     // setup everything
     return dataAccessService, dataAccessService.Close
}
```
```go
// inside the actual test func
dataAccessService, close := dataAccesesTestSetup(/* whatever params you need */)
defer close() // calls close of the data access service itself
```

backend/pkg/api/data_access/vdb_helpers_test.go
```diff
@@ -0,0 +1,568 @@
...
+
+	tests := []struct {
+		name          string
+		period        enums.TimePeriod
+		mockRows      *sqlmock.Rows
+		mockError     error
+		expectedVal   uint64
+		expectedError string
+	}{
```
LuccaBitfly, line 270: Thought (non-blocking): Since all of these mockTests use alot of similar fields, I'm wondering if we could maybe generalize some of this. E.g. a lot of the lines for a single test case iteration are repeated multiple times (setting the mock error, checking for expectedError, asserting exectedVal etc.)

LuccaBitfly (changes requested):
This is moving in a very nice direction, nice work! Still have a few comments, pls have a look.

backend/pkg/api/data_access/vdb_helpers.go
sasha-bitfly: Done

backend/pkg/api/data_access/vdb_helpers.go
sasha-bitfly: Done

backend/pkg/api/data_access/vdb_helpers.go
sasha-bitfly: Done

backend/pkg/api/data_access/vdb_helpers.go
sasha-bitfly: Done

backend/pkg/api/data_access/vdb_helpers_test.go
sasha-bitfly: Unable to hide utils.Config, utils are used in API function, this needs to be address as a separate PR

backend/pkg/api/data_access/vdb_helpers_test.go
sasha-bitfly: Done partially for those tests that can be generalised

backend/pkg/api/data_access/vdb_helpers_test.go
sasha-bitfly: It is hard to encapsulate, depending on test we pass 
```
dataService := &DataAccessService{
		clickhouseReader: sqlxDB,
	}
```

```
dataService := &DataAccessService{
		alloyReader: sqlxDB,
	}
```

```
dataService := &DataAccessService{
		readerDb: sqlxDB,
	}
```

LuccaBitfly: Would passing all of them work?
```go
	dataService := &DataAccessService{
		readerDb: sqlxDB,
		alloyReader: sqlxDB,
		clickhouseReader: sqlxDB,
	}
```

sasha-bitfly: Done

backend/pkg/api/data_access/vdb_helpers.go
```diff
@@ -400,3 +401,233 @@ func (d *DataAccessService) calculateValidatorDashboardBalance(ctx context.Conte
...
+
+func processPastSyncCommitteesResults(validatorIndices []uint64) (map[uint64]uint64, error) {
+	validatorCountMap := make(map[uint64]uint64)
+	for _, validatorIndex := range validatorIndices {
+		validatorCountMap[validatorIndex]++
+	}
+	return validatorCountMap, nil
+}
+
+func (d *DataAccessService) getTablesForPeriod(period enums.TimePeriod) (string, int, error) {
```
LuccaBitfly: Issue (non-blocking): IMO funcs that don't access the service shouldn't have it as an method receiver. You see how this causes you to unnecessarily create data access vars in their respective test methods.

backend/pkg/api/data_access/vdb_helpers_test.go
```diff
@@ -0,0 +1,388 @@
...
+	mockDB, mock, err := sqlmock.New()
+	assert.NoError(t, err)
+	sqlxDB := sqlx.NewDb(mockDB, "sqlmock")
+	dataAccess := &DataAccessService{
+		clickhouseReader: sqlxDB,
+		readerDb:         sqlxDB,
+		alloyReader:      sqlxDB,
+	}
+
+	return dataAccess, dataAccess.Close, mock
```
LuccaBitfly: TODO (non-blocking): should write unit tests for `dataAccess.Close`, can be done in another PR tho

LuccaBitfly (approved):
lgtm!

backend/pkg/api/data_access/vdb_helpers.go
```diff
@@ -400,3 +401,233 @@ func (d *DataAccessService) calculateValidatorDashboardBalance(ctx context.Conte
 	}
 	return balances, nil
 }
+
+// --- Generic Query Execution ---
```
mccreedy-bitfly: Should these be in a different file since they dont seem specific to vdb? Perhaps a general query helper?

backend/pkg/api/data_access/vdb_helpers.go
```diff
@@ -400,3 +401,233 @@ func (d *DataAccessService) calculateValidatorDashboardBalance(ctx context.Conte
...
+}
+
+func buildPastSyncCommitteesQuery(indices []uint64, pastSyncPeriodCutoff, currentSyncPeriod uint64) *goqu.SelectDataset {
+	return goqu.Dialect("postgres").
+		Select(goqu.L("sc.validatorindex")).
+		From(goqu.L("sync_committees sc")).
+		Where(goqu.L("period >= ? AND period < ? AND validatorindex = ANY(?)", pastSyncPeriodCutoff, currentSyncPeriod, pq.Array(indices)))
+}
+
+func processPastSyncCommitteesResults(validatorIndices []uint64) (map[uint64]uint64, error) {
```
mccreedy-bitfly, line 531: Name isnt descriptive and no comments describing what its doing

backend/pkg/api/data_access/vdb_helpers.go
```diff
@@ -400,3 +401,233 @@ func (d *DataAccessService) calculateValidatorDashboardBalance(ctx context.Conte
...
+	}
+
+	ds := buildEpochStartQuery(clickhouseTable)
+	epochStart, err := getContext[uint64](ctx, d.clickhouseReader, ds)
+	if err != nil {
+		return 0, err
+	}
+
+	return epochStart, nil
+}
```
mccreedy-bitfly: Don't we have this defined somewhere already as a general purpose query?

backend/pkg/api/data_access/vdb_helpers.go
```diff
@@ -400,3 +401,233 @@ func (d *DataAccessService) calculateValidatorDashboardBalance(ctx context.Conte
...
+	return goqu.Dialect("postgres").
+		Select(
+			goqu.L("validatorindex"),
+			goqu.L("period"),
+		).
+		From("sync_committees").
+		Where(goqu.L("period IN (?, ?)", currentSyncPeriod, currentSyncPeriod+1))
+}
+
+func processSyncCommitteeResults(queryResult []SyncCommitteeResult, currentSyncPeriod uint64) (map[uint64]bool, map[uint64]bool, error) {
```
mccreedy-bitfly, line 463: Hard to tell what this is doing and no comments

mccreedy-bitfly (commented):
Generally think we need to add a lot more comments, but think that this is moving things in the right direction!

backend/pkg/api/data_access/vdb_helpers.go
sasha-bitfly: We don't have general function for getting epoch start.
We need to make one during refactoring and include more fields: e.g. `epoch_start`, `epoch_end`, etc

backend/pkg/api/data_access/vdb_helpers.go
sasha-bitfly: Moved selectContext and getContext to general.go and added unit tests

backend/pkg/api/data_access/vdb_helpers.go
sasha-bitfly: Added comments

backend/pkg/api/data_access/general.go
```diff
@@ -131,3 +132,43 @@ func applySortAndPagination(defaultColumns []types.SortColumn, primary types.Sor
...
+	}
+
+	return result, nil
+}
+
+// Generic function to execute a query
+// Retrieves a single row and stores it in a struct or variable
+// Used when the query is expected to return only one row (e.g., SELECT * FROM users WHERE id = ?)
+// The destination must be a single struct or variable (T)
+func getContext[T any](ctx context.Context, db *sqlx.DB, ds *goqu.SelectDataset) (T, error) {
```
mccreedy-bitfly: Not a big fan of the name of this function, since it overloads the context field that is used in many other places. Why not abstract out the concept of a "Context" from this function by not naming it as such (the caller of this function doesn't need to know that the underlying implementation calls `db.GetContext()`). Same comment for the function above (`selectContext`)

mccreedy-bitfly (approved):

backend/pkg/api/data_access/vdb_helpers.go
sasha-bitfly: Done

backend/pkg/api/data_access/general.go
sasha-bitfly: Done

PR #1312, opened 2025-02-05
chore(ci): add cache

Tangui-Bitfly:
* specify go.sum path since it's not at the root
* make dummy change (remove irrelevant comments) to trigger CI


LuccaBitfly (approved):
lgtm!!

PR #1305, opened 2025-01-30
Beds 97/fix gnosis currency issues

marcel-bitfly:

PR #1314, opened 2025-02-06
(BEDS-1184) return el rewards for guest dashboards

remoterami:


backend/pkg/api/data_access/vdb_rewards.go
```diff
@@ -678,7 +677,7 @@ func (d *DataAccessService) GetValidatorDashboardGroupRewards(ctx context.Contex
 		ret.ProposalClSlashingIncReward = ret.ProposalClSlashingIncReward.Add(entry.SlasherRewards.Mul(gWei))
 	}
 
-	ret.Proposal.Income = ret.Proposal.Income.Add(elRewards)
+	ret.Proposal.Income = ret.Proposal.Income.Add(d.convertElToMain(elRewards).Mul(decimal.NewFromInt(d.config.Frontend.ElCurrencyDivisor)))
```
LuccaBitfly: Issue: You're converting the EL currency to the main currency and adding it to a field that contains values in the CL currency. On Gnosis, this would mean that you've converted xDai to GNO and then added it to a mGNO value. This field should be a `ClElValue` struct.

LuccaBitfly (changes requested):

backend/pkg/api/data_access/vdb_rewards.go
remoterami: Applied your suggestion. Could also just make use of the already existing `ret.ProposalElReward`, but keeping it explicit to reduce any future source of error also shouldn't hurt

Also if it's fine from the API side, would you please fix the frontend data type issue? :sweat_smile:

LuccaBitfly:
merged as part of #1305

PR #1315, opened 2025-02-07
refactor(balance): refactor balance indexing 

Tangui-Bitfly:
* extract logic for balance update
* create reusable batcher 
* use [multicall](https://github.com/mds1/multicall) batcher by default
* remove code from db.Bigtable and rpc clients
* add tests


backend/pkg/commons/db2/database/bigtable.go
```diff
@@ -180,9 +180,17 @@ func (b BigTable) BulkAdd(table string, itemsByKey map[string][]Item, opts ...Op
 		keys = append(keys, key)
 		muts = append(muts, mut)
 	}
+	return b.bulk(table, keys, muts, options)
+}
+
+func (b BigTable) bulk(table string, keys []string, mutations []*bigtable.Mutation, options options) error {
```
Monika-Bitfly: Would it be better to use a more descriptive name for this function eg parseBulk or applyBulk or something similair ?

backend/pkg/commons/db2/metadataupdates/store.go
```diff
@@ -1,14 +1,20 @@
...
 	"strings"
 
+	"github.com/ethereum/go-ethereum/common"
+
 	"github.com/gobitfly/beaconchain/pkg/commons/db2/data"
 	"github.com/gobitfly/beaconchain/pkg/commons/db2/database"
 	"github.com/gobitfly/beaconchain/pkg/commons/types"
 )
 
+const balanceKey = "B"
```
Monika-Bitfly: Can we store this value inside keys.go?

Monika-Bitfly (approved):

backend/cmd/eth1indexer/main.go
```diff
@@ -545,55 +555,15 @@ func HandleChainReorgs(bt *db.Bigtable, client *rpc.ErigonClient, depth int) err
...
 }
 
-func ProcessMetadataUpdates(bt *db.Bigtable, client *rpc.ErigonClient, prefix string, batchSize int, iterations int) {
-	lastKey := prefix
-
-	its := 0
-	for {
+func ProcessBalanceUpdates(balanceUpdater executionlayer.BalanceUpdater, batchSize int, iterations int) {
+	for its := 0; iterations == -1 || its < iterations; its++ {
 		start := time.Now()
```
invis-bitfly, line 569: explain magic number `-1` in doc comment

backend/cmd/eth1indexer/main.go
```diff
@@ -179,26 +179,36 @@ func Run() {
...
 	bigtable, err := database.NewBigTable(utils.Config.Bigtable.Project, utils.Config.Bigtable.Instance, nil)
 	if err != nil {
 		log.Fatal(err, "error connecting to bigtable", 0)
 	}
 
+	cache := freecache.NewCache(100 * 1024 * 1024) // 100 MB limit
+	metadataUpdatesStore := metadataupdates.NewStore(database.Wrap(bigtable, metadataupdates.Table), cache)
+	dataStore := data.NewStore(database.Wrap(bigtable, data.Table))
+	metadataStore := metadata.NewStore(database.Wrap(bigtable, metadata.Table))
+
```
invis-bitfly, line 191: if the Stores are different enough to require different `New*` functions, can we name them destinctly? Having `newStore` in 3 different packages all return a struct called `Store` with no shared Interface feels very confusing.

backend/pkg/commons/db2/database/option.go
```diff
@@ -1,7 +1,7 @@
 package database
```
invis-bitfly, line 1: i cant comment directly because its out-of-bounds for this review but `func apply(opts []Option) options {` is a bad name for what the function does - it doesn't apply anything, it returns a new options object based on a given list. `apply` would be more reasonable if i could call it on a `options` struct to update its values. I think `newOptions` or something would be more fitting?

backend/pkg/commons/db2/database/option.go
```diff
@@ -1,7 +1,7 @@
...
```
invis-bitfly, line 2:
```golang
type Option interface {
	 apply(*options)
}

type rowKeyFilterOption string

func (r rowKeyFilterOption) apply(opts *options) {
	 opts.RowKeyFilter = string(r)
}

func WithRowKeyFilter(regex string) Option {
	 return rowKeyFilterOption(regex)
}
```
could be simplified to
```golang
type Option func(opts *options)

func WithRowKeyFilter(regex string) Option {
	 return func (opts *options) {
	 	 opts.RowKeyFilter = regex
	 }
}
```
which would mean we could remove all the one-use custom types defined here

backend/pkg/commons/db2/metadataupdates/keys.go
```diff
@@ -67,8 +69,8 @@ func ContractUpdate(blockNumber uint64, chainID string, updates []ContractUpdate
 func MarkBalanceUpdate(chainID string, address []byte, token []byte, cache Cache) map[string][]database.Item {
 	items := make(map[string][]database.Item)
 
-	key := fmt.Sprintf("%s:B:%x", chainID, address) // format is B: for balance update as chainid:prefix:address (token id will be encoded as column name)
-	keyCache := []byte(fmt.Sprintf("%s:B:%x:%x", chainID, address, token))
+	key := fmt.Sprintf("%s:%s:%x", chainID, balanceKey, address) // format is B: for balance update as chainid:prefix:address (token id will be encoded as column name)
+	keyCache := []byte(fmt.Sprintf("%s:%s:%x:%x", chainID, balanceKey, address, token))
```
invis-bitfly:
```suggestion
	keyCache := []byte(fmt.Sprintf("%s:%x", key, token))
```

backend/pkg/commons/db2/metadataupdates/store.go
```diff
@@ -96,3 +100,43 @@ func mergeItems(dest map[string][]database.Item, source map[string][]database.It
...
+	if err != nil {
+		return nil, err
+	}
+	var pairs []Pair
+	for _, row := range rows {
+		for col := range row.Values {
+			pairs = append(pairs, Pair{
+				Address: common.HexToAddress(strings.Split(row.Key, ":")[2]),
+				Token:   common.HexToAddress(strings.Split(col, ":")[1]),
+			})
```
invis-bitfly: split out so it can be tested?

backend/pkg/commons/db2/metadata/metadata.go
```diff
@@ -0,0 +1,36 @@
+package metadata
```
invis-bitfly, line 1: this seems balance update scentric yet the package is called metadata?

backend/pkg/executionlayer/evm/batch.go
```diff
@@ -0,0 +1,219 @@
...
+	chain.IDs.Holesky.String():    common.HexToAddress("0xcA11bde05977b3631167028862bE2a173976CA11"),
+}
+
+// Batcher only supports eth_getBalance and eth_call for now
+// feel free to add more supported methods when necessary
+type Batcher interface {
+	Batch(elements []BatchElement) ([]BatchResponse, error)
+}
+
+func NewBatcher(chainID string, client *ethclient.Client) Batcher {
```
invis-bitfly: number typed as string 🫠

backend/pkg/executionlayer/evm/batch.go
```diff
@@ -0,0 +1,219 @@
...
+	multicall, exist := multicallerFor[chainID]
```
invis-bitfly: allow overwriting using config?

backend/pkg/executionlayer/evm/batch.go
```diff
@@ -0,0 +1,219 @@
...
+
+type BatchResponse struct {
+	Data   []byte
+	Failed bool
+}
+
+// on mainnet it closer to 7_000 for balance retrieval
+// 5_000 is good compromise between perf and potential higher gas consuming calls
+// if this start to impact performances, we will need to have a limit per network (different block gas limits)
+// and potentially per type of call
```
invis-bitfly, line 72: make configurable over config? so it can be changed if something breaks without having to redeploy.

backend/pkg/executionlayer/evm/batch.go
```diff
@@ -0,0 +1,219 @@
...
+		partial, err := m.batch(elements[i:end])
+		if err != nil {
+			return nil, err
+		}
+		res = append(res, partial...)
+	}
+	return res, nil
+}
+
+func (m MulticallBatcher) batch(elements []BatchElement) ([]BatchResponse, error) {
```
invis-bitfly: i kinda dislike the "have internal function with identical name, args and returns but different behavior" pattern?

backend/pkg/executionlayer/evm/batch.go
```diff
@@ -0,0 +1,219 @@
...
+		multicall:    multicall,
+		multicallABI: multicallABI,
+		address:      multicallAddress,
+		limit:        multicallDefaultLimit,
+	}
+}
+
+func (m MulticallBatcher) Batch(elements []BatchElement) ([]BatchResponse, error) {
+	var res []BatchResponse
+	for i := 0; i < len(elements); i = i + m.limit {
```
invis-bitfly: if ur already doing `i = i + m.limit` instead of `i += m.limit` why not go straight to `i = min(i + m.limit, len(elements))`?

backend/pkg/executionlayer/evm/batch.go
```diff
@@ -0,0 +1,219 @@
...
+func (m MulticallBatcher) batch(elements []BatchElement) ([]BatchResponse, error) {
+	var batch []contracts.IMulticall3Call3
+	var res []BatchResponse
+
+	for _, element := range elements {
+		switch element.Method {
+		case "eth_getBalance":
+			// in case of eth_getBalance we redirect the call to multicall.getEthBalance
+			element.To = &m.address
+			element.Data, _ = m.multicallABI.Pack("getEthBalance", common.BytesToAddress(element.Data))
```
invis-bitfly: prob could be split into a sub function with correct error handling

backend/pkg/executionlayer/evm/batch.go
```diff
@@ -0,0 +1,219 @@
...
+	for _, elem := range batch {
+		res = append(res, BatchResponse{
+			Data:   common.FromHex(*elem.Result.(*string)),
+			Failed: elem.Error != nil,
+		})
+	}
+	return res, nil
+}
+
+func Balance(batcher Batcher, pairs []metadataupdates.Pair) ([]*big.Int, error) {
```
invis-bitfly: bad function name. `GetBalancesForPairs` or smth

backend/pkg/executionlayer/evm/batch.go
```diff
@@ -0,0 +1,219 @@
...
+
+type RPCBatch struct {
+	client *rpc.Client
+}
+
+func (m RPCBatch) Batch(elements []BatchElement) ([]BatchResponse, error) {
+	var batch []rpc.BatchElem
+	var res []BatchResponse
+
+	for _, element := range elements {
```
invis-bitfly, line 169: seems error prone that `RPCBatch` does no automatic chunking but `MulticallBatcher` does. If i write my code under the assumption that `MulticallBatcher` does the splitting but then on a new chain `NewBatcher` falls over to `RPCBatch` it would potentially break-

backend/pkg/executionlayer/evm/batch.go
```diff
@@ -0,0 +1,219 @@
...
+// Batcher only supports eth_getBalance and eth_call for now
+// feel free to add more supported methods when necessary
+type Batcher interface {
+	Batch(elements []BatchElement) ([]BatchResponse, error)
+}
+
+func NewBatcher(chainID string, client *ethclient.Client) Batcher {
+	multicall, exist := multicallerFor[chainID]
+	if !exist {
+		return RPCBatch{client: client.Client()}
```
invis-bitfly: maybe log a warn or smth if this happens? we should know when an optimization doesnt trigger for some reason. (deploy against new chain => missing entry for multicall address => we wonder why perf is worse than expected)

backend/pkg/executionlayer/evm/batch_test.go
```diff
@@ -0,0 +1,204 @@
...
+		t.Run(tt.name, func(t *testing.T) {
+			poller := NewBatcher(tt.chainID, &ethclient.Client{})
+			if got, want := reflect.TypeOf(poller), reflect.TypeOf(tt.expected); got != want {
+				t.Errorf("got %v, want %v", got, want)
+			}
+		})
+	}
+}
+
+func TestBatcher(t *testing.T) {
```
invis-bitfly, line 54: this tests `Balance` not `Batcher`? If `Balance` has a change, this might trip, and on the other side, if `Batcher` has a change, this potentially might not trip.

backend/pkg/executionlayer/evm/batch_test.go
```diff
@@ -0,0 +1,204 @@
...
+		t.Errorf("got %v, want %v", got, want)
+	}
+	if got, want := catchClient.callCount, 2; got != want {
+		t.Errorf("got %v, want %v", got, want)
+	}
+}
+
+// TestBatcherRealCondition is useful to compare performances between the different Batcher
+// it tests over the same set of randomly generated addresses for eth and erc20 balances
+func TestBatcherRealCondition(t *testing.T) {
```
invis-bitfly: benchmark not test?

invis-bitfly (changes requested):

invis-bitfly (changes requested):
[ ](url)

backend/cmd/eth1indexer/main.go
Tangui-Bitfly: That something I plan to do later on. First I wanted to migrate each functionality individually to understand better the link between the different tables.
If that alright, I will do your suggestion in an other pr.

backend/pkg/commons/db2/metadata/metadata.go
Tangui-Bitfly: That's the name of table, yes that's not great but in my mind it's a temp name.

backend/pkg/executionlayer/evm/batch.go
Tangui-Bitfly: ChainID you mean ? It's not really use as a number, more like an id.

backend/pkg/executionlayer/evm/batch.go
Tangui-Bitfly: If the caller what to use a specific multicall address it can call `NewMulticallBatcher`.
I prefer to keep signature of what will be the most used constructor function the simplest possible.

backend/pkg/executionlayer/evm/batch.go
Tangui-Bitfly: I'm a partisan of not exposing too much configuration params when possible. Each params exposed is a complexity that leak to higher layers and to the config file.
For the current usage, I don't see us having an issue with this hardcoded limit.
I just want to write code for a practical reason not for an hypothetical future than never come.

backend/pkg/executionlayer/evm/batch.go
Tangui-Bitfly: I'm not sure of what will `i = min(i + m.limit, len(elements))` bring.
The min between the len and the limit is relative to the end value we will process in the current loop, I don't see how it's linked to i.

backend/pkg/executionlayer/evm/batch.go
Tangui-Bitfly: This will never error.
The only error that can happen here are when coding, for example if I don't provide an address but a number. Since it's covered, we know that it is working.

backend/pkg/executionlayer/evm/batch.go
Tangui-Bitfly: They both not using the same "protocol".
`RPCBatch` relies on the rpc client, so the limitation are inherited from http. Those limitation are probably handled but the `rpc.Client` and if not by the `http.Client`. I don't believe we have much to do here.
`MulticallBatcher` relies on `eth_call` which execute a read only function on the blockchain node. This call is still limited by gas consumed by the read only call (prevent node to be ddos with infinite loop).
That's why `MulticallBatcher` requires an extra protection compare to `RPCBatch`.

backend/pkg/executionlayer/evm/batch_test.go
Tangui-Bitfly: this test only one thing for now but will grow in time.
If you run this test you will get:
```
=== RUN   TestBatcher
=== RUN   TestBatcher/rpc_batch
=== RUN   TestBatcher/rpc_batch/Balance
=== RUN   TestBatcher/rpc_batch/Balance/erc20
--- PASS: TestBatcher/rpc_batch/Balance/erc20 (0.00s)
=== RUN   TestBatcher/rpc_batch/Balance/eth
--- PASS: TestBatcher/rpc_batch/Balance/eth (0.00s)
--- PASS: TestBatcher/rpc_batch/Balance (0.00s)
--- PASS: TestBatcher/rpc_batch (0.00s)
=== RUN   TestBatcher/multicall
=== RUN   TestBatcher/multicall/Balance
=== RUN   TestBatcher/multicall/Balance/erc20
--- PASS: TestBatcher/multicall/Balance/erc20 (0.00s)
=== RUN   TestBatcher/multicall/Balance/eth
--- PASS: TestBatcher/multicall/Balance/eth (0.00s)
--- PASS: TestBatcher/multicall/Balance (0.00s)
--- PASS: TestBatcher/multicall (0.00s)
--- PASS: TestBatcher (0.03s)
PASS
```

Let say multicall balance change and has a bug for eth, we will have
```
=== RUN   TestBatcher
=== RUN   TestBatcher/rpc_batch
=== RUN   TestBatcher/rpc_batch/Balance
=== RUN   TestBatcher/rpc_batch/Balance/erc20
--- PASS: TestBatcher/rpc_batch/Balance/erc20 (0.00s)
=== RUN   TestBatcher/rpc_batch/Balance/eth
--- PASS: TestBatcher/rpc_batch/Balance/eth (0.00s)
--- PASS: TestBatcher/rpc_batch/Balance (0.00s)
--- PASS: TestBatcher/rpc_batch (0.00s)
=== RUN   TestBatcher/multicall
=== RUN   TestBatcher/multicall/Balance
=== RUN   TestBatcher/multicall/Balance/erc20
--- PASS: TestBatcher/multicall/Balance/erc20 (0.00s)
=== RUN   TestBatcher/multicall/Balance/eth
    batch_test.go:103: error
--- FAIL: TestBatcher/multicall/Balance/eth (0.00s)
--- FAIL: TestBatcher/multicall/Balance (0.00s)
--- FAIL: TestBatcher/multicall (0.00s)
--- FAIL: TestBatcher (0.02s)
```
and if you have an IDE
image

My point here is to show that this test can handle all the function expose by Batcher, without loosing the granularity when a specific feature is broken.

backend/pkg/executionlayer/evm/batch_test.go
Tangui-Bitfly: Yes and no, benchmarks have a tighter structure.
I will remove it.

backend/pkg/commons/db2/metadataupdates/store.go
Tangui-Bitfly: I will wait a bit more before adding test to those store because
1. I want to regroup them soon
2. I don't want to over engineer test for those package if we migrate to clickhouse in a reasonable future, because those store are linked to an nosql database (not so much bigtable).

backend/pkg/executionlayer/evm/batch.go
invis-bitfly: `foo` is not a valid `ChainID`, so there is no reason for handling it as a string. it will always be a number. esp since the parent context fetches it from the consensus layer node, which can only return a number.

backend/pkg/executionlayer/evm/batch.go
invis-bitfly: the scenario is us deploying the eth1indexer to a new chain (rollup, L2), but our `multicallerFor` having no entry for that `chainID`. Instead of having to go through `ticket => code change => review => release`, devops could just pass it over the config/env. 

if the argument is that the caller should add support for config overwrites, then `NewBatcher` shouldn't exist and each caller should implement the same `check config => check mapping if not set => use NewMulticallBatcher if match / fallback to NewRPCBatcher if not`. i see that as a strictly worse option than `NewBatcher` having the support implemented at one place.

backend/pkg/executionlayer/evm/batch.go
invis-bitfly: > I'm a partisan of not exposing too much configuration params when possible. Each params exposed is a complexity that leak to higher layers and to the config file.

There is no overhead to the higher layers. All that has to be added is a `utils.Config.MultiCallBatchLimit` in here, and if its `0` use `5_000` instead.
> For the current usage, I don't see us having an issue with this hardcoded limit.

That is the issue. If i add something new, and it passes all testing and runs fine in prod for months, but then suddenly something causes the gas usage to go to high (badly implemented contract that has to be indexed or w/e) or external node limits go down, devops has to create a hot-patch release instead of lowering the limit over config/env.
> I just want to write code for a practical reason not for an hypothetical future than never come.

there are limits to everything. We don't need to handle cosmic radiation flipping bits. But this feels like a pretty reasonable case, esp if its meant to be a generic implementation used at many places.

backend/pkg/executionlayer/evm/batch.go
invis-bitfly: The limits of `RPCBatch` are (to be) handled by the Caller in the current (and previous) implementation. If you call `RPCBatch` with 1mil addresses it/the node will fail to complete that request. But the safety measures you added to `MulticallBatcher` allow me to just throw 1mil addresses at it and get a response. 

We both know that we can't blindly send 1mil addresses to `NewBatcher` and expect it to work in all cases. But we can't expect everyone to know that, so the best we can do is have the same behavior guarantees in both main and fallback implementations (that is, chunk the requests so it accepts any input amount).

backend/pkg/executionlayer/evm/batch.go
invis-bitfly: > This will never error. The only error that can happen here are when coding, for example if I don't provide an address but a number.

so... it will error. Currently it won't, but if i mess up the implementation of something new, instead of getting a clear error messages i will get incorrect results. Why not protect against that?

backend/pkg/executionlayer/evm/batch_test.go
invis-bitfly: It growing over time feels like a bad thing, not a good thing. 
Why isn't `Batcher` fully tested so that i can test `Balance` in complete isolation, under the assumption that `Batcher` is working correctly?

If i create something new, lets call it the `ContractDissector`, and it uses `Batcher`, among many other things, I need to write `Batcher` tests in addition to my `ContractDissector` tests? Imagine if everything else i used in `ContractDissector` also asked for that, at some point I'm spending more time writing tests for completely unrelated things rather than testing my actual implementation.

tl;dr: `Balance` is on a higher implementation level than `Batcher` (which is on a higher level than `RPCBatch` & `MulticallBatcher`), but the current test layout does not follow that (`TestBatcher/multicall/Balance` is `L2/L1/L3`)

backend/pkg/commons/db2/database/option.go
invis-bitfly: this was skipped i think?

Tangui-Bitfly: Sorry missed it.
It does apply the different option functions provided.
I will rename it if you find it confusing.

backend/pkg/executionlayer/evm/batch.go
Tangui-Bitfly: That make the signature of `NewBatcher` confusing IMO => you can provide an address or not, in any case it will work.
The multicaller address is an implementation detail, and since it's deployed at the same address on 250+ chains, this check is there more for peace of mind than anything else.
But fine I will change `NewBatcher` signature.

backend/pkg/executionlayer/evm/batch.go
Tangui-Bitfly: Because having a config file close to 300 lines is not an overhead ? The current version is hard to run, adding more and more config options make it slowly worse. We need to subtract, we cannot keep piling junk over junk.

And for your example, if something break in prod and the solution is just to increase the limit ... it's kinda risky, I mean if such case happen we need to understand why, adding a bandaid will not be enough in the long run.

Anyway, that was my opinion, now I will do the change.

backend/pkg/executionlayer/evm/batch.go
Tangui-Bitfly: I don't see any good way to patch this.
IMO it's out of scope of this PR, we could create another ticket and work on it later. But I really don't see this as a priority.

backend/pkg/executionlayer/evm/batch_test.go
Tangui-Bitfly: maybe have a look at https://github.com/gobitfly/beaconchain/pull/1319/files#diff-79e50ca28c249eb82f2609e619f3666cf1d141bb29ae5df1482e3977b1cea530
Test growing over time is ok, it reduce the number of tests and of boilerplate. I prefer to have a compact focused test file rather than an enormous one with test going everywhere with boilerplate code repeated in each.

invis-bitfly (approved):

PR #1319, opened 2025-02-07
refactor(token price): refactor token price

Tangui-Bitfly:
* introduce interface for external price fetching
* use mutlicall for erc20 supply retrieval

### DIFF
--- a/backend/pkg/commons/db2/metadata/metadata.go
+++ b/backend/pkg/commons/db2/metadata/metadata.go
@@ -4,8 +4,11 @@ import (
 	"fmt"
 	"math/big"
 
+	"github.com/ethereum/go-ethereum/common"
+
 	"github.com/gobitfly/beaconchain/pkg/commons/db2/database"
 	"github.com/gobitfly/beaconchain/pkg/commons/db2/metadataupdates"
+	"github.com/gobitfly/beaconchain/pkg/commons/types"
 )
 
 type Balance struct {
@@ -33,4 +36,35 @@ func (store Store) UpdateBalance(chainID string, balances []Balance) error {
 		})
 	}
 	return store.db.BulkAdd(updates)
+}
+
+func (store Store) UpdateToken(chainID string, tokens []*types.ERC20TokenPrice) error {
+	updates := make(map[string][]database.Item)
+	for _, token := range tokens {
+		key := fmt.Sprintf("%s:%x", chainID, token.Token)
+		updates[key] = append(updates[key], database.Item{
+			Family: erc20MetadataFamily,
+			Column: erc20ColumnPrice,
+			Data:   token.Price,
+		})
+		updates[key] = append(updates[key], database.Item{
+			Family: erc20MetadataFamily,
+			Column: erc20ColumnTotalSupply,
+			Data:   token.TotalSupply,
+		})
+	}
+	return store.db.BulkAdd(updates)
+}
+
+func (store Store) TokenPrice(chainID string, token common.Address) (*types.ERC20TokenPrice, error) {
+	key := fmt.Sprintf("%s:%x", chainID, token.Bytes())
+	row, err := store.db.GetRow(key)
+	if err != nil {
+		return nil, err
+	}
+	return &types.ERC20TokenPrice{
+		Token:       token.Bytes(),
+		Price:       row.Values[fmt.Sprintf("%s:%s", erc20MetadataFamily, erc20ColumnPrice)],
+		TotalSupply: row.Values[fmt.Sprintf("%s:%s", erc20MetadataFamily, erc20ColumnTotalSupply)],
+	}, nil
 }
--- a/backend/pkg/commons/db2/metadata/tables.go
+++ b/backend/pkg/commons/db2/metadata/tables.go
@@ -5,9 +5,13 @@ const Table = "metadata"
 var Schema = map[string][]string{
 	Table: {
 		accountFamily,
+		erc20MetadataFamily,
 	},
 }
 
 const (
-	accountFamily = "a"
+	accountFamily          = "a"
+	erc20MetadataFamily    = "erc20"
+	erc20ColumnPrice       = "PRICE"
+	erc20ColumnTotalSupply = "TOTALSUPPLY"
 )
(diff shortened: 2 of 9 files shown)


backend/pkg/commons/db2/metadata/tables.go
```diff
@@ -5,9 +5,13 @@ const Table = "metadata"
 var Schema = map[string][]string{
 	Table: {
 		accountFamily,
+		erc20MetadataFamily,
 	},
 }
 
 const (
-	accountFamily = "a"
+	accountFamily          = "a"
```
mccreedy-bitfly, line 13: What is this exactly?

backend/pkg/executionlayer/tokenprice.go
```diff
@@ -0,0 +1,129 @@
...
+
+type TokenPricer struct {
+	store    TokenStore
+	tokens   erc20.ERC20TokenList
+	external ExternalPricer
+	batcher  evm.Batcher
+	chainID  string
+}
+
+func NewTokenPricer(store TokenStore, chainID string, external ExternalPricer, tokens erc20.ERC20TokenList, batcher evm.Batcher) *TokenPricer {
```
mccreedy-bitfly, line 35: IMO this should be a repository

backend/pkg/commons/erc20/erc20.go
```diff
@@ -59,7 +59,7 @@ type ERC20TokenList struct {
 type ERC20TokenDetail struct {
 	Address  string `json:"address"`
 	Owner    string `json:"-"`
-	ChainID  int64  `json:"chainId"`
+	ChainID  int64  `json:"chainId"` // not used
```
Monika-Bitfly: If not used, can we remove it?

backend/pkg/executionlayer/tokenprice.go
```diff
@@ -0,0 +1,129 @@
...
+	prices, err := t.external.GetPrices(tokens)
+	if err != nil {
+		return fmt.Errorf("cannot get token prices: %w", err)
+	}
+
+	supplies, err := evm.ERC20Supply(t.batcher, tokens)
+	if err != nil {
+		return fmt.Errorf("cannot get token supplies: %w", err)
+	}
+	for i := 0; i < len(prices); i++ {
```
Monika-Bitfly, line 62: small comment explaining this operation would be nice here (or moving it into a small separate function)

backend/pkg/commons/db2/metadata/tables.go
```diff
@@ -5,9 +5,13 @@ const Table = "metadata"
...
+	erc20MetadataFamily    = "erc20"
+	erc20ColumnPrice       = "PRICE"
+	erc20ColumnTotalSupply = "TOTALSUPPLY"
```
Monika-Bitfly, line 16: I think we can remove now `ERC20_COLUMN_TOTALSUPPLY` and `ERC20_COLUMN_PRICE` const from `db/bigtable_eth1.go`

backend/pkg/commons/db2/metadata/metadata.go
```diff
@@ -34,3 +37,34 @@ func (store Store) UpdateBalance(chainID string, balances []Balance) error {
...
+		updates[key] = append(updates[key], database.Item{
+			Family: erc20MetadataFamily,
+			Column: erc20ColumnTotalSupply,
+			Data:   token.TotalSupply,
+		})
+	}
+	return store.db.BulkAdd(updates)
+}
+
+func (store Store) TokenPrice(chainID string, token common.Address) (*types.ERC20TokenPrice, error) {
```
Monika-Bitfly, line 59: I think we're not using it anywhere apart from the tests

backend/pkg/executionlayer/tokenprice_test.go
```diff
@@ -0,0 +1,155 @@
...
+
+func TestLlama_GetPrices(t *testing.T) {
+	tests := []struct {
+		name  string
+		coins []common.Address
+		resp  defiLlamaResponse
+		want  []*types.ERC20TokenPrice
+	}{
+		{
+			name:  "happy case",
```
Monika-Bitfly, line 36: Would be great to add tests for case with error

backend/pkg/executionlayer/tokenprice_test.go
```diff
@@ -0,0 +1,155 @@
...
+
+func (s stubExternalPricer) GetPrices(tokens []common.Address) ([]*types.ERC20TokenPrice, error) {
+	var prices []*types.ERC20TokenPrice
+	for _, address := range tokens {
+		prices = append(prices, s.prices[address])
+	}
+	return prices, s.err
+}
+
+func TestTokenPricer(t *testing.T) {
```
Monika-Bitfly, line 87: Also here it would be great to add tests for failed token update

backend/pkg/commons/db2/metadata/metadata.go
Tangui-Bitfly: Yes, but this but it seems to be important to be able to retrieve the data we write.
For now the reading is done by `db.Bigtable` object, I do not plan to rework that object in one go, but rather bite by bite.

backend/pkg/commons/db2/metadata/tables.go
Tangui-Bitfly: we cannot, they are used by `db.Bigtable` object

backend/pkg/executionlayer/tokenprice_test.go
Tangui-Bitfly: I don't think adding test cases that create moke http err, wrongly formated responses, wrong http code, ... add any real value beside increasing or coverage from 78% to 79%.
If we had a fallback mechanism, yes sure it would make sense.
But in the current state, I don't want to increase the complexity of our tests without a real reason.

backend/pkg/executionlayer/tokenprice_test.go
Tangui-Bitfly: Again, this test is complex enough. I don't see the value of adding cases just for testing that when there is an error, we get an error.
It would make sense if we had something that should not happen in case of error (like a delete) but that not the case here.

PR #1327, opened 2025-02-11
test: break up blobindexer logic into smaller func and add unit tests

Monika-Bitfly:
- Refactored `blobindexer.go` functions in the `/blobindexer` package by breaking them into smaller, more readable functions
- Added `S3Client` and consapi `Client` mocks using mockery package
- Added `blobindexer_test.go` unit tests
- Fixed `/consapi` methods typo


backend/pkg/blobindexer/s3/client.go
```diff
@@ -0,0 +1,13 @@
+package s3
+
+import (
+	"context"
+
+	"github.com/aws/aws-sdk-go-v2/service/s3"
+)
+
+type S3Client interface {
```
Tangui-Bitfly: rather than creating a s3 package with an S3Client, it could be the good opportunity for a minor refactor using repository pattern.
The package could be `blobstore`, have an interface `BlobStore` (equivalent to BlobStoreRepository) and an implementation `S3`.
```go
type BlobStore interface {
	Put(ctx context.Context, bucket string, key string, blob Blob) error
	Exist(ctx context.Context, bucket string, key string) (bool, error)
	Get(ctx context.Context, bucket string, key string) (Blob, error)
}

type Blob struct {
	Body io.Reader
	ContentType *string
	Metadata map[string]string
}
```
The goal is to remove dependency with s3 for the blobindexer and allow easier testing (no complex mock required when the interface is small).

Tangui-Bitfly: also this package `blobstore` can be in `db2.Database` making it more accessible for future reuse.

backend/pkg/blobindexer/blobindexer.go
```diff
@@ -392,22 +197,17 @@ func (bi *BlobIndexer) GetIndexerStatus() (*BlobIndexerStatus, error) {
...
 		Key:    &key,
 	})
 	if err != nil {
-		// If the object that you request doesn’t exist, the error that Amazon S3 returns depends on whether you also have the s3:ListBucket permission. If you have the s3:ListBucket permission on the bucket, Amazon S3 returns an HTTP status code 404 (Not Found) error. If you don’t have the s3:ListBucket permission, Amazon S3 returns an HTTP status code 403 ("access denied") error.
-		var httpResponseErr *awshttp.ResponseError
-		if errors.As(err, &httpResponseErr) && (httpResponseErr.HTTPStatusCode() == 404 || httpResponseErr.HTTPStatusCode() == 403) {
-			return &BlobIndexerStatus{}, nil
-		}
-		return nil, err
+		return handleGetObjectError(err)
```
mccreedy-bitfly: This should maybe all be combined in to a single function (getBlobObject) rather than splitting out the error in to its own function

backend/pkg/blobindexer/blobindexer.go
```diff
@@ -113,97 +121,254 @@ func (bi *BlobIndexer) index() error {
...
+
+	if err := bi.indexBlobsInBatches(status, headHeader, finalizedHeader, startSlot); err != nil {
+		return err
+	}
+	return nil
+}
+
+func (bi *BlobIndexer) getIndexerStatusFromS3() (*BlobIndexerStatus, error) {
+	key := fmt.Sprintf("%s/blob-indexer-status.json", bi.networkID)
+	label := "blobindexer_get_indexer_status"
```
Tangui-Bitfly, line 174: This is a suggestion not a required change:
Maybe we could wrap the metric inside a blobstore.BlobStore impl
```go
type BlobStoreWithMetric struct {
	blobstore blobstore.BlobStore
}

func (store BlobStoreWithMetric) Put(ctx context.Context, bucket string, key string, blob Blob) error {
	label := "blobindexer_get_indexer_status"
	start := time.Now()
	defer func() {
		metrics.TaskDuration.WithLabelValues(logLabel).Observe(time.Since(start).Seconds())
	}()
	store.blobstore.Put(ctx, bucket, key, blob)
}
// same for Exist and get
```

Tangui-Bitfly: This would remove the need of func `getObjectFromS3`

backend/pkg/blobindexer/blobindexer.go
```diff
@@ -113,97 +121,254 @@ func (bi *BlobIndexer) index() error {
...
+	obj, err := bi.getObjectFromS3(key, label)
+	if err != nil {
+		return nil, fmt.Errorf("error getting object from S3 with key %s: %w", key, err)
```
Tangui-Bitfly, line 177: if there is an error, the message will be 
`error getting object from S3 with key <key>: error getting object from S3 with key <key>:  <S3 err>`
There is the same message repeated twice, remove one of them

backend/pkg/blobindexer/blobindexer.go
```diff
@@ -113,97 +121,254 @@ func (bi *BlobIndexer) index() error {
...
+	}
+
+	body, err := json.Marshal(&status)
+	if err != nil {
+		return err
+	}
+
+	err = bi.putObjectInS3(key, "blobindexer_put_indexer_status", &contentType, body, metadata)
+	if err != nil {
+		return fmt.Errorf("error putting object in S3 with key %s: %w", key, err)
```
Tangui-Bitfly, line 203: same

backend/pkg/blobindexer/blobindexer.go
```diff
@@ -287,165 +418,191 @@ func (bi *BlobIndexer) index() error {
...
+		if err := bi.checkIfObjectExistsInS3(key, label); err != nil {
+			return err
+		}
 	}
-	err = g.Wait()
+
+	err := bi.putObjectInS3(key, "blobindexer_put_blob", nil, blob.Blob, metadata)
 	if err != nil {
-		return len(blobSidecar.Data), fmt.Errorf("error indexing blobs at slot %v: %w", slot, err)
+		return fmt.Errorf("error putting object in S3: %s (%v/%v): %w", key, blob.SignedBlockHeader.Message.Slot, blob.Index, err)
```
Tangui-Bitfly, line 521: same, message could be
```go
fmt.Errorf("blob %d:%d: %w", blob.SignedBlockHeader.Message.Slot, blob.Index, err)
```

Tangui-Bitfly (approved):

Tangui-Bitfly (approved):
Approved but I did not review the 1700 line of tests over 23 different tests.
That's is 3 times the amount of actual code written. It's too much.

invis-bitfly (approved):
need to resolve conflicts before merge

PR #1329, opened 2025-02-13
test: `GetValidatorDashboardSummaryChart`

LuccaBitfly:
Adds unit tests for the summary chart endpoint. It may be best to look at each commit individually


backend/pkg/api/handlers/public.go
```diff
@@ -1410,107 +1357,6 @@ func (h *HandlerService) PublicGetValidatorDashboardBlocks(w http.ResponseWriter
...
-//	@Produce		json
-//	@Param			dashboard_id	path		string	true	"The ID of the dashboard."
-//	@Param			group_id		path		integer	true	"The ID of the group."
-//	@Param			timestamp		path		integer	true	"The timestamp to get data for."
-//	@Param			modes			query		string	false	"Provide a comma separated list of protocol modes which should be respected for validator calculations. Possible values are `rocket_pool``."
-//	@Param			aggregation		query		string	false	"Aggregation type to get data for."	Enums(epoch, hourly, daily, weekly)	Default(hourly)
-//	@Success		200				{object}	types.GetValidatorDashboardGroupHeatmapResponse
-//	@Failure		400				{object}	types.ApiErrorResponse
-//	@Router			/validator-dashboards/{dashboard_id}/groups/{group_id}/heatmap/{timestamp} [get]
-func (h *HandlerService) PublicGetValidatorDashboardGroupHeatmap(w http.ResponseWriter, r *http.Request) {
```
sasha-bitfly, line 1476: Do we remove `/{dashboard_id}/heatmap` and `/{dashboard_id}/groups/{group_id}/heatmap/{timestamp} `endpoint? Is the idea to use `/{dashboard_id}/summary-chart` instead?

backend/pkg/api/handlers/validator_dashboard.go
```diff
@@ -105,3 +106,146 @@ func (h *HandlerService) GetValidatorDashboardGroupSummary(ctx context.Context,
...
+	efficiencyType enums.VDBSummaryChartEfficiencyType
+	aggregation    enums.ChartAggregation
+	afterTs        *uint64
+	beforeTs       *uint64
+}
+
+const chartDatapointLimit uint64 = 200
+
+// resolveAndValidateTimestamps calculates the missing timestamps for the chart data and/or validates them.
+func resolveAndValidateTimestamps(
```
sasha-bitfly, line 158: Please, write a unit test for this function

LuccaBitfly: already did https://github.com/gobitfly/beaconchain/blob/0eca5e4f545096357d7d6996ba28f4c3ece665e7/backend/pkg/api/handlers/validator_dashboard_test.go#L243

backend/pkg/api/handlers/public.go
LuccaBitfly: I've removed them because currently they're not really implemented (https://github.com/gobitfly/beaconchain/blob/0eca5e4f545096357d7d6996ba28f4c3ece665e7/backend/pkg/api/data_access/vdb_heatmap.go) and because they reused some of the code that the old chart handler did. Before I do the work to convert them to our new format, I'd just wait until we actually implement them to not waste time. The deleted code still exists in the commit history if we ever need it.

backend/pkg/api/handlers/validator_dashboard.go
LuccaBitfly: and also https://github.com/gobitfly/beaconchain/blob/0eca5e4f545096357d7d6996ba28f4c3ece665e7/backend/pkg/api/handlers/validator_dashboard_test.go#L355

sasha-bitfly (approved):

PR #1330, opened 2025-02-13
test: add unit test for `validationError` and all regex related checker funcs

LuccaBitfly:
See: BEDS-1187

### DIFF
--- a/backend/pkg/api/handlers/input_validation.go
+++ b/backend/pkg/api/handlers/input_validation.go
@@ -15,7 +15,6 @@ import (
 	"github.com/ethereum/go-ethereum/common/hexutil"
 	"github.com/gobitfly/beaconchain/pkg/api/enums"
 	"github.com/gobitfly/beaconchain/pkg/api/types"
-	"github.com/gorilla/mux"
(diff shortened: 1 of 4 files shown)


backend/pkg/api/handlers/input_validation.go
```diff
@@ -253,6 +252,7 @@ func (v *validationError) checkAdConfigurationKeys(keysString string) []string {
 	}
 	var keys []string
 	for _, key := range splitParameters(keysString, ',') {
+		key = strings.TrimSpace(key)
```
sasha-bitfly, line 255: Is this change covered with unit test?

sasha-bitfly: Other than that LGTM

sasha-bitfly (approved):

mccreedy-bitfly (approved):

backend/pkg/api/handlers/input_validation.go
LuccaBitfly: Isn't covered, I have the code locally but didn't include in this PR by mistake. I'll just include it in the next one

PR #1335, opened 2025-02-14
BEDS-839/remove timestamp pending and exiting validators

enzo-bitfly:
**Changes besides ticket description:**
- fix dependency vulnerabilities
- use api-provided `PostValidatorDashboardValidatorsRequest` for validator dashboard `update` api call - api added this type recently and we needed to use it in the frontend instead of our hardcoded type


frontend/components/dashboard/validator/subset/DashboardValidatorSubsetList.vue
```diff
@@ -105,12 +105,16 @@ function mapDutyLinks(
...
+
   if (path) {
     return dutyObjects.map(o => ({
       label: `${formatValue ? formatNumber(o) : o}`,
       to: `${path}${o}`,
     }))
   }
+  else if (props.category === 'pending' || props.category === 'exiting') {
+    return []
+  }
```
marcel-bitfly: would this also work?
```suggestion
  if (props.category === 'pending' || props.category === 'exiting') return []
 
  if (path) {
    return dutyObjects.map(o => ({
      label: `${formatValue ? formatNumber(o) : o}`,
      to: `${path}${o}`,
    }))
  }
```

frontend/components/dashboard/validator/subset/DashboardValidatorSubsetList.vue
```diff
@@ -139,9 +143,9 @@ function mapDutyLinks(
         </BcLink>
         <template v-if="v.duty_objects?.length">
           <span class="round-brackets">
-            <span class="label">{{ mapDutyLabel(v.duty_objects) }}</span>
+            <span class="label">{{ createDutyLabels(v.duty_objects) }}</span>
             <template
-              v-for="link in mapDutyLinks(v.duty_objects)"
+              v-for="link in createDutyLinksData(v.duty_objects)"
```
marcel-bitfly, line 150: would this make sense?
```suggestion
              v-for="link in dutylinks(v.duty_objects)"
```

frontend/components/dashboard/validator/subset/DashboardValidatorSubsetList.vue
```diff
@@ -18,7 +18,6 @@ const { t: $t } = useTranslation()
 const paging = ref<null | Paging>(null)
 const cursor = ref<Cursor>(undefined)
 const VALIDATORS_PER_PAGE = 100
-
```
marcel-bitfly: np: ideally we are very rigorous with touching `whitespace`, as it will make it `line blaming` a lot harder the overtime

frontend/components/dashboard/validator/subset/DashboardValidatorSubsetList.vue
enzo-bitfly: Good catch, I applied your suggestion

frontend/components/dashboard/validator/subset/DashboardValidatorSubsetList.vue
enzo-bitfly: It's a function so the name is an action + high context + low context. Also, it returns link data, not a link.

frontend/components/dashboard/validator/subset/DashboardValidatorSubsetList.vue
enzo-bitfly: That was an accident, I reverted it

frontend/nuxt.config.ts
```diff
@@ -122,5 +118,6 @@ export default defineNuxtConfig({
   },
   postcss: { plugins: { autoprefixer: {} } },
   eslint: { config: { stylistic: true } },
+
```
marcel-bitfly: np: whitespace

marcel-bitfly (approved):

frontend/nuxt.config.ts
enzo-bitfly: Don't know what happened there. Fixed

PR #1336, opened 2025-02-17
Nobids/refactor store

Tangui-Bitfly:
* regroup all store into one object
* remove data filters, it can restored later when needed
* fix: contract updates where stored in metadataupdates and not metadata


backend/pkg/commons/db2/keys.go
```diff
@@ -0,0 +1,19 @@
...
+
+// toSuccessor add suffix ";" has it comes after ":" in the ascii order
+// this is a simple way to have an infinite bound limit
+// prefix must be a real prefix and not a key
+func toSuccessor(prefix string) string {
+	return prefix + ";"
+}
+
+const (
+	maxInt                       = 9223372036854775807
```
Monika-Bitfly, line 11: Can we either use the values from `bigtable_eth1.go` or remove them pls?

backend/pkg/commons/db2/keys_updates.go
```diff
@@ -0,0 +1,42 @@
...
+
+import (
+	"fmt"
+
+	"github.com/gobitfly/beaconchain/pkg/commons/db2/database"
+	"github.com/gobitfly/beaconchain/pkg/commons/utils"
+)
+
+const (
+	balanceKey = "B"
```
Monika-Bitfly, line 11: this should be stored inside `keys.go`

backend/pkg/commons/db2/keys_metadata.go
```diff
@@ -39,3 +42,11 @@ func encodeIsContractUpdateTs(block_number, tx_idx, trace_idx uint64) (int64, er
 
 	return int64(res * timestampGBTScale), nil
 }
+
+func addressFromKey(key string) common.Address {
```
Monika-Bitfly, line 46: these functions don't offer much benefit tbh since they are one liner and just split a string and convert it to an address, it might be simpler to just do it directly in the code

backend/pkg/executionlayer/balance_test.go
```diff
@@ -5,57 +5,50 @@ import (
 	"strings"
 	"testing"
 
-	"github.com/gobitfly/beaconchain/pkg/commons/db2/metadata"
-	"github.com/gobitfly/beaconchain/pkg/commons/db2/metadataupdates"
+	"github.com/gobitfly/beaconchain/pkg/commons/db2"
 	"github.com/gobitfly/beaconchain/pkg/executionlayer/evm"
 )
 
 func TestBalanceUpdater_UpdateBalancesErr(t *testing.T) {
```
Monika-Bitfly, line 89: I think the tests are missing successful balance update

backend/pkg/executionlayer/indexer_test.go
```diff
@@ -9,21 +9,16 @@ import (
...
-	"github.com/gobitfly/beaconchain/pkg/commons/db2/data"
+	"github.com/gobitfly/beaconchain/pkg/commons/db2"
 	"github.com/gobitfly/beaconchain/pkg/commons/db2/database"
 	"github.com/gobitfly/beaconchain/pkg/commons/db2/database/databasetest"
-	"github.com/gobitfly/beaconchain/pkg/commons/db2/metadataupdates"
 	"github.com/gobitfly/beaconchain/pkg/commons/rpc"
 	"github.com/gobitfly/beaconchain/pkg/commons/types"
 )
 
 func TestIndexerWithBigTable(t *testing.T) {
```
Monika-Bitfly, line 19: Would be great to also add tests for the remaining transformers ITX, BlobTx, Withdrawals and Uncles

backend/pkg/commons/db2/keys.go
Tangui-Bitfly: No i don't want any dependency between `db2` and `db`, and `db.MAX_INT` is still used.

backend/pkg/commons/db2/keys_metadata.go
Tangui-Bitfly: This is not easily understandable
```go
common.HexToAddress(strings.Split(key, ":")[2])
```
To understand that you have to know the key db is a composed of different strings and that index 2 is the address
This is easily understandable
```go
addressFromKey(row.Key)
```

backend/pkg/commons/db2/keys_updates.go
Tangui-Bitfly: I tried to regroup keys definition to there tables. Regrouping all the key in one file make things harder to understand where a key is used.
The issue here is more that the file `keys.go` is badly named. `keys_utils.go` would be more appropriated.

backend/pkg/executionlayer/indexer_test.go
Tangui-Bitfly: Yes but that requires a lot of work to integrate function that generate those.
Honestly, I don't like this is test. It test directly keys in bigtable and not object retrieval. I did it while I was migrating things, I think we will remove it soon.

PR #1339, opened 2025-02-19
refactor(executionlayer/reorg): separate reorg process + tests

Tangui-Bitfly:
* extract reorg logic from main and `db/bigtable`
* put cache interface into `db2/database`
* add some options filter for `db2/database/Database`
* `db2/database/databasetest` export a test utility function to run a Redis in docker using testcontainer
  * skip test if docker is not running
* migrate indexing func from main and `db/bigtable` to `executionlayer/indexer`
  * I'm not happy with the current implementation, I will work on an overall refactor of the main and indexer interaction later on


backend/pkg/commons/db2/database/bigtable.go
```diff
@@ -437,5 +444,33 @@ func bigtableReadOptions(options options, rowRange bigtable.RowRange) []bigtable
...
+				bigtable.Timestamp(options.TimestampRangeFilter[0]),
+				bigtable.Timestamp(options.TimestampRangeFilter[1]),
+			),
+		)
+	}
+	if options.WithoutValue {
+		filters = append(filters, bigtable.StripValueFilter())
+	}
+	if len(filters) != 0 {
+		if len(filters) == 1 {
```
Monika-Bitfly, line 469: We don't need to check if the filters len equals 1, can directly pass the slice to `bigtable.ChainFilters(filters...)`

backend/pkg/commons/db2/database/bigtable.go
```diff
@@ -437,5 +444,33 @@ func bigtableReadOptions(options options, rowRange bigtable.RowRange) []bigtable
...
+		filters = append(filters, bigtable.ColumnFilter(options.ColumnFilter))
+	}
+	if options.FamilyFilter != "" {
+		filters = append(filters, bigtable.FamilyFilter(options.FamilyFilter))
+	}
+	if options.TimestampRangeFilter != nil {
+		filters = append(filters,
+			bigtable.TimestampRangeFilterMicros(
+				bigtable.Timestamp(options.TimestampRangeFilter[0]),
+				bigtable.Timestamp(options.TimestampRangeFilter[1]),
```
Monika-Bitfly, line 461: this will throw an error if only one of the timestamp filters is supplied eg. only start time or only end time is provided, so would be good to check for TimestampRangeFilter length before appending the values

backend/pkg/commons/db2/database/cache.go
```diff
@@ -0,0 +1,98 @@
...
+type RemoteCache interface {
+	Set(ctx context.Context, key string, value []byte, expiration time.Duration) error
+	Get(ctx context.Context, key string) ([]byte, error)
+}
+
+type MemCache struct {
+	values map[string][]byte
+}
+
+func (m *MemCache) Set(ctx context.Context, key string, value []byte, expiration time.Duration) error {
```
Monika-Bitfly, line 28: ctx is unused

backend/pkg/commons/db2/database/option.go
```diff
@@ -39,21 +49,39 @@ func WithRowKeyFilter(regex string) Option {
...
+
+func WithOpenRange() Option {
 	return func(opts *options) {
-		opts.OpenRange = r
+		opts.OpenRange = true
 	}
 }
 
-func WithClosedOpenRangeOption(r bool) Option {
+func WithClosedOpenRangeOption() Option {
```
Monika-Bitfly, line 80: would be great to drop a comment here explaining whats ClosedOpenRange and OpenCloseRange option

backend/pkg/commons/db2/db.go
```diff
@@ -8,16 +8,32 @@ import (
 
 type Store interface {
 	AddIndexedBlock(block IndexedBlock) error
+	RevertBlock(chainID string, number uint64, blockHash []byte) error
 }
 
 type MetadataStore interface {
-	UpdateContract(chainID string, blockNumber uint64, updates []ContractUpdateWithAddress) error
```
Monika-Bitfly, line 14: Was the removal of the `UpdateContract` func intended?

backend/pkg/commons/db2/store.go
```diff
@@ -258,6 +262,176 @@ func (store StoreV1) TokenPrice(chainID string, token common.Address) (*types.ER
...
+
+func (store StoreV1) GetLastBlockInDataTable(chainID string) (uint64, error) {
+	prefix := chainID + ":B:"
+	rows, err := store.data.Read(prefix, database.WithLimit(1), database.WithoutValue())
+	if err != nil {
+		return 0, err
+	}
+	if len(rows) != 1 {
+		return 0, nil
+		//		return 0, fmt.Errorf("cannot found row with prefix %s in data table", prefix)
```
Monika-Bitfly:
```suggestion
```

backend/pkg/commons/db2/store.go
```diff
@@ -258,6 +262,176 @@ func (store StoreV1) TokenPrice(chainID string, token common.Address) (*types.ER
...
+	if err != nil {
+		return err
+	}
+
+	// handle contract state updates
+	// TODO: this is potentially very resources consuming
+	rows, err := store.metadata.Read(fmt.Sprintf("%s:S:", chainID),
+		database.WithFamilyFilter(accountFamily),
+		database.WithColumnFilter(accountIsContractColumn),
+		database.WithTimestampRangeFilter(start, end-1),
```
Monika-Bitfly, line 349: why is `end-1` used in `WithTimestampRangeFilter` inside the `Read()` func but `end` is used in DeleteRowsWithKeys?

backend/pkg/executionlayer/indexer.go
```diff
@@ -1,39 +1,163 @@
...
+	return nil
+}
+
+// IndexEvents retrieve read the types.Eth1Block from the store and apply the transformers to them
+func (indexer *Indexer) IndexEvents(chainID string, start, end uint64, concurrency uint64) error {
+	retrieval := new(errgroup.Group)
+	retrieval.SetLimit(int(concurrency))
+
+	indexing := new(errgroup.Group)
+	indexing.SetLimit(int(concurrency * concurrency))
```
Monika-Bitfly, line 96: Why setting it this way? Kinda worried it might overwhelm the system if concurrency values are set too high

backend/pkg/executionlayer/indexer.go
```diff
@@ -1,39 +1,163 @@
...
+		retrieval.Go(func() error {
+			log.Infof("querying blocks from %v to %v", firstBlock, lastBlock)
+			high := lastBlock
+			low := firstBlock
+
+			blocks, err := indexer.store.GetBlocksRange(chainID, high, low)
+			if err != nil {
+				log.Error(err, "error getting blocks descending", 0, map[string]interface{}{"high": high, "low": low})
+			}
+
```
Monika-Bitfly, line 116: would add check for empty blocks like
```
if len(blocks) == 0 {
   return nil
}
```

backend/pkg/executionlayer/reorg.go
```diff
@@ -0,0 +1,114 @@
...
+		nodeBlock, err := r.client.HeaderByNumber(ctx, big.NewInt(int64(i)))
+		if err != nil {
+			return err
+		}
+		dbBlock, err := r.store.GetBlock(r.chainID, i)
+		if err != nil {
+			// exit if we hit a block that is not yet in the db
+			// it means that we never processed that block or that we revert that block
+			if errors.Is(err, database.ErrNotFound) {
+				return nil
```
Monika-Bitfly, line 70: could be useful to add logging if block is not found in db
```suggestion
                                 log.Warnf("block %d not found in the database ", i)
				return nil
```

backend/pkg/executionlayer/reorg.go
```diff
@@ -0,0 +1,114 @@
...
+	// ensure we will not try to retrieve blocks that do not exist
+	depth := r.depth
+	if r.depth > head.Number.Uint64() {
+		depth = head.Number.Uint64()
+	}
+
+	ctx := context.Background()
+
+	// for each block check if block node hash and block db hash match
+	for i := head.Number.Uint64() - depth; i <= head.Number.Uint64(); i++ {
```
Monika-Bitfly, line 60: also could be useful to add concurrency here when querying for blocks

backend/pkg/commons/db2/database/bigtable.go
Tangui-Bitfly: this is needed, without it
```
reorg_test.go:219: table blocks: could not read rows: rpc error: code = InvalidArgument desc = Chain must contain at least two RowFilters
```

backend/pkg/commons/db2/store.go
```diff
@@ -258,6 +262,176 @@ func (store StoreV1) TokenPrice(chainID string, token common.Address) (*types.ER
...
+
+func (store StoreV1) deleteBlockKeys(chainID string, number uint64, blockHash []byte) error {
+	keys, err := store.getBlockKeys(chainID, number, blockHash)
+	if err != nil {
+		return err
+	}
+	if err := store.data.DeleteRowsWithKeys(keys); err != nil {
+		return err
+	}
+	return nil
```
Monika-Bitfly, line 376: just a small suggestion, we could add logging to these functions or inside the `DeleteRowsWithKeys` func to indicate that the rows have been successfully removed

backend/pkg/commons/db2/database/bigtable.go
Tangui-Bitfly: The only way to set `options.TimestampRangeFilter` is with [this func](https://github.com/gobitfly/beaconchain/blob/992af76f9cdad6fb96782eba48cb9b7d8e158b39/backend/pkg/commons/db2/database/option.go#L64)
```go
func WithTimestampRangeFilter(start, end int64) Option {
	return func(opts *options) {
		opts.TimestampRangeFilter = []int64{start, end}
	}
}
```
This function requires two int64, this is enough validation, we don't have to do further checking downstream.

backend/pkg/commons/db2/db.go
Tangui-Bitfly: It something I forgot to remove in an other pr. There is nothing that implements this `UpdateContract` func

backend/pkg/commons/db2/store.go
Tangui-Bitfly: Good question, I don't know that the behavior of the current code.
[source](https://github.com/gobitfly/beaconchain/blob/0fa6e39d4b937f82065bcb28b88e98a958011b02/backend/pkg/commons/db/bigtable_eth1.go#L2413)

backend/pkg/executionlayer/indexer.go
Tangui-Bitfly: I reproduce the current behavior, the only difference here is that it's explicit now rather than hidden before.
This is the [original func](https://github.com/gobitfly/beaconchain/blob/NOBIDS/refactor-store/backend/pkg/commons/db/bigtable_eth1.go#L695)
```go
func (bigtable *Bigtable) IndexEventsWithIndexer(start, end int64, indexer *executionlayer.Indexer, concurrency int64) error {
	g := new(errgroup.Group)
	g.SetLimit(int(concurrency))
	...
	for i := start; i <= end; i += batchSize {
		...
		g.Go(func() error {
			...
			subG := new(errgroup.Group)
			subG.SetLimit(int(concurrency))
```
So we start n go routines and for each of those we start again n go routines.
This is equivalent to limiting the writing globally to n² and this make the real concurrency explicit.

backend/pkg/commons/db2/store.go
Tangui-Bitfly: We must try to limit our usage of logging, in most of the cases it should be done at the top level.
What would be the point of firing a log for a normal operation ? This create monitoring noise. More noise = more chances to miss important log.

backend/pkg/executionlayer/indexer.go
Tangui-Bitfly: If `len(blocks) == 0` the `range` next line will return directly.
Adding this check increase the overall code for no benefit.

backend/pkg/executionlayer/reorg.go
Tangui-Bitfly: That should be in its own task.

backend/pkg/executionlayer/reorg.go
Tangui-Bitfly: No this the expected behavior, look at the comment
```
// exit if we hit a block that is not yet in the db
// it means that we never processed that block or that we revert that block
```
It would create noise for a normal process.

PR #1342, opened 2025-02-21
test: list and number related input validation funcs

LuccaBitfly:


backend/pkg/api/handlers/input_validation_test.go
```diff
@@ -657,3 +659,463 @@ func TestCheckAddress(t *testing.T) {
...
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			var v validationError
+
+			result := v.checkInt(tt.param, "param")
+
+			if tt.expectErr {
```
sasha-bitfly: Can we make it a generic function?
Pretty much every test uses very similar chain of calls
```
		t.Run(tt.name, func(t *testing.T) {
			var v validationError

			result := v.checkWeiMinMax(tt.param, "param", min, max)

			if tt.expectErr {
				assert.True(t, v.hasErrors(), "Expected an error but got none")
				assert.Contains(t, v["param"], tt.errMsg)
			} else {
				assert.False(t, v.hasErrors(), "Expected no errors but found some")
				assert.True(t, tt.expected.Equal(result), "Expected decimal conversion to be correct")
			}
```

sasha-bitfly (changes requested):

sasha-bitfly (approved):

PR #1344, opened 2025-02-21
test: list and dashboard/group id related input validation

LuccaBitfly:


sasha-bitfly (commented):
@LuccaBitfly Does it make sense to make generic func for similar code?
```
		t.Run(tt.name, func(t *testing.T) {
			var v validationError

			result := v.checkAdConfigurationKeys(tt.param)

			if tt.expectErr {
				assert.True(t, v.hasErrors(), "Expected an error but got none")
				assert.Contains(t, v["keys"], tt.errMsg)
			} else {
				assert.False(t, v.hasErrors(), "Expected no errors but found some")
				assert.Equal(t, tt.expected, result, "Expected keys to be correctly parsed and validated")
			}
		})
```

sasha-bitfly (approved):

PR #1348, opened 2025-02-24
refactor: breakup, improve functionality and add tests for `exporter/modules/base.go` functions

Monika-Bitfly:
Changes:
- refactored `pkg/exporter/modules` files and added unit tests:
   - `base.go`
   - `genesis_deposit.go`
   - `network_liveness.go`
   - `pubkey_tags.go`
   - `relays.go`
   - `ssv.go`
   - `sync_committees_count.go`
   - `sync_committees.go`


backend/pkg/monitoring/services/base.go
```diff
@@ -88,30 +88,33 @@ func NewStatusReport(id constants.Event, timeout time.Duration, check_interval t
...
-				"deployment_type": utils.Config.DeploymentType,
+				"deployment_type": deploymentType,
 				"insert_id":       flake,
 				"expires_at":      expires_at,
 				"timeouts_at":     timeouts_at,
 				"metadata":        metadata,
 			}, "sending status report")
 			var err error
+			if deploymentType == "test" {
+				return
```
Tangui-Bitfly, line 99: two things:
* this behavior was not present before, what made you make those changes ?
* `test` is an invalid state. If you look [at this](https://github.com/gobitfly/beaconchain/blob/42361b6f5d20dd865aa02f8f602c2563a0b5048c/backend/pkg/commons/utils/config.go#L113) the valid types are `development`, `development_noisy`, `staging`, `production`

Is it code to allow testing ? If that the case, it's really not a good practice to add complexity to our code for testing purpose.

backend/pkg/consapi/client.go
```diff
@@ -6,10 +6,13 @@ import (
 	"github.com/gobitfly/beaconchain/pkg/consapi/types"
 )
 
-type Client struct {
-	ClientInt
+type ClientCreator interface {
```
Tangui-Bitfly, line 9: why do we need a `ClientCreator` interface ?
This impact the syntax
```go
// before
client := consapi. NewClient(...)

// now
creator := &consapi.DefaultClientCreator{}
client := creator.NewClient(...)
```
I guess you did that to make it easier to test, but that is not the right solution. You should rather use dependency injection.
```go
// bad
func Foo() {
	creator := &consapi.DefaultClientCreator{}
	client := creator.NewClient(...)
	...
}

// good
func main() {
	client := consapi. NewClient(...)
	Foo(client)
}

func Foo(client consapi.Client) { // where consapi.Client is an interface
	...
}
```

backend/pkg/commons/cache/redis_cache.go
```diff
@@ -13,7 +13,7 @@ import (
 )
 
 type RedisCache struct {
-	redisRemoteCache *redis.Client
+	RedisRemoteCache *redis.Client
```
Tangui-Bitfly, line 16: Why exporting this field ? Usually, exporting a field is a smell that the code lack of proper abstraction.
My guess here is that you access `RedisCache.RedisRemoteCache` in a test outside of the package `pkg/commons/cache`.
It would be better to rely on mocking / faking / stubbing `RemoteCache`

backend/pkg/commons/cache/tiered_cache.go
```diff
@@ -14,8 +14,8 @@ import (
 
 // Tiered cache is a cache implementation combining a
 type TieredCacheBase struct {
-	localGoCache *freecache.Cache
-	remoteCache  RemoteCache
+	LocalGoCache *freecache.Cache
+	RemoteCache  RemoteCache
```
Tangui-Bitfly, line 18: Same thing.
I guess you choose this solution because we have a global variable `var TieredCache *TieredCacheBase`, which is not in any way a good practice, and this make thing more testable.
But this is like adding an other bandaid on a broken pipe, now the code is little bite more complex than before.
Again the good solution is to rely on dependency injection
```go
// the best solution
func main() {
	tiredCache := NewTiredCache(...)
	foo := Foo{cache: tiredCache}
}

type Foo struct {
	...
	cache TiredCacher
}

type TiredCacher interface {
	SetString(key, value string, expiration time.Duration) error
	....
}

type TiredCache {
	LocalGoCache *freecache.Cache
	RemoteCache  RemoteCache
}
```
Or rework this global variable with an interface and a default value that actually work (even a noop implementation).

```go
// intermediary solution
type TiredCacher interface {
	SetString(key, value string, expiration time.Duration) error
	....
}

var TieredCache TiredCacher = noopCacher{}

type noopCacher struct {}

func (noopCacher) SetString(key, value string, expiration time.Duration) error {
	return nil
}

```

backend/pkg/commons/rpc/lighthouse.go
```diff
@@ -33,7 +33,7 @@ var LighthouseLatestHeadEpoch uint64 = 0
 
 // LighthouseClient holds the Lighthouse client info
 type LighthouseClient struct {
-	cl                  *consapi.NodeClient
+	CL                  *consapi.NodeClient
```
Tangui-Bitfly, line 36: same this is not the good solution
Either use an interface for LighthouseClient (maybe `rpc.Client` is already doing the trick) or use `NewLighthouseClient` to initialize your object while testing.

backend/pkg/commons/types/exporter.go
```diff
@@ -575,6 +575,11 @@ type WeiString struct {
 	pgtype.Numeric
 }
 
+func (b *WeiString) Set(s string) error {
+	err := b.Numeric.Set(s)
```
Tangui-Bitfly: This function doesn't bring anything is just a wrapper around `Numeric.Set(s)` which `WeiString` already extends.

backend/pkg/exporter/modules/base_test.go
```diff
@@ -0,0 +1,471 @@
...
+			event: &constypes.EventResponse{
+				Event: constypes.EventHead,
+			},
+			expectedError: true,
+		},
+	}
+
+	utils.Config = &types.Config{
+		DeploymentType: "test",
+		Indexer: types.IndexerConfig{
```
Tangui-Bitfly, line 199: I don't think those lines are required

backend/pkg/exporter/modules/base_test.go
```diff
@@ -0,0 +1,471 @@
...
+				DeploymentType: "test",
+			}
+
+			err := initializeModules(tt.modules)
+			if tt.expectedError {
+				if err == nil {
+					t.Error("expected error, got nil")
+				}
+			}
+			if !tt.expectedError {
```
Tangui-Bitfly, line 56: This is a bite hard to read
```go
if tt.expectedError {
	if err == nil {
		t.Errorf("...")
	}
}
if !tt.expectedError {
	if err != nil {
		t.Errorf("...")
	}
}
```
Try to simplify that by using
```go
if err != nil {
	if tt.expectedError  {
		return
	}
	t.Fatalf("...")
}
if tt.expectedError {
	t.Fatalf("expected an error")
}
```

backend/pkg/exporter/modules/base_test.go
```diff
@@ -0,0 +1,471 @@
...
+			name:          "empty modules list",
+			modules:       []ModuleInterface{},
+			expectedError: true,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			utils.Config = &types.Config{
+				DeploymentType: "test",
```
Tangui-Bitfly, line 47: Yes that's not a good practice IMO

backend/pkg/exporter/modules/genesis_deposit.go
```diff
@@ -47,76 +28,60 @@ func genesisDepositsExporter(client rpc.Client) {
...
-			log.Error(err, "error hydrating eth1 data into genesis-deposits", 0)
-			time.Sleep(time.Minute)
-			continue
-		}
+func isBeaconChainStarted() bool {
+	latestEpoch, err := db.GetLatestEpoch()
+	if err != nil {
+		log.Error(err, "error retrieving latest epoch from the database", 0)
+		time.Sleep(time.Second * 10)
+		return false
```
Tangui-Bitfly, line 48: There is an issue with the sleep here.
We will wait 10 second, return false, and wait 1 minute. Previous behavior was to only wait 10s

backend/pkg/exporter/modules/genesis_deposit_test.go
```diff
@@ -0,0 +1,283 @@
...
+	tests := []struct {
+		name           string
+		mockRows       *sqlmock.Rows
+		mockError      error
+		expectedResult bool
+	}{
+		{
+			name:           "beacon chain started",
+			mockRows:       sqlmock.NewRows([]string{"epoch"}).AddRow(10),
+			mockError:      nil,
```
Tangui-Bitfly, line 23: It's a general comment, but I prefer to remove default value from my test definition.
This way it's easier to pin point with which value we are testing
```go
		{
			name:           "beacon chain started",
			mockRows:       sqlmock.NewRows([]string{"epoch"}).AddRow(10),
			expectedResult: true,
		},
		{
			name:           "beacon chain not started",
			mockRows:       sqlmock.NewRows([]string{"epoch"}).AddRow(0),
		},
		{
			name:           "database error",
			mockError:      errors.New("database error"),
		},
```

backend/pkg/exporter/modules/genesis_deposit_test.go
```diff
@@ -0,0 +1,283 @@
...
+			}
+			defer dbMock.Close()
+
+			sqlxDB := sqlx.NewDb(dbMock, "sqlmock")
+			db.WriterDb = sqlxDB
+
+			// mock GetLatestEpoch query
+			query := `SELECT COALESCE\(MAX\(epoch\), 0\) FROM epochs`
+			if tt.mockError != nil {
+				mock.ExpectQuery(query).WillReturnError(tt.mockError)
```
Tangui-Bitfly, line 54: This summarizes one of my issues with mocks.
Mocks are so strong that it encourages linking dependency in higher level of the code.
Yes you test is working, but relying on mocking sql query should be a smell for bad abstraction.
Let say in the future we change the query in `commons/db`, this test will fail even so the db "api" did not change.
Here we should not use `db.GetLatestEpoch()`, we should have an interface like
```
type ConsensusDB interface {
	GetLatestEpoch() (uint64, error)
}
```
This is easier to mock because it remove the need of sql queries (which are implementation details that package `exporter/modules` should not be aware).

Tangui-Bitfly: This comment also applies to other mocked queries in this PR.

backend/pkg/exporter/modules/genesis_deposit_test.go
```diff
@@ -0,0 +1,283 @@
...
+			result := areGenesisDepositsExported()
+
+			if result != tt.expectedResult {
+				t.Errorf("expected result: %v, got: %v", tt.expectedResult, result)
+			}
+		})
+	}
+}
+
+func TestExportGenesisDeposits(t *testing.T) {
```
Tangui-Bitfly, line 117: Too many if else, mocked queries, lack of clear final test, ... This test is too complex.
I bet than the next time this test turn red we either fully refactor it or remove it.

backend/pkg/exporter/modules/module_log.go
```diff
@@ -0,0 +1,59 @@
+package modules
+
+import (
+	"fmt"
+
+	"github.com/gobitfly/beaconchain/pkg/commons/log"
+)
+
+type ModuleLog struct {
```
Tangui-Bitfly, line 9: I know this is not your code, but I just want to point out that this code is one of the reason our `log` package is bad.
The proper way to do that with logrus (or other logging lib) would be 
```go
type MyModule struct {
	logger logrus.FieldLogger
}

func NewMyModule() MyModule {
	return MyModule{
		logger: logrus.NewEntry(logrus.New()).WithFields(logrus.Fields{
			"module": "MyModule",
		})
	}
}
```

backend/pkg/exporter/modules/network_liveness.go
```diff
@@ -23,55 +23,73 @@ func networkLivenessUpdater(client rpc.Client) {
...
+}
+
+func createNetworkStatusReport(slotDuration time.Duration) func(status constants.StatusType, metadata map[string]string) {
+	return services.NewStatusReport(constants.Event_ExporterLegacyNetworkLiveness, constants.Default, slotDuration, utils.Config.DeploymentType)
+}
 
-		time.Sleep(slotDuration)
+func handleNetworkSuccess(slotDuration time.Duration, statusReport func(status constants.StatusType, metadata map[string]string)) {
+	statusReport(constants.Success, nil)
+	time.Sleep(slotDuration)
```
Tangui-Bitfly, line 72: do we really need a function for those two lines ?

backend/pkg/exporter/modules/network_liveness.go
```diff
@@ -23,55 +23,73 @@ func networkLivenessUpdater(client rpc.Client) {
...
+}
+
+func getPreviousHeadEpoch() (uint64, error) {
+	prevHeadEpoch, err := db.GetNetworkLivenessPreviousHeadEpoch()
+	if err != nil {
+		return 0, err
+	}
+	return prevHeadEpoch, nil
```
Tangui-Bitfly, line 80: what the point of this function ? it's only a wrapper around `db.GetNetworkLivenessPreviousHeadEpoch()`
If we do that every time we basically double our code lines and more code lines means more complexity, more code to maintain, ...

backend/pkg/exporter/modules/network_liveness_test.go
```diff
@@ -0,0 +1,261 @@
...
+	"github.com/gobitfly/beaconchain/pkg/commons/cache"
+	"github.com/gobitfly/beaconchain/pkg/commons/db"
+	"github.com/gobitfly/beaconchain/pkg/commons/types"
+	"github.com/gobitfly/beaconchain/pkg/commons/utils"
+	"github.com/gobitfly/beaconchain/pkg/monitoring/constants"
+	"github.com/jmoiron/sqlx"
+	"github.com/pkg/errors"
+)
+
+func TestCreateNetworkStatusReport(t *testing.T) {
```
Tangui-Bitfly, line 19: do we really need a test for a that simple one line function ?

Tangui-Bitfly, line 19: My point is we have some tests for little function but we don't have a test for the main function `networkLivenessUpdater` which seems to be the core business logic of that file.
This means we are testing implementation details and not the real behavior we expect to happen in production.
This will create lot of tests that will change after each modification in implementation vs a shorter amount of tests that change only when the whole behavior change.

backend/pkg/exporter/modules/network_liveness_test.go
```diff
@@ -0,0 +1,261 @@
...
+			name:         "valid network success",
+			slotDuration: time.Second * 12,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			statusReport := func(status constants.StatusType, metadata map[string]string) {}
+			handleNetworkSuccess(tt.slotDuration, statusReport)
+		})
```
Tangui-Bitfly, line 59: what are we testing here ? There is no check. This is just testing for raising the coverage, which is not good.

backend/pkg/exporter/modules/network_liveness_test.go
```diff
@@ -0,0 +1,261 @@
...
+			if !tt.expectedError {
+				if err != nil {
+					t.Errorf("unexpected error: %v", err)
+				}
+
+				if prevHeadEpoch != tt.expectedHeadEpoch {
+					t.Errorf("expected head epoch: %v, got: %v", tt.expectedHeadEpoch, prevHeadEpoch)
+				}
+			}
+			if tt.expectedError {
```
Tangui-Bitfly, line 116: the order is confusing, check first for error and then for success

backend/pkg/exporter/modules/network_liveness_test.go
```diff
@@ -0,0 +1,261 @@
...
+			}
+
+			prevHeadEpoch, err := getPreviousHeadEpoch()
+
+			if !tt.expectedError {
+				if err != nil {
+					t.Errorf("unexpected error: %v", err)
+				}
+
+				if prevHeadEpoch != tt.expectedHeadEpoch {
```
Tangui-Bitfly, line 112: This should not be indented, this is the core check of test, by being indented it loose its importance for the reader.
This is related to one of my previous comment with how to handle expected error

backend/pkg/exporter/modules/pubkey_tags.go
```diff
@@ -1,51 +1,58 @@
...
-		}
-		_ = tx.Rollback()
+		time.Sleep(constants.Duration10Mins)
+	}
+}
 
-		log.Infof("Updating Pubkey Tags took %v sec.", time.Since(start).Seconds())
-		r(constants.Success, map[string]string{"took": time.Since(start).String(), "took_raw": time.Since(start).String()})
-		metrics.TaskDuration.WithLabelValues("validator_pubkey_tag_updater").Observe(time.Since(start).Seconds())
+func updatePubkeyTagOnce() error {
```
Tangui-Bitfly, line 26: Why suffic `Once` ?

backend/pkg/exporter/modules/relays.go
```diff
@@ -13,237 +12,230 @@ import (
...
-		if shouldLogExportAsError(r) {
-			log.Error(err, "", 0, map[string]interface{}{"relay": r.ID})
-		} else {
-			log.WarnWithFields(log.Fields{"relay": r.ID}, errMsg.Error())
+func exportRelaysConcurrently(relays []types.Relay) {
+	var wg sync.WaitGroup
+	var mux sync.Mutex
+
+	for _, relay := range relays {
+		if shouldTryToExportRelay(relay) {
```
Tangui-Bitfly, line 38:
```go
if !shouldTryToExportRelay(relay) {
	continue
}
wg.Add(1)
go func(r types.Relay) {
	// create relay logger
	defer wg.Done()
	singleRelayExport(r, &mux)
}(relay)
```

backend/pkg/exporter/modules/relays.go
```diff
@@ -13,237 +12,230 @@ import (
...
-	err := exportRelayBlocks(r)
-	if err != nil {
-		errMsg := fmt.Errorf("failed to export blocks for relay: %v", err)
-		if shouldLogExportAsError(r) {
-			log.Error(err, "", 0, map[string]interface{}{"relay": r.ID})
-		} else {
-			log.WarnWithFields(log.Fields{"relay": r.ID}, errMsg.Error())
+func exportRelaysConcurrently(relays []types.Relay) {
+	var wg sync.WaitGroup
+	var mux sync.Mutex
```
Tangui-Bitfly, line 35: do we really need the mutex ?

Tangui-Bitfly: Well we don't have unit test for that function, which seems odd

backend/pkg/exporter/modules/relays.go
```diff
@@ -13,237 +12,230 @@ import (
...
-		return nil, fmt.Errorf("error retrieving delivered payloads for relay: %v, offset: %v, url: %v: %w", r.ID, offset, url, err)
+		log.Error(err, "could not update successful relay export", 0, map[string]interface{}{"relay": r.ID})
+		return err
 	}
 
-	defer resp.Body.Close()
+	return nil
+}
+
+func updateRelayExportFailure(r types.Relay, mux *sync.Mutex) error {
```
Tangui-Bitfly, line 92: Honestly the flow start be very complicated with all those `hanldle` func

backend/pkg/exporter/modules/relays.go
```diff
@@ -13,237 +12,230 @@ import (
...
 	}
-	if firstUsage.BlockSlot == 0 {
-		return nil
-	}
-	err = retrieveAndInsertPayloadsFromRelay(r, 0, firstUsage.BlockSlot)
-	if err != nil {
-		log.Error(err, "failed to retrieve and insert possibly missing payloads", 0, map[string]interface{}{"relay": r.ID})
-		return err
+
+	if firstUsage.BlockSlot != 0 {
```
Tangui-Bitfly, line 132: previous code was better, keep indentation to the minimum
Have a look at the [return early pattern](https://medium.com/swlh/return-early-pattern-3d18a41bba8)

backend/pkg/exporter/modules/relays.go
```diff
@@ -13,237 +12,230 @@ import (
...
+	resp, err := client.Get(url)
+	if err != nil {
+		return nil, fmt.Errorf("failed to fetch delivered payloads from relay %v: %w", r.ID, err)
+	}
+	defer resp.Body.Close()
+
+	return decodePayloads(resp, r.ID)
+}
+
+func decodePayloads(resp *http.Response, tagID string) ([]exportertypes.BidTrace, error) {
```
Tangui-Bitfly, line 232: This function is just a wrapper around json decode.
It should not exist.

backend/pkg/commons/db/db.go
```diff
@@ -2492,6 +2649,236 @@ func CacheQuery(query string, viewName string, indexes ...[]string) error {
...
+	ON CONFLICT DO NOTHING`, tagID, slot, blockHash)
+
+	if err != nil {
+		return err
+	}
+
+	return tx.Commit()
+}
+
+func SaveBlocksRelays(tagID string, slot uint64, payloadValue types.WeiString, blockHash, builderPubkey, proposerPubkey, proposerFeeRecipient []byte) error {
```
Tangui-Bitfly, line 2824: we should accept a `exportertypes.BidTrace` rather than all those params

backend/pkg/exporter/modules/relays_test.go
```diff
@@ -0,0 +1,544 @@
...
+			if tt.expectedError {
+				if err == nil {
+					t.Errorf("expected error, got nil")
+				}
+			} else {
+				if err != nil {
+					t.Errorf("unexpected error: %v", err)
+				}
+			}
+		})
```
Tangui-Bitfly, line 83: Again I don't know what are we really testing here. That we call the db with the right query ? Shouldn't it be tested elsewhere ?

Tangui-Bitfly: This comment applies for other the tests in that file

backend/pkg/exporter/modules/relays_test.go
```diff
@@ -0,0 +1,544 @@
...
+			if tt.expectedError {
+				if err == nil {
+					t.Errorf("expected error, got nil")
+				}
+			}
+		})
+	}
+}
+
+func TestDecodePayloads(t *testing.T) {
```
Tangui-Bitfly, line 356: IMO test should be on `fetchDeliveredPayloads` and not `decodePayloads`

backend/pkg/exporter/modules/relays.go
```diff
@@ -13,237 +12,230 @@ import (
...
+		utils.MustParseHex(payload.ProposerFeeRecipient))
+	if err != nil {
+		return fmt.Errorf("failed to insert payload into relays_blocks table: %w", err)
 	}
-	return tx.Commit()
+
+	return nil
+}
+
+func fetchDeliveredPayloads(r types.Relay, offset uint64) ([]exportertypes.BidTrace, error) {
```
Tangui-Bitfly, line 216: You could introduce an interface here
```go
type RelayClient interface {
	DeliveredPayloads(endpoint string, id string, offset uint64) ([]exportertypes.BidTrace, error)
}

type relayClient struct {}

func (relayClient) DeliveredPayloads(endpoint string, id string, offset uint64) ([]exportertypes.BidTrace, error) {
	....
}
```
This would make testing higher functions easier.

backend/pkg/exporter/modules/relays_test.go
```diff
@@ -0,0 +1,544 @@
...
+			}
+		})
+	}
+}
+
+func TestShouldTryToExportRelay(t *testing.T) {
+	tests := []struct {
+		name           string
+		relay          types.Relay
+		expectedResult bool
```
Tangui-Bitfly, line 422: `expected` is shorter and bring the same amount of information

backend/pkg/exporter/modules/relays_test.go
```diff
@@ -0,0 +1,544 @@
...
+	}
+}
+
+func toWeiString(value string) types.WeiString {
+	weiValue := &types.WeiString{}
+	err := weiValue.Set(value)
+	if err != nil {
+		panic(err)
+	}
+	return *weiValue
```
Tangui-Bitfly, line 543: Remove `&` at line 538 and `*` here.

backend/pkg/exporter/modules/ssv.go
```diff
@@ -94,73 +58,116 @@ func exportSSV() error {
 	}
 }
 
-func saveSSV(res *SSVExporterResponse) error {
-	tx, err := db.WriterDb.Beginx()
+func connectToWebSocket() (*websocket.Conn, *http.Response, error) {
```
Tangui-Bitfly, line 60: I don't see the added value of this function

backend/pkg/exporter/modules/ssv.go
```diff
@@ -94,73 +58,116 @@ func exportSSV() error {
...
-		time.Sleep(time.Millisecond * 100)
+	}
+}
+
+func readWebSocketMessage(conn *websocket.Conn) ([]byte, error) {
+	_, message, err := conn.ReadMessage()
+	return message, err
+}
+
+func unmarshalSSVResponse(message []byte) (*types.SSVExporterResponse, error) {
```
Tangui-Bitfly, line 96: function brings no value

backend/pkg/exporter/modules/ssv.go
```diff
@@ -94,73 +58,116 @@ func exportSSV() error {
...
+
+		if err := processSSVResponse(res); err != nil {
+			log.Error(err, "error processing ssv validators", 0)
+			continue
 		}
-		time.Sleep(time.Millisecond * 100)
+	}
+}
+
+func readWebSocketMessage(conn *websocket.Conn) ([]byte, error) {
```
Tangui-Bitfly, line 91: function brings no value

backend/pkg/exporter/modules/ssv_test.go
```diff
@@ -0,0 +1,354 @@
...
+	}
+	for i := range a.Data {
+		if a.Data[i].Publickey != b.Data[i].Publickey {
+			return false
+		}
+	}
+	return true
+}
+
+func equal(a, b []byte) bool {
```
Tangui-Bitfly, line 344: remove that function, you can use `bytes.Equal`

backend/pkg/exporter/modules/sync_committees.go
```diff
@@ -1,150 +1,176 @@
...
+func createStatusReport() func(status constants.StatusType, metadata map[string]string) {
+	return services.NewStatusReport(constants.Event_ExporterLegacySyncCommittees, constants.Default, time.Second*12, utils.Config.DeploymentType)
+}
+
+func handleExportError(err error, startTime time.Time, statusReport func(status constants.StatusType, metadata map[string]string)) {
+	log.Error(err, "error exporting sync_committees", 0, map[string]interface{}{"duration": time.Since(startTime)})
+	statusReport(constants.Failure, map[string]string{"error": err.Error()})
+}
+
+func handleExportSuccess(startTime time.Time, statusReport func(status constants.StatusType, metadata map[string]string)) {
```
Tangui-Bitfly, line 43: those three functions bring no value

backend/pkg/exporter/modules/sync_committees.go
```diff
@@ -1,150 +1,176 @@
...
+			if err := ExportSyncCommitteeData(rpcClient, period); err != nil {
+				return fmt.Errorf("error exporting sync-committee at period %v: %w", period, err)
+			}
+		}
+	}
+
+	return nil
+}
+
+func createDBPeriodsMap(dbPeriods []uint64) map[uint64]bool {
```
Tangui-Bitfly, line 71: function bring no value

backend/pkg/exporter/modules/sync_committees.go
```diff
@@ -1,150 +1,176 @@
...
+	dbPeriodsMap := createDBPeriodsMap(dbPeriods)
+	firstPeriod, lastPeriod := calculateSyncPeriodRange()
+
+	for period := firstPeriod; period <= lastPeriod; period++ {
+		_, exists := dbPeriodsMap[period]
+		if !exists {
+			if err := ExportSyncCommitteeData(rpcClient, period); err != nil {
+				return fmt.Errorf("error exporting sync-committee at period %v: %w", period, err)
+			}
+		}
```
Tangui-Bitfly, line 65:
```go
if _, exists := dbPeriodsMap[period]; exists {
	continue
}
if err := ExportSyncCommitteeData(rpcClient, period); err != nil {
	return fmt.Errorf("error exporting sync-committee at period %v: %w", period, err)
}
```

backend/cmd/misc/main.go
```diff
@@ -2083,12 +2084,12 @@ func reExportSyncCommittee(rpcClient rpc.Client, p uint64, dryRun bool) error {
...
 			}
 		}()
-		_, err = tx.Exec(`DELETE FROM sync_committees WHERE period = $1`, p)
+		_, err = tx.Exec(`DELETE FROM sync_committees WHERE period = $1`, period)
 		if err != nil {
 			return errors.Wrap(err, "delete old entries")
 		}
 
-		err = modules.ExportSyncCommitteeAtPeriod(rpcClient, p, tx)
+		err = modules.ExportSyncCommitteeData(rpcClient, period)
```
Tangui-Bitfly, line 2092: we loose tx param here, is that intended ?

backend/pkg/exporter/modules/sync_committees.go
```diff
@@ -1,150 +1,176 @@
...
-				"epoch":    utils.FirstEpochOfSyncPeriod(p),
-				"duration": time.Since(t0),
-			}, "exported sync_committee")
-		}
-	}
-	return nil
+	return firstPeriod, lastPeriod
 }
 
-func ExportSyncCommitteeAtPeriod(rpcClient rpc.Client, p uint64, providedTx *sqlx.Tx) error {
```
Tangui-Bitfly, line 70: `providedTx` is not there anymore, is that intended ?

backend/pkg/exporter/modules/sync_committees_count_test.go
```diff
@@ -0,0 +1,447 @@
...
+	tests := []struct {
+		name          string
+		mockRows      *sqlmock.Rows
+		mockError     error
+		mockRows2     *sqlmock.Rows
+		mockError2    error
+		mockRows3     *sqlmock.Rows
+		mockError3    error
+		mockRows4     *sqlmock.Rows
+		mockError4    error
```
Tangui-Bitfly, line 24: This is unmaintainable. What is difference between mockRows2 and mockRows4 ? I can bet that in 6 months no one will know without spending 10min reading the test

backend/pkg/exporter/modules/sync_committees.go
```diff
@@ -1,150 +1,176 @@
...
-func GetSyncCommitteAtPeriod(rpcClient rpc.Client, p uint64) ([]SyncCommittee, error) {
-	stateID := uint64(0)
-	if p > 0 {
-		stateID = utils.FirstEpochOfSyncPeriod(p-1) * utils.Config.Chain.ClConfig.SlotsPerEpoch
-	}
-	epoch := utils.FirstEpochOfSyncPeriod(p)
-	if stateID/utils.Config.Chain.ClConfig.SlotsPerEpoch <= utils.Config.Chain.ClConfig.AltairForkEpoch {
-		stateID = utils.Config.Chain.ClConfig.AltairForkEpoch * utils.Config.Chain.ClConfig.SlotsPerEpoch
-		epoch = utils.Config.Chain.ClConfig.AltairForkEpoch
+func parseSyncArgsAndIDs(data []SyncCommittee) ([]interface{}, []string) {
```
Tangui-Bitfly, line 114: Can we forbid ourself from using `interface{}` when possible, please. I know you didn't write that piece of code but this is not acceptable in a typed language.
Here, it could be replaced by `[]uint64`.

Tangui-Bitfly: The real issue here is that complicated conversion from struct to array should not be done here. It should happen in `db.SaveSyncCommitteeData`. This function should accept a `[]SyncCommittee` and then do the conversion for the sql query.

backend/pkg/exporter/modules/sync_committees_test.go
```diff
@@ -0,0 +1,298 @@
...
+					Period:         1,
+					ValidatorIndex: 2,
+					CommitteeIndex: 3,
+				},
+			},
+			expectedArgs: []interface{}{
+				uint64(1),
+				uint64(2),
+				uint64(3),
+			},
```
Tangui-Bitfly, line 222: this should not be like that

backend/pkg/exporter/modules/sync_committees_test.go
```diff
@@ -0,0 +1,298 @@
...
+			expectedIDs: []string{"($1,$2,$3)"},
```
Tangui-Bitfly, line 223: this is sql testing, it should not be here and should not take this form. Here you're testing the sql go what we are sending to the sql go lib and not if we are successfully writing something to a sql db

backend/pkg/exporter/modules/sync_committees_test.go
```diff
@@ -0,0 +1,298 @@
...
+	}
+	for i := range a {
+		if a[i] != b[i] {
+			return false
+		}
+	}
+	return true
+}
+
+func compareSyncCommittees(a, b []SyncCommittee) bool {
```
Tangui-Bitfly, line 288: those compare function could be replaced by [slices.Equal](https://pkg.go.dev/slices#Equal)

backend/pkg/exporter/types/bid.go
```diff
@@ -0,0 +1,15 @@
+package types
```
Tangui-Bitfly, line 1: Do we really want to introduce a new `types` package ? This an anti pattern in go that we already brake.
What are we winning by moving those types definition here ? We could simply put them in a `modules/types.go`

backend/pkg/exporter/types/ssv.go
```diff
@@ -0,0 +1,19 @@
+package types
```
Tangui-Bitfly, line 1: same

backend/pkg/monitoring/constants/main.go
```diff
@@ -13,6 +13,7 @@ const (
 	Success                                 StatusType    = "success"
 	Failure                                 StatusType    = "failure"
 	Default                                 time.Duration = -1 * time.Second
+	Duration10Mins                          time.Duration = 10 * time.Minute
```
Tangui-Bitfly: I don't think it should be in a constant

Tangui-Bitfly (changes requested):
## Review
I'm concerned by the way we are doing test.

Don't reflect your internal code structure within your unit tests. Test for observable behaviour instead. Think about
_if I enter values x and y, will the result be z?_
instead of
_if I enter x and y, will the method call class A first, then call class B and then return the result of class A plus the result of class B?_

In a lot of case we have tests for the lower functions but nothing for the higher level function where the magic happens.
I fear that our tests:
- provide a false sense of security, because in reality we are not testing anything
- lock ourself by protecting sub optimal code, refactoring the implementation will break A LOT of tests. So much that it will discourage developers from doing any refactoring
- explode the total number of line of code in our repo, this is not really a good metric but IMO more code = more complexity = more cognitive load = less developers moral = general lower code quality

## Sidenote
Your tests are taking [too long to execute](https://github.com/gobitfly/beaconchain/actions/runs/13518334518/job/37771720633#step:4:85)
```
ok  	github.com/gobitfly/beaconchain/pkg/exporter/modules	39.318s	coverage: 7.1% of statements
```
It should not be over, let say, 5s (and it's very generous)

PR #1349, opened 2025-02-25
change fraction digits of crypto currencies to 6 (18)

marcel-bitfly:
See: BEDS-1275

### DIFF
--- a/frontend/components/block/table/BlockTableRewardItem.vue
+++ b/frontend/components/block/table/BlockTableRewardItem.vue
@@ -11,6 +11,7 @@ defineProps<{
   <BcTooltip
     v-if="status === 'success' && reward"
     class="combine-rewards"
+    tooltip-text-align="left"
     fit-content
   >
     <BcFormatAmount
@@ -27,8 +28,10 @@ defineProps<{
     <span v-else>{{ $t("dashboard.validator.blocks.cl_pending") }}</span>
     <template #tooltip>
       <div>
-        <div class="tt-row">
-          <span>{{ $t("dashboard.validator.blocks.el_rewards") }}: </span>
+        <div class="tooltip-row">
+          <h3 class="tooltip-title">
+            {{ $t("dashboard.validator.blocks.el_rewards") }}
+          </h3>
           <BcFormatAmount
             :value="reward?.el"
             source-currency="elCurrency"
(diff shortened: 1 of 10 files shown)


frontend/components/block/table/BlockTableRewardItem.vue
```diff
@@ -27,8 +28,10 @@ defineProps<{
     <span v-else>{{ $t("dashboard.validator.blocks.cl_pending") }}</span>
     <template #tooltip>
       <div>
-        <div class="tt-row">
-          <span>{{ $t("dashboard.validator.blocks.el_rewards") }}: </span>
+        <div class="tooltip-row">
+          <h3 class="tooltip-titel">
```
enzo-bitfly: `tooltip-title` in english

frontend/components/block/table/BlockTableRewardItem.vue
```diff
@@ -37,8 +40,10 @@ defineProps<{
             has-higher-precision
           />
         </div>
-        <div class="tt-row">
-          <span>{{ $t("dashboard.validator.blocks.cl_rewards") }}: </span>
+        <div class="tooltip-row">
+          <h3 class="tooltip-titel">
```
enzo-bitfly: same

frontend/components/block/table/BlockTableRewardItem.vue
```diff
@@ -58,11 +63,11 @@ defineProps<{
...
-.tt-row {
-  display: flex;
-  flex-wrap: nowrap;
-  white-space: nowrap;
-  gap: 3px;
+.tooltip-titel {
+  text-align: center;
+}
+.tooltip-row:not(:first-child) {
+  margin-top: var(--padding-small);
```
enzo-bitfly, line 70: suggestion: using `margin-bottom` as a general rule is better to go with the block flow of the document (pushing from top to bottom). You could probably achieve the same thing adding `margin-bottom` to `not(:last-child)`

enzo-bitfly (commented):
Great that you added those `h3`s. Nice job 👍

frontend/components/block/table/BlockTableRewardItem.vue
marcel-bitfly: nope we always use `margin-top`,

e.g. see: https://matthewjamestaylor.com/css-margin-top-vs-bottom

enzo-bitfly: I don't agree with this. That would be choosing `margin-top` based on an edge case that can be solved easily with a negative `margin-top` on the 'ad' element, which is the one causing the disruption. 
For a predictable consistent spacing I still find `margin-bottom` a better solution.

PR #1354, opened 2025-02-26
refactor: revise and add tests for `exporter/modules/genesis_deposit.go`

Monika-Bitfly:


backend/pkg/exporter/modules/genesis_deposit_test.go
```diff
@@ -0,0 +1,183 @@
...
+
+import (
+	"testing"
+
+	"github.com/gobitfly/beaconchain/pkg/commons/db/mocks"
+	"github.com/gobitfly/beaconchain/pkg/consapi/types"
+	"github.com/pkg/errors"
+)
+
+func TestGetLatestEpoch(t *testing.T) {
```
Tangui-Bitfly: What real testing is done in that test ?
We only test the mocked call here, the func `getLatestEpoch` is just a wrapper around `db.GetLatestEpoch`.
So we are testing if when we call our mock we got our mock data, this is telling nothing about our code.

backend/pkg/exporter/modules/genesis_deposit.go
```diff
@@ -47,76 +38,60 @@ func genesisDepositsExporter(client rpc.Client) {
...
-		if err != nil {
-			err := tx.Rollback()
-			if err != nil && !errors.Is(err, sql.ErrTxDone) {
-				log.Error(err, "error rolling back transaction", 0)
-			}
-			log.Error(err, "error hydrating eth1 data into genesis-deposits", 0)
-			time.Sleep(time.Minute)
-			continue
-		}
+func getLatestEpoch(db db.ConsensusDBI) (uint64, error) {
```
Tangui-Bitfly: what's the point of that func ?
It's a wrapper without any additional value than the original.
And by hiding the sleep from the "main" func it obfuscate the code.

backend/pkg/exporter/modules/genesis_deposit.go
```diff
@@ -47,76 +38,60 @@ func genesisDepositsExporter(client rpc.Client) {
...
-			if err != nil && !errors.Is(err, sql.ErrTxDone) {
-				log.Error(err, "error rolling back transaction", 0)
-			}
-			log.Error(err, "error updating deposit count for the genesis slot", 0)
-			time.Sleep(time.Minute)
-			continue
-		}
+	return latestEpoch, nil
+}
+func getGenesisDepositCount(db db.ConsensusDBI) (uint64, error) {
```
Tangui-Bitfly: same

backend/pkg/exporter/modules/genesis_deposit.go
```diff
@@ -47,76 +38,60 @@ func genesisDepositsExporter(client rpc.Client) {
...
-				time.Sleep(time.Minute)
-				continue
-			}
-		}
+		log.Infof("exported genesis-deposits for %v genesis-validators", len(genesisValidators.Data))
+		return
+	}
+}
 
-		// hydrate the eth1 deposit signature for all genesis validators that have a corresponding eth1 deposit
```
Tangui-Bitfly, line 77: we lost that comment, it should be in `func (d *ConsensusDB) UpdateBlockDepositsSignature() error`

backend/pkg/exporter/modules/genesis_deposit_test.go
```diff
@@ -0,0 +1,183 @@
...
+				t.Error("expected error, got nil")
+			}
+			if result != tt.expectedResult {
+				t.Errorf("expected result: %v, got: %v", tt.expectedResult, result)
+			}
+		})
+	}
+}
+
+func TestGetGenesisDepositCount(t *testing.T) {
```
Tangui-Bitfly: same as TestGetLatestEpoch

backend/pkg/exporter/modules/genesis_deposit_test.go
```diff
@@ -0,0 +1,183 @@
...
+				t.Error("expected error, got nil")
+			}
+			if result != tt.expectedResult {
+				t.Errorf("expected result: %v, got: %v", tt.expectedResult, result)
+			}
+		})
+	}
+}
+
+func TestExportGenesisDeposits(t *testing.T) {
```
Tangui-Bitfly: again I don't see any business logic testing here just mock calling verification.

backend/pkg/commons/db/db.go
```diff
@@ -64,6 +64,19 @@ const DefaultInfScrollRows = 25
 
 var ErrNoStats = errors.New("no stats available")
 
+type ConsensusDB struct {
+	WriterDb *sqlx.DB
+	ReaderDb *sqlx.DB
```
Tangui-Bitfly: I don't think this field is used.

Tangui-Bitfly (changes requested):
Your tests are taking [too long to execute](https://github.com/gobitfly/beaconchain/actions/runs/13544664464/job/37853297528#step:4:86)
```
ok  	github.com/gobitfly/beaconchain/pkg/exporter/modules	71.193s	coverage: 0.7% of statements
```
It should not be over, let say, 5s (and it's very generous)

backend/pkg/exporter/modules/genesis_deposit.go
Monika-Bitfly: The goal of this change is to make the code more modular and improve readability and make testing easier since  the logic is more isolated

backend/pkg/exporter/modules/genesis_deposit_test.go
```diff
@@ -0,0 +1,81 @@
...
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			mockRPCClient := new(rpcmocks.Client)
+			mockRPCClient.On("GetValidatorState", uint64(0)).Return(genesisValidators, tt.mockValidatorStateError)
+
+			mockConsDBClient := new(mocks.ConsensusDBI)
+			mockConsDBClient.On("GetLatestEpoch").Return(tt.mockEpochResponse, nil)
+			mockConsDBClient.On("GetDepositsCountForBlockSlot").Return(tt.mockDepositCountResponse, nil)
```
Tangui-Bitfly, line 46: this mocked make an early return at [line 43](https://github.com/gobitfly/beaconchain/blob/003efcdddb9b91888fef01849c2e144443ef8496/backend/pkg/exporter/modules/genesis_deposit.go#L43).
This mean that the `mockRPCClient` is never called.

backend/pkg/exporter/modules/genesis_deposit.go
```diff
@@ -1,122 +1,107 @@
...
 
-		// check if genesis-deposits have already been exported
-		var genesisDepositsCount uint64
-		err = db.WriterDb.Get(&genesisDepositsCount, "SELECT COUNT(*) FROM blocks_deposits WHERE block_slot=0")
-		if err != nil {
-			log.Error(err, "error retrieving genesis-deposits-count when exporting genesis-deposits", 0)
-			time.Sleep(time.Minute)
-			continue
-		}
+func processGenesisDeposits(client rpc.Client, dbs db.ConsensusDBI) (bool, error) {
```
Tangui-Bitfly: I still don't get why you create a new function.
I guess it's to simplify testing with the sleep ?
It would be clearer to use a struct.

```go
type genesisDepositsExporter struct {
	client rpc.Client
	db     db.ConsensusDBI

	offset time.Duration
}

func newGenesisDepositsExporter(client rpc.Client, db db.ConsensusDBI) genesisDepositsExporter {
	return genesisDepositsExporter{
		client: client,
		db: db,
		offset: time.Minute,
	}
}

func (exporter genesisDepositsExporter) Export() {
	for {
		...
	}
}

// in backend/pkg/exporter/modules/genesis_deposit_test.go

func TestGenesisDepositsExporter(t *testing.T) {
	...
	exporter :=  genesisDepositsExporter{
		client: mockRPCClient,
		db: mockDBClient,
		offset: 0,
	}
	...
}

```

backend/pkg/exporter/modules/genesis_deposit.go
```diff
@@ -1,122 +1,107 @@
...
 	"time"
 
 	"github.com/gobitfly/beaconchain/pkg/commons/db"
 	"github.com/gobitfly/beaconchain/pkg/commons/log"
 	"github.com/gobitfly/beaconchain/pkg/commons/rpc"
+	"github.com/gobitfly/beaconchain/pkg/consapi/types"
 )
 
-func genesisDepositsExporter(client rpc.Client) {
+func genesisDepositsExporter(client rpc.Client, dbs db.ConsensusDBI) {
```
Tangui-Bitfly: It would be nice to re create an interface for the `rpc.Client`. Right now it come with a big a api (11 functions).
It could be reduced to
```go
type ValidatorClient interface {
	GetValidatorState(epoch uint64) (*constypes.StandardValidatorsResponse, error)
}
```

backend/pkg/commons/db/db.go
Tangui-Bitfly: This was not addressed

backend/pkg/exporter/modules/genesis_deposit.go
```diff
@@ -1,122 +1,121 @@
...
-		// check if the beaconchain has started
-		var latestEpoch uint64
-		err := db.WriterDb.Get(&latestEpoch, "SELECT COALESCE(MAX(epoch), 0) FROM epochs")
-		if err != nil {
-			log.Error(err, "error retrieving latest epoch from the database", 0)
-			time.Sleep(time.Second * 10)
-			continue
-		}
+type genesisDepositsExporter struct {
+	client rpc.ValidatorClient
```
Tangui-Bitfly: You don't have to rely on `rpc.ValidatorClient` this introduce coupling and enlarge the api we have access.
Locally create an interface in `genesis_deposit.go` with only one function.
The idea behind that is to allow flexibility. Let's say we want to add a logger on this `client.GetValidatorState`. We want it to be invisible for our current `genesisDepositsExporter`, so we create a wrapper around the `rpc.ValidatorClient`. This make us implements 3 functions even if we only need one.
Small interfaces are really powerful in go.

backend/pkg/exporter/modules/genesis_deposit.go
```diff
@@ -1,122 +1,121 @@
...
-			log.Error(err, "error retrieving genesis-deposits-count when exporting genesis-deposits", 0)
-			time.Sleep(time.Minute)
-			continue
-		}
+func newGenesisDepositsExporter(client rpc.Client, db db.ConsensusDBI) genesisDepositsExporter {
+	return genesisDepositsExporter{
+		client: client,
+		db:     db,
+		offset: time.Minute,
+		ctx:    context.Background(),
```
Tangui-Bitfly: ctx should be a param of `Export`
The usage would be
```go
// in main.go
ctx, cancel := context.WithCancel(context.Background())
go modules.StartAll(ctx, ...)
// wait for something
cancel()

// in backend/pkg/exporter/modules/base.go
func StartAll(ctx context.Context, ...) {
...
	go newGenesisDepositsExporter(context.ConsClient, dbs).Export(ctx)

```
This a good change but I think it's too soon.

backend/pkg/exporter/modules/genesis_deposit.go
```diff
@@ -1,122 +1,121 @@
...
+	}
+}
 
-		// if genesis-deposits have already been exported exit this go-routine
-		if genesisDepositsCount > 0 {
+func (e genesisDepositsExporter) Export() {
```
Tangui-Bitfly: This function should contains all the `log` and the `time.Sleep`.
Try to log everything in the same function (ideally `main` but the code is not in ready for it).
Same for sleep.

backend/pkg/exporter/modules/genesis_deposit.go
```diff
@@ -1,122 +1,121 @@
...
+	for {
+		select {
+		case <-e.ctx.Done():
+			log.Info("export loop cancelled", 0)
 			return
-		}
+		default:
+			latestEpoch, err := e.getLatestEpoch(e.db)
+			if err != nil {
+				continue
```
Tangui-Bitfly, line 47: log and sleep here

backend/pkg/exporter/modules/genesis_deposit.go
```diff
@@ -1,122 +1,121 @@
...
 
-		tx, err := db.WriterDb.Beginx()
-		if err != nil {
-			log.Error(err, "error beginning db-tx when exporting genesis-deposits", 0)
-			time.Sleep(time.Minute)
-			continue
-		}
+			genesisDepositCount, err := e.getGenesisDepositCount(e.db)
+			if err != nil {
+				continue
```
Tangui-Bitfly, line 60: log and sleep here

backend/pkg/exporter/modules/genesis_deposit.go
```diff
@@ -1,122 +1,121 @@
...
+	latestEpoch, err := db.GetLatestEpoch()
+	if err != nil {
+		log.Error(err, "error retrieving latest epoch from the database", 0)
+		time.Sleep(time.Second * 10)
+		return 0, err
+	}
+
+	return latestEpoch, nil
+}
+func (e genesisDepositsExporter) getGenesisDepositCount(db db.ConsensusDBI) (uint64, error) {
```
Tangui-Bitfly: if you move log and sleep in the parent function you will see that this func `getGenesisDepositCount ` has no purpose

backend/pkg/exporter/modules/genesis_deposit.go
```diff
@@ -1,122 +1,121 @@
...
-			log.Error(err, "error updating deposit count for the genesis slot", 0)
-			time.Sleep(time.Minute)
-			continue
+			log.Infof("exported genesis-deposits for %v genesis-validators", len(genesisValidators.Data))
+			return
 		}
+	}
+}
+
+func (e genesisDepositsExporter) getLatestEpoch(db db.ConsensusDBI) (uint64, error) {
```
Tangui-Bitfly: if you move log and sleep in the parent function you will see that this func `getLatestEpoch` has no purpose

backend/pkg/exporter/modules/genesis_deposit.go
```diff
@@ -1,122 +1,121 @@
...
 
-		if latestEpoch == 0 {
-			time.Sleep(time.Minute)
-			continue
-		}
+	offset time.Duration
+	ctx    context.Context
+}
 
-		// check if genesis-deposits have already been exported
```
Tangui-Bitfly, line 29: we lost that comment

backend/pkg/exporter/modules/genesis_deposit.go
```diff
@@ -1,122 +1,121 @@
...
+func newGenesisDepositsExporter(client rpc.Client, db db.ConsensusDBI) genesisDepositsExporter {
+	return genesisDepositsExporter{
+		client: client,
+		db:     db,
+		offset: time.Minute,
+		ctx:    context.Background(),
+	}
+}
 
-		// if genesis-deposits have already been exported exit this go-routine
```
Tangui-Bitfly, line 38: we lost that comment

backend/pkg/exporter/modules/genesis_deposit.go
Tangui-Bitfly: In the end this is probably not needed, we have an exit condition on this go routine with
```go
if genesisDepositCount > 0 {
	return
}
```

backend/pkg/commons/db/db.go
```diff
@@ -64,6 +64,19 @@ const DefaultInfScrollRows = 25
...
+}
+
+type ConsensusDBI interface {
+	GetLatestEpoch() (uint64, error)
+	GetDepositsCountForBlockSlot() (uint64, error)
+	SaveBlockDeposits(validatorIndex uint64, pubkey []byte, withdrawalCredentials []byte, balance uint64) error
```
Tangui-Bitfly: this should accept `types.StandardValidator`

backend/pkg/exporter/modules/genesis_deposit_test.go
```diff
@@ -0,0 +1,94 @@
...
+	"testing"
+	"time"
+
+	"github.com/gobitfly/beaconchain/pkg/commons/db/mocks"
+	rpcmocks "github.com/gobitfly/beaconchain/pkg/commons/rpc/mocks"
+	"github.com/gobitfly/beaconchain/pkg/consapi/types"
+	"github.com/pkg/errors"
+)
+
+func TestGenesisDepositsExporter_Export(t *testing.T) {
```
Tangui-Bitfly: This is getting better but it's not perfect yet.
What does this function really do ?
- verify if the genesis validators were already written to the db
- if it did return
- if not retrieve them and write them to the db

That the behavior we want to test. Yes we can test for all the possible error, but our two main path are `no records in db, retrieve from node, write to the db` and `records in db, exit`.
For case one, we should provide a list of validator. This list should be the response of the mocked rpc client. And we should expect this list on the mocked call `SaveBlockDeposits`.

backend/pkg/commons/db/db.go
Monika-Bitfly: using `types.StandardValidator` triggers a lint error due to a dependency rule tho

Tangui-Bitfly: Even with this syntax ?
```go
import (
	constypes "github.com/gobitfly/beaconchain/pkg/consapi/types"
)


func (d *ConsensusDB) SaveBlockDeposits(validator constypes.StandardValidator) error {
	...
}
```

backend/pkg/exporter/modules/mocks/ValidatorClient.go
Tangui-Bitfly, line 1: We don't have to put this mock here.
We can generate a mock for rpc.Client inside `rpc/mock` and use it in our test.

Tangui-Bitfly: Moreover, do we really want a mock here ? a simple stub would do the trick.
But I let you choose here.

backend/pkg/commons/db/db.go
Monika-Bitfly: Yeah

backend/pkg/exporter/modules/genesis_deposit_test.go
Monika-Bitfly: ok updated it now

Tangui-Bitfly (approved):

invis-bitfly (approved):

PR #1356, opened 2025-02-27
(BEDS-1282) fix notifications network prefix

remoterami:


.github/workflows/backend-integration-test.yml
```diff
@@ -37,9 +37,10 @@ jobs:
         working-directory: backend
         run: |
           go install github.com/swaggo/swag/cmd/swag@latest && swag init --ot json -o ./pkg/api/docs -d ./pkg/api/ -g ./handlers/public.go --overridesFile ./pkg/api/docs/.swaggo
-      - name: Tests
-        working-directory: backend
```
sasha-bitfly: no need to comment it out, you can always
```
name: Tests
if: false
```

to disable this step running

sasha-bitfly (changes requested):

backend/pkg/api/data_access/notifications.go
```diff
@@ -1541,10 +1541,22 @@ func (d *DataAccessService) GetNotificationSettingsDashboards(ctx context.Contex
 		Threshold float64         `db:"event_threshold"`
 	}{}
 
-	networkName := "mainnet"
-	if utils.Config.Chain.ClConfig.DepositChainID == 17000 {
-		networkName = "holesky"
+	networks, err := d.GetAllNetworks()
```
sasha-bitfly, line 1544: Can we write small and tiny unit test for `GetAllNetworks` ?

sasha-bitfly (changes requested):

backend/pkg/api/data_access/notifications.go
remoterami: It's currently just a placeholder for the actual network config selection, the method basically just returns a constant array so it's of limit use here. But sure I can try to write a short test.

sasha-bitfly (approved):

PR #1358, opened 2025-02-27
BEDS-1305: dispatch integration tests run

sasha-bitfly:
We run integration tests pipeline only when manually triggered from Actions UI
Added `-count=1` flag to stop caching results of tests run


remoterami (approved):
lgtm, but imo this should be more of a temporary change until we have integration tests tailored to a non-broken network again

PR #1360, opened 2025-02-27
(BEDS-1283) fix: only count notification subscriptions of current network

remoterami:


sasha-bitfly (changes requested):
can we make // `subscription counts` as a separate function that follows a pattern: build query (mandatory), execute query (mandatory), process the result (optional).

We have a generic call to execute a query:
https://github.com/gobitfly/beaconchain/blob/staging/backend/pkg/api/data_access/vdb_helpers.go#L486

remoterami:
> can we make // `subscription counts` as a separate function that follows a pattern: build query (mandatory), execute query (mandatory), process the result (optional).
> 
> We have a generic call to execute a query: https://github.com/gobitfly/beaconchain/blob/staging/backend/pkg/api/data_access/vdb_helpers.go#L486

This issue and similar ones are rather urgent bug fixes. Users are faced with wrong numbers, and including the fix in prod deployments is somewhat important. So in order to not increase scope delaying further work and slowing things down with detailed testing efforts I suggest extracting tests into separate tickets. This also allows applying individual priorities to both tasks. The fix should only be focused on the correct implementation of the fix itself.

The individual requirements might be small for this one, but this is more of a general comment.

PR #1364, opened 2025-02-27
refactor: revise and add tests for `exporter/modules/ssv.go`

Monika-Bitfly:


backend/pkg/commons/db/db.go
```diff
@@ -2422,6 +2422,91 @@ func GetValidatorAttestationHistoryForNotifications(startEpoch uint64, endEpoch
 	return epochParticipation, nil
 }
 
+func (c *ConsensusDB) SaveValidatorTags(valueStrings []string, valueArgs []interface{}) error {
```
Tangui-Bitfly: Prevent using `interface{}` here you can replace it with `[]byte`

backend/pkg/exporter/modules/ssv.go
```diff
@@ -15,152 +17,209 @@ import (
...
 		if err != nil {
-			return err
-		}
-		if rows == 0 {
-			break
+			log.Error(err, "error decoding public key", 0)
+			continue
 		}
-		time.Sleep(time.Millisecond * 100)
+		valueStrings = append(valueStrings, fmt.Sprintf("($%d, 'ssv')", i*index+1))
```
Tangui-Bitfly: This is sql query building. This should not be here.
The whole function should be inside `commons/db`.

backend/pkg/exporter/modules/ssv.go
```diff
@@ -15,152 +17,209 @@ import (
...
+		valueArgs = append(valueArgs, pubkey)
 	}
 
-	err = tx.Commit()
+	return valueStrings, valueArgs
+}
+
+type Dialer interface {
```
Tangui-Bitfly: Abstraction is at the wrong level, you can see it because you have to introduce an other interface for the response.

backend/pkg/exporter/modules/ssv_test.go
```diff
@@ -0,0 +1,140 @@
...
+		SSVExporter: types.SSVExporterConfig{
+			Address: "ws://localhost:8080",
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if tt.mockWebsocketError {
+				exporter.dialer = &StubDialerWebsocketError{}
+				exporter.Export()
```
Tangui-Bitfly: and then what ? what the test here ?

Tangui-Bitfly (changes requested):
What the real goal of this ssv exporter ? It connects to a client and every 10 minutes retrieves the ssv (or at least that what I understood). 
I find it harder to understand it after the refactoring. To be clear the original version is not well coded either.
But this refactoring adds layers of complexities: 
* ctx.Done which is not used but checked everywhere
* path hard to follow between all the functions: Export -> exportSSV -> handleWebSocketMessages -> saveSSV -> insertSSVTags -> prepareBatchInsert
* complex go routines

beside removing core sql query, it miss the removal of
* websockets
* sql manipulation in prepareBatchInsert

I think you went too far with this refactoring without speeding enough time on it. I totally understand wanting to improve the code, but if you start modifying touching something it need to be better afterward. It could be its own mulitiple days task since the original code is so overly convoluted.

If the task is just to add tests here, you should keep the `ConsensusDBI` and create a websocket server for tests.

backend/pkg/exporter/modules/ssv_test.go
```diff
@@ -0,0 +1,81 @@
...
+			defer conn.Close()
+
+			if err := conn.WriteJSON(&types.SSVExporterResponse{
+				Data: []types.SSVExporterData{{Publickey: "0xabcd"}},
+			}); err != nil {
+				log.Printf("Error writing JSON: %v", err)
+			}
+		}),
+	}
+}
```
Tangui-Bitfly: * use `httptest`
* extract the resp, here it's declared in two different places
* remove the `context.WithTimeout` it's not the purpose of this test, it only bring noise
```go
func TestSSVExport(t *testing.T) {
	mockConsDBClient := new(mocks.ConsensusDBI)

	response := &types.SSVExporterResponse{
		Data: []types.SSVExporterData{
			{
				Publickey: "0xabcd",
			},
		},
	}
	s := httptest.NewServer(http.HandlerFunc(wsHandler(t, response)))
	defer s.Close()

	utils.Config = &types.Config{
		SSVExporter: types.SSVExporterConfig{
			Address: strings.Replace(s.URL, "http://", "ws://", -1),
		},
	}

	t.Run("websocket works, data is saved and deleted from db", func(t *testing.T) {
		mockConsDBClient.On("DeleteInvalidTags").Return(nil)
		mockConsDBClient.On("SaveValidatorTags", response.Data).Return(nil)
		mockConsDBClient.On("DeleteValidatorTags").Return(nil)

		exporter := ssvExporter{
			db: mockConsDBClient,
		}

		err := exporter.exportSSV(context.Background())
		if err != nil {
			t.Errorf("expected no error, got: %v", err)
		}

		mockConsDBClient.AssertCalled(t, "DeleteInvalidTags")
		mockConsDBClient.AssertCalled(t, "SaveValidatorTags", response.Data)
		mockConsDBClient.AssertCalled(t, "DeleteValidatorTags")
	})
}

func wsHandler(t *testing.T, resp interface{}) func(w http.ResponseWriter, r *http.Request) {
	upgrader := &websocket.Upgrader{CheckOrigin: func(r *http.Request) bool { return true }}
	return func(w http.ResponseWriter, r *http.Request) {
		conn, err := upgrader.Upgrade(w, r, nil)
		if err != nil {
			t.Fatal(err)
		}
		defer conn.Close()

		if err := conn.WriteJSON(resp); err != nil {
			t.Fatal(err)
		}
	}
}
```

backend/pkg/exporter/modules/ssv.go
```diff
@@ -1,163 +1,124 @@
...
+			if err != nil {
+				return err
+			}
 		}
 	}
 }
 
-func saveSSV(res *SSVExporterResponse) error {
-	tx, err := db.WriterDb.Beginx()
+func (ssv *ssvExporter) saveSSV(res *types.SSVExporterResponse) error {
```
Tangui-Bitfly, line 100:
```go

func (ssv *ssvExporter) saveSSV(data []types.SSVExporterData) error {
```

Tangui-Bitfly (approved):
Nice it's a simpler amelioration of what we had previously

invis-bitfly (commented):
failing tests, but we dont use the ssv module on any network as far as i can tell, so prob not worth fixing. maybe just close the pr or even delete the module?

PR #1367, opened 2025-02-28
(BEDS-1236) use EB for dashboard limits

remoterami:
Based on #1365

- converts multiple queries to goqu
- introduces general approach for EB checking
- changes vdb retrieval and validator addition accordingly
- adjusts some related API response fields
- also applies different gnosis premium perks


backend/pkg/api/data_access/user.go
```diff
@@ -361,6 +367,66 @@ func (d *DataAccessService) GetUserDashboards(ctx context.Context, userId uint64
 		return nil
 	})
 
+	// TODO could merge with above query
```
sasha-bitfly, line 371: We need to create a separate function for this, otherwise our functions are too many lines

sasha-bitfly (approved):

backend/pkg/api/handlers/handler_service.go
```diff
@@ -146,8 +146,12 @@ func (h *HandlerService) getDashboardId(ctx context.Context, dashboardIdParam in
...
 			return nil, newNotFoundErr("no validators found for given id")
 		}
-		if len(validators) > maxValidatorsInList {
-			return nil, newBadRequestErr("too many validators in list, maximum is %d", maxValidatorsInList)
+		validatorEb, err := h.daService.GetValidatorDashboardEffectiveBalanceTotal(ctx, types.VDBId{Validators: validators})
+		if err != nil {
+			return nil, err
+		}
+		if validatorEb > maxEBInList {
+			return nil, newBadRequestErr("validators in list contain too much ETH, maximum is %d", maxEBInList)
```
LuccaBitfly: Question: Are we sure that we want an EB limit for guest dashboards? Our limiting issue here was URL space, so we don't want to run into that problem again allowing potentially more than 20 validators.

backend/pkg/api/handlers/public.go
```diff
@@ -551,31 +549,34 @@ func (h *HandlerService) PublicPostValidatorDashboardValidators(w http.ResponseW
...
 	if err != nil {
 		handleErr(w, r, err)
 		return
 	}
-	var limit uint64
-	if dashboardLimit > existingValidatorCount {
-		limit = dashboardLimit - existingValidatorCount
+	var spaceLeftEB uint64
+	if isUserAdmin(userInfo) {
+		spaceLeftEB = math.MaxUint64 // no limit for admins
```
LuccaBitfly: Issue: Don't need the admin check since we already get the admin perks from data access.

backend/pkg/api/data_access/vdb_management.go
```diff
@@ -758,14 +758,59 @@ func (d *DataAccessService) GetValidatorDashboardGroupExists(ctx context.Context
 	return groupExists, err
 }
 
-func (d *DataAccessService) AddValidatorDashboardValidators(ctx context.Context, dashboardId t.VDBIdPrimary, groupId uint64, validators []t.VDBValidator) ([]t.VDBPostValidatorsData, error) {
+func (d *DataAccessService) AddValidatorDashboardValidators(ctx context.Context, dashboardId t.VDBIdPrimary, groupId uint64, validators []t.VDBValidator, limitEB uint64) ([]t.VDBPostValidatorsData, error) {
```
LuccaBitfly: Nitpick: Should have consistent naming. `limitEB` is used here, but in `ebLimit` is used in https://github.com/gobitfly/beaconchain/blob/957d579ceed90cd90d16d82e56bae03f93bfd3ab/backend/pkg/api/data_access/vdb_management.go#L899

backend/pkg/api/data_access/vdb_management.go
```diff
@@ -758,14 +758,59 @@ func (d *DataAccessService) GetValidatorDashboardGroupExists(ctx context.Context
...
+			goqu.I("uvdv.dashboard_id").Eq(dashboardId),
+		)
+	existingValidatorsQuery, args, err := existingValidatorsDs.Prepared(true).ToSQL()
+	if err != nil {
+		return nil, fmt.Errorf("error preparing query: %w", err)
+	}
+	err = d.alloyReader.SelectContext(ctx, &existingValidators, existingValidatorsQuery, args...)
+	if err != nil {
+		return nil, err
+	}
```
LuccaBitfly: Issue: A function for getting existing validators in a dashboard from a slice already exists, see `GetValidatorDashboardValidatorsOfList`

backend/pkg/api/data_access/vdb_management.go
```diff
@@ -758,14 +758,59 @@ func (d *DataAccessService) GetValidatorDashboardGroupExists(ctx context.Context
...
+			return nil, err
+		}
+		newValidators, err = d.applyEBFiler(newValidatorEbs, limitEB)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	// keep existing validators so we can update their group
+	validators = append(existingValidators, newValidators...)
```
LuccaBitfly: Issue: All of this logic for determining whether the limit is reached should be in the handler, not data access.

backend/pkg/api/data_access/vdb_management.go
```diff
@@ -983,20 +1097,100 @@ func (d *DataAccessService) AddValidatorDashboardValidatorsByGraffiti(ctx contex
...
-		    dashboard_id = EXCLUDED.dashboard_id,
-		    group_id = EXCLUDED.group_id,
-		    validator_index = EXCLUDED.validator_index
-		RETURNING validator_index`, uniqueValidatorIndexesQuery)
+func (d *DataAccessService) applyEBFiler(validatorEbs map[t.VDBValidator]uint64, ebLimit uint64) ([]t.VDBValidator, error) {
+	// if shrink is true, new validators will be added until the ebLimit is reached; otherwise an error is returned
+	shrink := false
+	newValidatorsList := maps.Keys(validatorEbs)
+
+	if shrink {
```
LuccaBitfly: Issue: All "shrink" related code, which is currently unreachable, should be removed and not "iffed" out. If we find ourselves needing it again, we have it in our commit history.

backend/pkg/api/data_access/vdb_management.go
```diff
@@ -983,20 +1097,100 @@ func (d *DataAccessService) AddValidatorDashboardValidatorsByGraffiti(ctx contex
...
+			}
+		}
+		newEbAccumulator += validatorEbs[validator]
+	}
+	return newValidatorsList, nil
+}
+
+// takes a goqu ds of new validators to add to a vdb,
+// determines whether they fit in the ebLimit and returns an error if not
+func (d *DataAccessService) applyNewValidatorsDSEBFilter(ctx context.Context, newValidatorsDs *goqu.SelectDataset, ebLimit uint64) error {
```
LuccaBitfly, line 1071: Issue: I feel like this func entangles all `AddValidatorsBy<Thing>` functions with logic that doesn't need to happen there. I think all code that could potentially return a "eb limit reached" error should be performed as early as possible (as much as is possible in the handler), and only then should `AddValidatorsBy<Thing>` be called, which blindly adds the validators without any checks. IMO this would give us more clarity of what's happening.

backend/pkg/api/handlers/public.go
```diff
@@ -600,24 +601,8 @@ func (h *HandlerService) addValidatorDashboardValidatorsBySlice(r *http.Request,
...
-	validatorMap := utils.SliceToMap(existingValidatorsInDashboard)
-	slices.Sort(validators)
-	for i := 0; i < len(validators) && limit > 0; i++ {
-		validator := validators[i]
-		if _, ok := validatorMap[validator]; ok {
-			continue
-		}
-		validatorMap[validator] = struct{}{}
-		limit--
-	}
```
LuccaBitfly: Addendum: Already mentioned in a previous comment, but effectively the limit-related logic of the handler here got put in the data access func. Should be the other way around.

backend/pkg/api/handlers/public.go
```diff
@@ -974,7 +959,7 @@ func (h *HandlerService) PublicPutValidatorDashboardArchiving(w http.ResponseWri
 			returnConflict(w, r, errors.New("maximum number of groups in dashboards reached"))
 			return
 		}
-		if dashboardInfo.ValidatorCount >= userInfo.PremiumPerks.ValidatorsPerDashboard {
+		if dashboardInfo.EffectiveBalance.GreaterThanOrEqual(userInfo.PremiumPerks.EffectiveBalancePerDashboard) {
```
LuccaBitfly, line 988: Issue: Error message below should mention max EB, not validator count.

LuccaBitfly (changes requested):

backend/pkg/api/handlers/public.go
remoterami: I could swear I touched this somewhere already, good find :sweat_smile:

backend/pkg/api/data_access/vdb_management.go
remoterami: Then I think we don't have `AddValidatorsBy<Thing>` at all anymore, just one simple `AddValidators(dashboardId uint64, validators []uint64)` function. That allows us to move all checks and filterings based on search criteria etc. up to the handler, otherwise I don't see how these could be split.

I'd implement that in another PR though (see BEDS-1319)

backend/pkg/api/handlers/handler_service.go
remoterami: Discussed internally that we should indeed try going that way

LuccaBitfly (approved):
lgtm

PR #1369, opened 2025-02-28
(BEDS-1313) pectra calculations

remoterami:
Based on #1367

### DIFF
--- a/backend/pkg/api/data_access/vdb_helpers.go
+++ b/backend/pkg/api/data_access/vdb_helpers.go
@@ -151,7 +151,14 @@ func (d *DataAccessService) getTotalRewardsColumns() string {
 	return rewardColumns
 }
 
-func (d *DataAccessService) getElClAPR(ctx context.Context, dashboardId t.VDBId, groupId int64, hours int) (elIncome decimal.Decimal, elAPR float64, clIncome decimal.Decimal, clAPR float64, err error) {
+type IncomeInfo struct {
+	Rewards t.ClElValue[decimal.Decimal]
+
+	Apr t.ClElValue[float64]
+}
+
+func (d *DataAccessService) getElClAPR(ctx context.Context, dashboardId t.VDBId, groupId int64, hours int) (rewardsApr IncomeInfo, err error) {
+	result := IncomeInfo{}
 	table := ""
 
 	switch hours {
@@ -166,7 +173,7 @@ func (d *DataAccessService) getElClAPR(ctx context.Context, dashboardId t.VDBId,
 	case -1:
 		table = "validator_dashboard_data_rolling_90d"
 	default:
-		return decimal.Zero, 0, decimal.Zero, 0, fmt.Errorf("invalid hours value: %v", hours)
+		return IncomeInfo{}, fmt.Errorf("invalid hours value: %v", hours)
 	}
 
 	type RewardsResult struct {
@@ -204,16 +211,16 @@ func (d *DataAccessService) getElClAPR(ctx context.Context, dashboardId t.VDBId,
 
 	query, args, err := rewardsDs.Prepared(true).ToSQL()
 	if err != nil {
-		return decimal.Zero, 0, decimal.Zero, 0, fmt.Errorf("error preparing query: %w", err)
+		return IncomeInfo{}, fmt.Errorf("error preparing query: %w", err)
 	}
 
 	err = d.clickhouseReader.GetContext(ctx, &rewardsResultTable, query, args...)
 	if err != nil || !rewardsResultTable.Reward.Valid {
-		return decimal.Zero, 0, decimal.Zero, 0, err
+		return IncomeInfo{}, err
 	}
 
 	if rewardsResultTable.ValidatorCount == 0 {
-		return decimal.Zero, 0, decimal.Zero, 0, nil
+		return IncomeInfo{}, nil
 	}
 
 	aprDivisor := hours
@@ -223,15 +230,15 @@ func (d *DataAccessService) getElClAPR(ctx context.Context, dashboardId t.VDBId,
 
 	investedAmountUInt, err := d.GetValidatorDashboardEffectiveBalanceTotal(ctx, dashboardId, true)
 	if err != nil {
-		return decimal.Zero, 0, decimal.Zero, 0, fmt.Errorf("error retrieving total effective balance: %w", err)
+		return IncomeInfo{}, fmt.Errorf("error retrieving total effective balance: %w", err)
 	}
 	// invested amount is wrong if the effective balance changed during the period (because of auto compound, consolidation, partial withdrawal etc.)
 	// would need to split at eb changes and weigh results
 	investedAmount := d.convertClToMain(decimal.NewFromUint64(investedAmountUInt))
 
-	clAPR = calcAPR(d.convertClToMain(decimal.NewFromInt(rewardsResultTable.Reward.Int64)), investedAmount, aprDivisor)
+	result.Apr.Cl = calcAPR(d.convertClToMain(decimal.NewFromInt(rewardsResultTable.Reward.Int64)), investedAmount, aprDivisor)
 
-	clIncome = decimal.NewFromInt(rewardsResultTable.Reward.Int64).Mul(decimal.NewFromInt(1e9))
+	result.Rewards.Cl = decimal.NewFromInt(rewardsResultTable.Reward.Int64).Mul(decimal.NewFromInt(1e9))
 
 	if hours == -1 {
 		rewardsDs = rewardsDs.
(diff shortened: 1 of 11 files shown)


sasha-bitfly (approved):

backend/pkg/api/data_access/general.go
```diff
@@ -173,3 +177,83 @@ func runQuery[T any](ctx context.Context, db *sqlx.DB, ds *goqu.SelectDataset) (
 
 	return result, nil
 }
+
+// returns the effective balances of the provided validators
+// if onlyActive = true: executed from the vdb premium limits pov, i.e. exited validators account for the EB at exit time
+func (d *DataAccessService) GetValidatorsEffectiveBalances(ctx context.Context, validators []t.VDBValidator, onlyActive bool) (map[t.VDBValidator]uint64, error) {
```
LuccaBitfly: Thought: should maybe go into a `validators.go` file.

backend/pkg/api/data_access/general.go
```diff
@@ -173,3 +177,83 @@ func runQuery[T any](ctx context.Context, db *sqlx.DB, ds *goqu.SelectDataset) (
...
+		if !onlyActive &&
+			(status == constypes.DbSlashed || status == constypes.DbExited) &&
+			validatorMapping.ValidatorMetadata[validator].EffectiveBalance == 0 {
+			// exited & balance withdrawn, need to query latest EB before exit
+			if !validatorMapping.ValidatorMetadata[validator].ExitEpoch.Valid {
+				return nil, fmt.Errorf("validator %d has no exit epoch", validator)
+			}
+			// goqu can't pass tuples directly
+			epochTs := utils.EpochToTime(uint64(validatorMapping.ValidatorMetadata[validator].ExitEpoch.Int64)).Unix()
+			validatorExitEpochs = append(validatorExitEpochs, goqu.L("(fromUnixTimestamp(?), ?)", epochTs, validator))
```
LuccaBitfly: Nitpick: should do early continue and have the bulk of the code outside the if, something along these lines:
```suggestion
		plsNameThisVar := !onlyActive &&
			(status == constypes.DbSlashed || status == constypes.DbExited) &&
			validatorMapping.ValidatorMetadata[validator].EffectiveBalance == 0
		if !plsNameThisVar {
			effectiveBalances[validator] = validatorMapping.ValidatorMetadata[validator].EffectiveBalance
			continue
		}
		// exited & balance withdrawn, need to query latest EB before exit
		if !validatorMapping.ValidatorMetadata[validator].ExitEpoch.Valid {
			return nil, fmt.Errorf("validator %d has no exit epoch", validator)
		}
		// goqu can't pass tuples directly
		epochTs := utils.EpochToTime(uint64(validatorMapping.ValidatorMetadata[validator].ExitEpoch.Int64)).Unix()
		validatorExitEpochs = append(validatorExitEpochs, goqu.L("(fromUnixTimestamp(?), ?)", epochTs, validator))
```

backend/pkg/api/data_access/general.go
```diff
@@ -173,3 +177,83 @@ func runQuery[T any](ctx context.Context, db *sqlx.DB, ds *goqu.SelectDataset) (
...
+		}{}
+		err = d.clickhouseReader.SelectContext(ctx, &ebsBeforeExit, query, args...)
+		if err != nil && err != sql.ErrNoRows {
+			return nil, err
+		}
+		for _, eb := range ebsBeforeExit {
+			effectiveBalances[eb.ValidatorIndex] = eb.EffectiveBalance
+		}
+	}
+	return effectiveBalances, nil
```
LuccaBitfly: Nitpick: Should early return to have less indented code:
```suggestion
	if len(validatorExitEpochs) == 0 {
		return effectiveBalances, nil
	}

	ds := goqu.Dialect("postgres").
		Select(
			// goqu.SUM(goqu.I("balance_effective_end")).As("balance_effective_end"),
			goqu.I("validator_index"),
			goqu.I("balance_effective_end"),
		).
		From("validator_dashboard_data_epoch").
		Where(
			goqu.L("(epoch_timestamp, validator_index)").In(validatorExitEpochs),
		)
	query, args, err := ds.Prepared(true).ToSQL()
	if err != nil {
		return nil, err
	}

	ebsBeforeExit := []struct {
		ValidatorIndex   uint64 `db:"validator_index"`
		EffectiveBalance uint64 `db:"balance_effective_end"`
	}{}
	err = d.clickhouseReader.SelectContext(ctx, &ebsBeforeExit, query, args...)
	if err != nil && err != sql.ErrNoRows {
		return nil, err
	}
	for _, eb := range ebsBeforeExit {
		effectiveBalances[eb.ValidatorIndex] = eb.EffectiveBalance
	}
	return effectiveBalances, nil
```

backend/pkg/api/data_access/general.go
```diff
@@ -173,3 +177,83 @@ func runQuery[T any](ctx context.Context, db *sqlx.DB, ds *goqu.SelectDataset) (
...
+		return 0, err
+	}
+
+	var totalEb uint64
+	for _, validator := range validators {
+		if eb, ok := validatorEbs[validator]; !ok {
+			return 0, fmt.Errorf("effective balance for validator %d not found", validator)
+		} else {
+			totalEb += eb
+		}
```
LuccaBitfly: Nitpick: Unnecessary else
```suggestion
		}
		totalEb += eb
```

backend/pkg/api/data_access/user.go
```diff
@@ -361,6 +367,61 @@ func (d *DataAccessService) GetUserDashboards(ctx context.Context, userId uint64
 		return nil
 	})
 
+	// TODO could merge with above query
+	validatorDashboardEBs := make(map[uint64]uint64, 0)
+	wg.Go(func() error {
+		validatorsDs := goqu.Dialect("postgres").
```
LuccaBitfly: Nitpick: All of the code inside this goroutine should be inside it's own function (e.g. `getUserValidatorDashboardEBs`) as to not bloat `GetUserDashboards` with more lines.

backend/pkg/api/data_access/user.go
```diff
@@ -361,6 +367,61 @@ func (d *DataAccessService) GetUserDashboards(ctx context.Context, userId uint64
...
+		}
+
+		for dashboard_id, validators := range dashboardValidatorsMap {
+			var dashboardTotalEb uint64
+			for _, validator := range validators {
+				if eb, ok := validatorEbs[validator]; !ok {
+					return fmt.Errorf("effective balance for validator %d not found", validator)
+				} else {
+					dashboardTotalEb += eb
+				}
```
LuccaBitfly: Nitpick: Unnecessary else
```suggestion
				dashboardTotalEb += eb
```

backend/pkg/api/data_access/vdb_helpers.go
```diff
@@ -211,10 +213,15 @@ func (d *DataAccessService) getElClAPR(ctx context.Context, dashboardId t.VDBId,
 		aprDivisor = 90 * 24
 	}
 
-	// invested amount is not post-pectra safe
-	investedAmount := d.convertClToMain(decimal.NewFromUint64(d.config.ClConfig.MaxEffectiveBalance))
+	investedAmountUInt, err := d.GetValidatorDashboardEffectiveBalanceTotal(ctx, dashboardId, true)
+	if err != nil {
+		return decimal.Zero, 0, decimal.Zero, 0, fmt.Errorf("error retrieving total effective balance: %w", err)
```
LuccaBitfly: Thought: we should maybe just return an internal struct here instead of 4 values + 1 error

LuccaBitfly (changes requested):

backend/pkg/api/data_access/general.go
remoterami: Well it allows you to reuse `eb`, without the else you've to extract the inline assignment from the if (or invert the condition and `continue` over throwing an error). But that's also fine :+1:

backend/pkg/api/data_access/user.go
remoterami: solved in the base PR, see https://github.com/gobitfly/beaconchain/pull/1367#discussion_r1975439129 (sorry)

backend/pkg/api/data_access/user.go
remoterami: also applied in https://github.com/gobitfly/beaconchain/pull/1367

backend/pkg/api/data_access/general.go
LuccaBitfly: IMO using `else` should be avoided as much as possible so I prefer extracting the inline assignment. If you think otherwise I'm fine with conceding, this is only a nitpick after all :wink:

LuccaBitfly (approved):
lgtm

PR #1372, opened 2025-02-28
refactor: revise and add tests for `exporter/modules/relays.go`

Monika-Bitfly:


backend/pkg/commons/db/consensus_db.go
```diff
@@ -0,0 +1,22 @@
...
+	"github.com/gobitfly/beaconchain/pkg/commons/types"
+	"github.com/jmoiron/sqlx"
+)
+
+type ConsensusDB struct {
+	WriterDb *sqlx.DB
+	ReaderDb *sqlx.DB
+}
+
+type ConsensusDBI interface {
```
Tangui-Bitfly: it conflicts with https://github.com/gobitfly/beaconchain/pull/1354 `ConsensusDBI` definition.
I suggest to find a better name (suffix `I` for an interface is not a good pattern) or remove that interface (the real go way would be to let caller define interface unless we have multiple implementation)

backend/pkg/exporter/modules/relays.go
```diff
@@ -27,34 +27,61 @@ type BidTrace struct {
 	Value                types.WeiString `json:"value"`
 }
 
-func mevBoostRelaysExporter() {
-	var relays []types.Relay
+type RelayClient interface {
+	fetchDeliveredPayloads(endpoint string, id string, offset uint64) ([]BidTrace, error)
+}
+
+type relayClient struct{}
```
Tangui-Bitfly: this should be close to its methods. Order should be like this
```go
type foo struct{}

func newFoo() foo { ... }

func (foo) Foo() { ... }

func (foo) Bar() { ... }
```

backend/pkg/exporter/modules/relays.go
```diff
@@ -27,34 +27,61 @@ type BidTrace struct {
 	Value                types.WeiString `json:"value"`
 }
 
-func mevBoostRelaysExporter() {
-	var relays []types.Relay
+type RelayClient interface {
```
Tangui-Bitfly: Try not export if possible.
Here you could
```go
type relayClient interface {
	fetchDeliveredPayloads(endpoint string, id string, offset uint64) ([]BidTrace, error)
}

type nodeClient struct{}

func (nodeClient) fetchDeliveredPayloads(endpoint string, id string, offset uint64) ([]BidTrace, error) {
	...
}

```

backend/pkg/exporter/modules/relays.go
```diff
@@ -27,34 +27,61 @@ type BidTrace struct {
...
-			for _, relay := range relays {
-				if shouldTryToExportRelay(relay) {
-					// create relay logger
+		select {
+		case <-rs.ctx.Done():
+			log.Info("MEV boost relays export loop cancelled")
+			return
+		default:
+			// we retrieve the relays from the db each loop to prevent having to restart the exporter for changes
+			relays, err := rs.db.GetRelays()
```
Tangui-Bitfly, line 42:
```go
relays, err := rs.db.GetRelays()
if err != nil {
	log.Error(err, "failed to retrieve relays from db", 0)
	time.Sleep(rs.delay)
	continue
}
var wg sync.WaitGroup
for _, relay := range relays {
...
time.Sleep(rs.delay)	
```
or move that into a function so we don't have to repeat the sleep

backend/pkg/exporter/modules/relays.go
```diff
@@ -158,92 +172,75 @@ func exportRelayBlocks(r types.Relay) error {
 	return nil
 }
 
-func retrieveAndInsertPayloadsFromRelay(r types.Relay, low_bound uint64, high_bound uint64) error {
-	tx, err := db.WriterDb.Beginx()
```
Tangui-Bitfly, line 162: This tx contains two modifications, which would be the equivalent now of `SaveBlocksTags` and `SaveBlocksRelays`. But now those two modifications happen in two different txs.
You could create a function `SaveBlocksTagsAndRelays` to solve this.

Tangui-Bitfly: Hum... that's even more than two modifications, it's 2 * number of payloads.
So `SaveBlocksTagsAndRelays` need to accept a list of payloads. If the linter scream again we need to raise to the rest of the team, I don't see the reason behind this linter rule and I don't want it to be a reason to shitify our codebase.

backend/pkg/commons/db/consensus_db.go
```diff
@@ -0,0 +1,22 @@
...
+	GetRelays() ([]types.Relay, error)
+	UpdateRelays(tagID, endpoint string) error
```
Tangui-Bitfly: `UpdateRelay`

backend/pkg/commons/db/consensus_db.go
```diff
@@ -0,0 +1,22 @@
...
+	UpdateRelayLastExportTry(tagID, endpoint string) error
+	UpdateRelayExportFailureCount(exportFailureCount uint64, tagID, endpoint string) error
+	GetFirstRelayBlock(tagID string) (types.RelayBlock, error)
+	GetLastRelayBlock(tagID string) (types.RelayBlock, error)
+	SaveBlocksTags(tagID string, slot uint64, blockHash []byte) error
```
Tangui-Bitfly: `SaveBlockTags`

backend/pkg/commons/db/consensus_db.go
```diff
@@ -0,0 +1,22 @@
...
+	SaveBlocksRelays(tagID string, slot uint64, payloadValue types.WeiString, blockHash, builderPubkey, proposerPubkey, proposerFeeRecipient []byte) error
```
Tangui-Bitfly: `SaveBlockRelays`

Tangui-Bitfly (commented):
I still not a fan of using mock it makes tests complicated to read compared to a using real implementation, but it's fine since we didn't have a clear decision around this subject.

backend/pkg/commons/db/consensus_db.go
mccreedy-bitfly: Why don't we call it `IndexedConsensusRepository`? The implementation of this repository would then be `IndexedDBConsensusRepository` (as opposed to `IndexedInMemoryConsensusRepository` or some other way of implementing this. This suggestions tries to follow the Repository pattern, that we use for accessing our Databases.

backend/pkg/commons/db/db.go
```diff
@@ -817,6 +817,130 @@ func GetRelayDataForIndexedBlocks(blocks []*types.Eth1BlockIndexed) (map[common.
 	return relaysDataMap, nil
 }
 
+func (c *ConsensusDB) GetRelays() ([]types.Relay, error) {
```
mccreedy-bitfly: Comment just on this line for simplicity, but everything here which implements ConsensusDB should go in a separate "ConsensusRepository" style file which implements only the interface defined in `consensus_db.go`

backend/pkg/exporter/modules/relays.go
Monika-Bitfly: ok updated it

backend/pkg/commons/db/consensus_db.go
Monika-Bitfly: ok updated it now

backend/pkg/exporter/modules/base.go
```diff
@@ -40,7 +42,7 @@ var Client *rpc.Client
 func StartAll(moduleCtx ModuleContext, modules []ModuleInterface, justV2 bool) {
 	if !justV2 {
 		ctx := context.Background()
-		consDB := db.NewDBConsensusRepository(db.WriterDb, db.ReaderDb)
+		consDB := consensus.NewConsensusRepository(db.WriterDb, db.ReaderDb)
```
Tangui-Bitfly: That creates repetition.
It should either be
* `consensus.NewRepository`
* `db.NewConsensusRepository` <- I prefer this one

backend/pkg/exporter/modules/base.go
```diff
@@ -7,8 +7,9 @@ import (
 	"time"
 
 	"github.com/gobitfly/beaconchain/pkg/commons/config"
+
 	"github.com/gobitfly/beaconchain/pkg/commons/db"
-	"github.com/gobitfly/beaconchain/pkg/commons/db2/consensus"
+	db2 "github.com/gobitfly/beaconchain/pkg/commons/db2/consensus"
```
Tangui-Bitfly: No that's not what I meant, don't rename package import unless there is a conflict.

Monika-Bitfly: there is a conflict coz we've already imported the db package for both ReaderDb and WriterDb, hence the rename of the consensus package to db2

Tangui-Bitfly: What the point of the `consensus` package ? Keeping those files in `db2` make more sense.

invis-bitfly (approved):

PR #1373, opened 2025-02-28
refactor: revise and add tests for `exporter/modules/sync_committees_count.go`

Monika-Bitfly:


backend/pkg/exporter/modules/sync_committees_count_test.go
```diff
@@ -0,0 +1,88 @@
...
+				mockConsDBClient.On("SaveSyncCommitteesCount", tt.mockDBPeriod, tt.mockCountSoFar).Return(nil)
+
+				exporter.Export()
+
+				mockConsDBClient.AssertCalled(t, "GetSyncCommitteesCountPerValidator")
+				mockConsDBClient.AssertCalled(t, "GetLatestFinalizedEpoch")
+				mockConsDBClient.AssertCalled(t, "SaveSyncCommitteesCount", tt.mockDBPeriod, tt.mockCountSoFar)
+
+				return
+			}
```
Tangui-Bitfly: This is basically an other test, imo it's better to split it two tests in that case without using table testing

Tangui-Bitfly (approved):

invis-bitfly (approved):

PR #1374, opened 2025-03-03
feat(ens): refactor ens + add tests

Tangui-Bitfly:
* separate ens indexing logic from `db.Bigtable` object
* add tests
* introduce postgres testcontainers


backend/pkg/executionlayer/ens.go
```diff
@@ -0,0 +1,326 @@
...
+		Expires:   expires,
+	})
+}
+
+func (importer ENSImporter) getEnsNamesForAddress(address common.Address, alreadyChecked *ensChecked) ([]string, error) {
+	if alreadyChecked.AddressAlreadyChecked(address) {
+		return nil, nil
+	}
+
+	names, err := importer.store.GetNamesForAddress(address)
```
Monika-Bitfly, line 191: In the old version of the code we were retrieving data for multiple addresses at once in a single database call. With the new approach we are fetching the ENS name for each individual address separately, is this change intentional?

backend/pkg/commons/db2/ens.go
```diff
@@ -0,0 +1,124 @@
...
+	WHERE
+		ens_name = $1
+	;`, name)
+	if err != nil {
+		if strings.Contains(fmt.Sprintf("%v", err), "invalid byte sequence") {
+			log.Warnf("could not delete ens name [%v]: %v", name, err)
+			return nil
+		}
+		return fmt.Errorf("error deleting ens name [%v]: %v", name, err)
+	}
```
Monika-Bitfly, line 87: In the previous code we had clear indication that the ens name has been [removed from db](https://github.com/gobitfly/beaconchain/blob/staging/backend/pkg/commons/db/ens.go#L418), can you add it back?

backend/pkg/commons/db2/ens.go
```diff
@@ -0,0 +1,124 @@
...
+	Expires   time.Time      `db:"valid_to"`
+}
+
+type ENSStore struct {
+	db *sqlx.DB
+}
+
+func NewENSStore(db *sqlx.DB) *ENSStore {
+	return &ENSStore{
+		db: db,
```
Monika-Bitfly, line 30: In the old code we were using both ReaderDb and WriterDb to handle separate read and write operations for ENS data - is there a specific reason for consolidating the read and write operations into a single db connection?

Tangui-Bitfly: Yes, because it was badly used and implemented. I suspect the read / write dissociation come from frontend, but from the indexing point of view we will not more heavily read than write. There is no scalability issues in our case.

When the api team will use that code, they will use it with `db2.NewENSStore(db.ReaderDb)` (opposed to us using `db2.NewENSStore(db.WriterDb)`).

This is configuration problem, it should not leak into our implementation.

backend/pkg/commons/db2/ens.go
Tangui-Bitfly: Spamming a log for each name deleted is not a clear indication.
If we judge that is an important data to log and monitor, it should not be done here.
Function `ENSImporter.Import` could return list of added and deleted names and this could be logged in the main loop, this create one log than can be enriched with context vs multiple logs without context.

backend/pkg/executionlayer/ens.go
Tangui-Bitfly: You will have to point me to where in the old code it was done like you describe, because it use the same sql query under the hood. I don't find the `retrieving data for multiple addresses at once in a single database call` code.

backend/pkg/commons/db2/ens.go
Tangui-Bitfly: The question we need to ask ourself when logging something is "Will someone read that log on purpose ? If yes, does they have enough context to understand when, why and what happened ?"

Monika-Bitfly: I get what you're saying but imo it's still important to log when data is removed so we can be sure it was actually deleted. I know logging every single deletion might be too much but having some kind of log for the removal itself feels like it adds a layer of safety

backend/pkg/executionlayer/ens.go
Monika-Bitfly: I was referring to [this function](https://github.com/gobitfly/beaconchain/blob/staging/backend/pkg/commons/db/ens.go#L371)

Tangui-Bitfly: I'm not sure how this function is relevant in our case. It was not used in the old version of the code for ens processing.
That would be a big change in the logic, we can do it in the future if needed.

PR #1382, opened 2025-03-06
(BEDS-1319) split validator addition

remoterami:
tackles https://github.com/gobitfly/beaconchain/pull/1367#discussion_r1979380445 / https://github.com/gobitfly/beaconchain/pull/1367#discussion_r1979424906

### DIFF
--- a/backend/pkg/api/data_access/validators.go
+++ b/backend/pkg/api/data_access/validators.go
@@ -7,13 +7,79 @@ import (
 	"strings"
 
 	"github.com/doug-martin/goqu/v9"
+	"github.com/doug-martin/goqu/v9/exp"
 	t "github.com/gobitfly/beaconchain/pkg/api/types"
+	"github.com/gobitfly/beaconchain/pkg/commons/utils"
+	constypes "github.com/gobitfly/beaconchain/pkg/consapi/types"
 )
 
 type ValidatorRepository interface {
+	GetValidatorsEffectiveBalances(ctx context.Context, validators []t.VDBValidator, onlyActive bool) (map[t.VDBValidator]uint64, error)
+
 	GetValidatorsByDepositAddress(ctx context.Context, depositAddress string) ([]t.VDBValidator, error)
 	GetValidatorsByWithdrawalCredentials(ctx context.Context, withdrawalCredentials string) ([]t.VDBValidator, error)
 	GetValidatorsByGraffiti(ctx context.Context, graffiti string) ([]t.VDBValidator, error)
+}
+
+// returns the effective balances of the provided validators
+// if onlyActive = true: executed from the vdb premium limits pov, i.e. exited validators account for the EB at exit time
+func (d *DataAccessService) GetValidatorsEffectiveBalances(ctx context.Context, validators []t.VDBValidator, onlyActive bool) (map[t.VDBValidator]uint64, error) {
+	validatorMapping, err := d.services.GetCurrentValidatorMapping()
+	if err != nil {
+		return nil, err
+	}
+
+	// active
+	effectiveBalances := make(map[t.VDBValidator]uint64)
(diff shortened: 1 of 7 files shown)


LuccaBitfly:
image
:clap:

backend/pkg/api/data_access/validators.go
```diff
@@ -2,98 +2,97 @@ package dataaccess
...
 		return nil, err
 	}
 
-	ebsBeforeExit := []struct {
-		ValidatorIndex   uint64 `db:"validator_index"`
-		EffectiveBalance uint64 `db:"balance_effective_end"`
-	}{}
-	err = d.clickhouseReader.SelectContext(ctx, &ebsBeforeExit, query, args...)
-	if err != nil && err != sql.ErrNoRows {
+	return validators, nil
```
LuccaBitfly: Nitpick: could simply do:
```go
return runQueryRows[[]uint64](ctx, d.alloyReader, validatorsDs)
```
(same for all other queries)

backend/pkg/api/handlers/public.go
```diff
@@ -549,104 +549,93 @@ func (h *HandlerService) PublicPostValidatorDashboardValidators(w http.ResponseW
...
-		return nil, err
+		handleErr(w, r, err)
+		return
+	}
+	var totalExistingEb uint64
+	for _, eb := range existingEBs {
+		totalExistingEb += eb
 	}
-	existingValidatorsMap := utils.SliceToMap(existingValidators)
+	ebSpaceLeft := ebLimit - totalExistingEb
```
LuccaBitfly: Issue: this could overflow if the archiver is slow, should check if `totalExistingEb` is smaller than `ebLimit` before.
```go
var ebSpaceLeft uint64
if (ebLimit > totalExistingEb){
       ebSpaceLeft = ebLimit - totalExistingEb
}
```

backend/pkg/api/data_access/vdb_management.go
```diff
@@ -1153,7 +873,9 @@ func (d *DataAccessService) RemoveValidatorDashboardValidators(ctx context.Conte
 	return err
 }
 
-func (d *DataAccessService) GetValidatorDashboardEffectiveBalanceTotal(ctx context.Context, dashboardId t.VDBId, onlyActive bool) (uint64, error) {
+// returns the effective balances of the provided validators
+// if onlyActive = true: executed from the vdb premium limits pov, i.e. exited validators account for the EB at exit time
+func (d *DataAccessService) GetValidatorDashboardEffectiveBalances(ctx context.Context, dashboardId t.VDBId, onlyActive bool) (map[t.VDBValidator]uint64, error) {
```
LuccaBitfly: Issue: This should not be a dashboard func where you pass the dashboard ID but simply `GetValidatorEffectiveBalances` where you pass a list of validators, since in https://github.com/gobitfly/beaconchain/blob/c38d76e42af164653f8a0e8041f7bf780c729211/backend/pkg/api/handlers/public.go#L577 you're "abusing" it as exactly that. Looking inside the func, if an actual dashboard ID is passed, the code simply fetches all validators in the dashboard and proceeds with the whole validator set, so there's no real performance gain here. Rather, the handler should fetch all existing dashboard validators and pass it to a `GetValidatorEffectiveBalances` here https://github.com/gobitfly/beaconchain/blob/c38d76e42af164653f8a0e8041f7bf780c729211/backend/pkg/api/handlers/public.go#L586

LuccaBitfly (changes requested):

backend/pkg/api/handlers/public.go
LuccaBitfly, line 1: Otherwise very nice way of unifying all types of adding validators. IMO it's much clearer what's happening now. :raised_hands:

LuccaBitfly (approved):
lgtm!

PR #1393, opened 2025-03-10
fix(DashboardValidatorManagmentModal): use correct value for current effective balance

enzo-bitfly:
current effective balance

See: BEDS-1271


tufunga:
https://github.com/gobitfly/beaconchain/pull/1393#top

PR #1395, opened 2025-03-10
fix: use current effective balance from `dashboards` data

enzo-bitfly:
See: BEDS-1258 and BEDS-1271

### DIFF
--- a/frontend/components/dashboard/DashboardValidatorManagementModal.vue
+++ b/frontend/components/dashboard/DashboardValidatorManagementModal.vue
@@ -273,6 +273,17 @@ const {
   displayCurrencyDefault,
   formatAmount,
 } = useCurrency()
+const userDashboardStore = useUserDashboardStore()
+const {
+  dashboards,
+} = storeToRefs(userDashboardStore)
+
+const currentEffectiveBalance = computed(() => {
+  const currentDashboard = dashboards.value?.validator_dashboards.find(
+    validatorDashboard => `${validatorDashboard.id}` === dashboardKey.value,
+  )
+  return currentDashboard?.effective_balance
+})
 
 const EFFECTIVE_BALANCE_LIMIT_GUEST_DASHBOARD_IN_ETH = '640'
 const effectiveBalanceLimitGuestDashboard = formatAmount(EFFECTIVE_BALANCE_LIMIT_GUEST_DASHBOARD_IN_ETH, {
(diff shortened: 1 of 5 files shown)


marcel-bitfly:
np: commit-msg
(if I remember correctly, it should be done this way for automations like changelog creation)
```
fix: use current effective balance from dashboards data

See: BEDS-1258
See: BEDS-1271
```

frontend/components/dashboard/DashboardValidatorManagementModal.vue
```diff
@@ -269,33 +280,51 @@ const totalValidators = computed(() => {
...
-    return effectiveBalanceLimitGuestDashboard
-  }
-  return premium_perks.value?.effective_balance_per_dashboard
-})
+  const hasReachedLimit = computed(() => {
+    if (!premium_perks.value?.effective_balance_per_dashboard || !currentEffectiveBalance.value) {
+      return false
+    }
+    return premium_perks.value?.effective_balance_per_dashboard
+      <= currentEffectiveBalance.value
```
marcel-bitfly: **nitpick:**  this way around is how we think about this

```suggestion
    return currentEffectiveBalance.value >= premium_perks.value?.effective_balance_per_dashboard
```

frontend/pages/dashboard/[[id]]/index.vue
```diff
@@ -25,6 +25,8 @@ const {
 const showInDevelopment = Boolean(useRuntimeConfig().public.showInDevelopment)
 const { t: $t } = useTranslation()
 
+const { hasReachedLimit } = useEffectiveBalance()
```
Comment (marcel-bitfly) [inline frontend/pages/dashboard/[[id]]/index.vue]: **thought (non-blocking):** tbh I think these inline composables add another `abstraction layer` which in general doesn't make it easier to understand the component

(I totally get that you want to make things clearer by encapsulating them)
imo `inline composables` are nice if they encapsulate `stateful logic` that have to do with the `use case` the component is created for but at the same time should not be used in other components, e.g.:

```
// BaseVideo.vue
const {play, pause, stop, isMuted} = useVideoPlayer()
```

In general a composable should not return a single property, but be used like that:
```vue
const count = useCounter()
```

frontend/pages/dashboard/[[id]]/index.vue
```diff
@@ -76,11 +78,11 @@ const {
   refreshDashboards,
   updateGuestDashboardKey,
 } = userDashboardStore
-
```
Comment (marcel-bitfly) [inline frontend/pages/dashboard/[[id]]/index.vue]: **nitpick:**  👀

marcel-bitfly (approved):

frontend/pages/dashboard/[[id]]/index.vue
Comment (enzo-bitfly) [reply frontend/pages/dashboard/[[id]]/index.vue]: Removed

frontend/pages/dashboard/[[id]]/index.vue
Comment (enzo-bitfly) [reply frontend/pages/dashboard/[[id]]/index.vue]: I know what you mean with the added layer of abstraction. 
I'll remove the `inline composables` added in this PR and we can continue working without them.

frontend/components/dashboard/DashboardValidatorManagementModal.vue
enzo-bitfly: True, done.

marcel-bitfly (approved):

PR #1397, opened 2025-03-11
fix: compare `effective balances` as big numbers

enzo-bitfly:
See: BEDS-1258
See: BEDS-1271

### DIFF
--- a/frontend/components/dashboard/DashboardValidatorManagementModal.vue
+++ b/frontend/components/dashboard/DashboardValidatorManagementModal.vue
@@ -307,7 +307,7 @@ const hasReachedLimit = computed(() => {
   if (!currentEffectiveBalance.value || !effectiveBalanceLimitPerDashboard.value) {
     return false
   }
-  return currentEffectiveBalance.value >= effectiveBalanceLimitPerDashboard.value
+  return isGreaterOrEqual(currentEffectiveBalance.value, effectiveBalanceLimitPerDashboard.value)
 })
 
 const hasPremiumPerkBulkAdding = computed(() => !!premium_perks.value?.bulk_adding)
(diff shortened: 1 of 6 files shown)


frontend/utils/bigMath.ts
```diff
@@ -24,3 +24,17 @@ export const divideBigNumbers = (
...
+ *   - `0` if both values are equal
+ *   - `1` if `leftHandValue` is greater than `rightHandValue`
+ */
+export const compareBigNumbers = (
+  leftHandValue: number | string,
+  rightHandValue: number | string,
+) => {
+  assertIsNumber(leftHandValue, rightHandValue)
+  return Big(leftHandValue).cmp(Big(rightHandValue))
+}
```
marcel-bitfly: **suggestion (blocking):** plz use `.gte method`

I think otherwise you are missing the `0` case and working with `number abstraction` does not work that well

See: https://mikemcl.github.io/big.js/#gte

marcel-bitfly:
**thought**(non-blocking): I don't know if it would be helpful to mention where `effective balance` is `compared` in the `commit message` 🤷

frontend/utils/bigMath.ts
enzo-bitfly: Done

frontend/components/dashboard/DashboardValidatorManagementModal.vue
```diff
@@ -307,7 +307,7 @@ const hasReachedLimit = computed(() => {
   if (!currentEffectiveBalance.value || !effectiveBalanceLimitPerDashboard.value) {
     return false
   }
-  return currentEffectiveBalance.value >= effectiveBalanceLimitPerDashboard.value
+  return isGreaterOrEqual(currentEffectiveBalance.value, effectiveBalanceLimitPerDashboard.value)
```
marcel-bitfly: **suggestion (non-blocking):** I think it reads easier (and a bit fewer to think about 🤷 )
```suggestion
  return isGreaterEquals(currentEffectiveBalance.value, effectiveBalanceLimitPerDashboard.value)
```

marcel-bitfly (approved):

frontend/components/dashboard/DashboardValidatorManagementModal.vue
enzo-bitfly: Done

PR #1398, opened 2025-03-12
fix(statistics): consider deposit-requests when calculating validator…

guybrush:
…-stats

BEDS-1335

### DIFF
--- a/backend/pkg/commons/db/statistics.go
+++ b/backend/pkg/commons/db/statistics.go
@@ -831,10 +831,11 @@ func gatherValidatorDepositWithdrawals(day uint64, data []*types.ValidatorStatsT
 		DepositsAmount uint64 `db:"deposits_amount"`
 	}
 	resDeposits := make([]*resRowDeposits, 0, 1024)
+	// In this query we are collecting all deposits per validator and sum them up. We need to consider that all deposits after the first valid deposits are valid. Note that there are currently only valid requests in the deposit_requests table.
 	depositsQry := `
-		with first_valid_deposits as (
-			select
-				distinct on (publickey)
+		with 
+		first_valid_deposits as (
+			select distinct on (publickey)
 				publickey,
 				block_slot,
 				block_index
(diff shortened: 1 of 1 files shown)


backend/pkg/commons/db/statistics.go
```diff
@@ -832,53 +832,106 @@ func gatherValidatorDepositWithdrawals(day uint64, data []*types.ValidatorStatsT
...
+		first_valid_deposit_requests as (
+			select
+				distinct on (pubkey)
+				pubkey,
+				block_slot,
+				request_index
+			from
+				blocks_deposit_requests
+			where
+				pubkey in (select bdr.pubkey from blocks_deposit_requests bdr where bdr.block_slot >= $1 and bdr.block_slot <= $2)
```
invis-bitfly, line 861: missing valid filter?

invis-bitfly (changes requested):

backend/pkg/commons/db/statistics.go
```diff
@@ -832,53 +832,106 @@ func gatherValidatorDepositWithdrawals(day uint64, data []*types.ValidatorStatsT
...
+			inner join first_valid_deposits on
+				blocks_deposits.publickey = first_valid_deposits.publickey
+			where
+				blocks.slot >= $1
+				and blocks.slot <= $2
+				and (blocks.status = '1'
+					or blocks.slot = 0)
+				and (blocks_deposits.block_slot > first_valid_deposits.block_slot  -- any slot after the first valid deposit
+					or (blocks_deposits.block_slot = first_valid_deposits.block_slot -- or the same slot but a higher block-index
+						and blocks_deposits.block_index >= first_valid_deposits.block_index)
```
LuccaBitfly: **issue:** Comment says higher block-index, but actual code contains `>=` (higher or equal). Is this correct? Same goes for deposit requests.

backend/pkg/commons/db/statistics.go
```diff
@@ -832,53 +832,106 @@ func gatherValidatorDepositWithdrawals(day uint64, data []*types.ValidatorStatsT
 	}
 	resDeposits := make([]*resRowDeposits, 0, 1024)
 	depositsQry := `
```
LuccaBitfly, line 835: **nitpick (non-blocking):** since this is quite a large query, a short comment explaining what this query does and why would prob. be helpful for future maintainers. I.e. explaining that after the first valid deposit all deposits are valid, which is why we fetch xzy, and that since pectra deposit requests are a thing, which is why we need to do it twice or something.

backend/pkg/commons/db/statistics.go
LuccaBitfly: +1

LuccaBitfly (changes requested):

backend/pkg/commons/db/statistics.go
invis-bitfly: comment is wrong, `>=` is correct. we want to include the "first valid deposit" itself as well, therefor `>=` not `>`

backend/pkg/commons/db/statistics.go
guybrush: can you elaborate on this? i thought only valid requests make it into this table?
```
\d blocks_deposit_requests
              Table "public.blocks_deposit_requests"
         Column         |  Type   | Collation | Nullable | Default 
------------------------+---------+-----------+----------+---------
 block_slot             | integer |           | not null | 
 block_root             | bytea   |           | not null | 
 request_index          | integer |           | not null | 
 pubkey                 | bytea   |           | not null | 
 withdrawal_credentials | bytea   |           | not null | 
 amount                 | bigint  |           | not null | 
 signature              | bytea   |           | not null | 
 queued_at_epoch        | integer |           |          | 
 processed_at_epoch     | integer |           |          | 
Indexes:
    "blocks_deposit_requests_pkey" PRIMARY KEY, btree (block_slot, block_root, request_index)

```

backend/pkg/commons/db/statistics.go
guybrush: :+1: added a comment

backend/pkg/commons/db/statistics.go
guybrush: i think you meant the query above which selects from blocks_deposits

i added check for valid_signature there

backend/pkg/commons/db/statistics.go
guybrush: updated the comment

PR #1400, opened 2025-03-12
refactor: revise and add tests for `exporter/modules/slot_exporter`

Monika-Bitfly:


backend/pkg/exporter/db/cache.go
```diff
@@ -0,0 +1,100 @@
...
+	defer cancel()
+	res, err := c.cache.Get(ctx, key)
+	if err != nil {
+		return 0, err
+	}
+	return new(big.Int).SetBytes(res).Uint64(), nil
+}
+
+func (c *SlotExporterCache) SetLatestEpoch(epoch uint64) error {
+	key := fmt.Sprintf("%d:frontend:latestEpoch", utils.Config.Chain.ClConfig.DepositChainID)
```
Tangui-Bitfly: Can you remove the use of `utils.Config.Chain.ClConfig.DepositChainID`.
That global value need to be removed, it creates hidden dependency and make thing a bite harder to test (you have to init `utils.Config` and this since it's global it can have border effect on other tests).
Either accept `chainID` as a param here or at `NewSlotExporterCache`

backend/pkg/exporter/db/db.go
```diff
@@ -392,7 +614,9 @@ func saveGraffitiwall(block *types.Block, tx *sqlx.Tx) error {
             validator
         )
         VALUES ($1, $2, $3, $4, $5)
-        ON CONFLICT (slot) DO UPDATE SET
+        ON CONFLICT 
+			(slot) 
+		DO UPDATE SET
```
Tangui-Bitfly, line 679: weird padding

backend/pkg/exporter/db/db.go
```diff
@@ -437,77 +661,19 @@ func saveGraffitiwall(block *types.Block, tx *sqlx.Tx) error {
...
 
-	for ; ; time.Sleep(time.Second) { // wait till the last attestation in memory cache has been populated by the exporter
-		db.BigtableClient.LastAttestationCacheMux.Lock()
-		if db.BigtableClient.LastAttestationCache != nil {
-			db.BigtableClient.LastAttestationCacheMux.Unlock()
-			break
-		}
-		db.BigtableClient.LastAttestationCacheMux.Unlock()
-		log.Infof("waiting until LastAttestation in memory cache is available")
-	}
```
Tangui-Bitfly, line 486: this for loop logic seems to be missing

Tangui-Bitfly (commented):
So I don't know how to review the pr. It's so big, moves code around, sometime changes the logic, sometime doesn't...
We have one foot in the old way and one foot in the new way.
All that to get the test `TestSlotExporter` which is too complex.

backend/pkg/exporter/db/db.go
Monika-Bitfly: nice catch, updated it now

PR #1404, opened 2025-03-14
fix: display latest effective balance to unauthenticated users

enzo-bitfly:
This adapts to `API` change that sends `latest effective balance` to a public endpoint: `dashboard.validator_dashboards[index].effective_balance` (private) is now `overview.balances.effective_latest` (public)

Also `overview.balances.effective` is now `overview.balances.effective_current`

This PR also fixes [BEDS-1337](https://bitfly1.atlassian.net/browse/BEDS-1337) because the current effective balance is now fetched from the overview, which was already refreshed when adding/removing validators before this change.

[BEDS-1337]: https://bitfly1.atlassian.net/browse/BEDS-1337?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/frontend/components/dashboard/DashboardValidatorManagementModal.vue
+++ b/frontend/components/dashboard/DashboardValidatorManagementModal.vue
@@ -270,44 +270,25 @@ const totalValidators = computed(() => {
 })
 
 const {
-  displayCurrencyDefault,
-  formatAmount,
-} = useCurrency()
-const userDashboardStore = useUserDashboardStore()
-const {
-  dashboards,
-} = storeToRefs(userDashboardStore)
+  premiumProducts,
+} = useProductsStore()
 
-const currentEffectiveBalance = computed(() => {
-  const currentDashboard = dashboards.value?.validator_dashboards.find(
-    validatorDashboard => `${validatorDashboard.id}` === dashboardKey.value,
-  )
-  return currentDashboard?.effective_balance
-})
-
-const EFFECTIVE_BALANCE_LIMIT_GUEST_DASHBOARD_IN_ETH = '640'
-const effectiveBalanceLimitGuestDashboard = formatAmount(EFFECTIVE_BALANCE_LIMIT_GUEST_DASHBOARD_IN_ETH, {
-  hasCurrencyDisplay: false,
-  maximumFractionDigits: 0,
-  minimumFractionDigits: 0,
-  sourceUnit: 'base',
-  targetCurrency: displayCurrencyDefault.main,
-  targetUnit: 'wei',
-  useGrouping: false,
+const latestEffectiveBalance = computed(() => {
+  return overview.value?.balances.effective_latest
 })
 
 const effectiveBalanceLimitPerDashboard = computed(() => {
-  if (isGuestDashboard.value || !premium_perks.value?.effective_balance_per_dashboard) {
-    return effectiveBalanceLimitGuestDashboard
-  }
-  return premium_perks.value?.effective_balance_per_dashboard
+  const freeProduct = premiumProducts.value['Free']
+  const effectiveBalanceLimitFreeProduct = freeProduct?.premium_perks.effective_balance_per_dashboard
+
+  return premium_perks.value?.effective_balance_per_dashboard ?? effectiveBalanceLimitFreeProduct
 })
 
 const hasReachedLimit = computed(() => {
-  if (!currentEffectiveBalance.value || !effectiveBalanceLimitPerDashboard.value) {
+  if (!latestEffectiveBalance.value || !effectiveBalanceLimitPerDashboard.value) {
     return false
   }
-  return isGreaterEquals(currentEffectiveBalance.value, effectiveBalanceLimitPerDashboard.value)
+  return isGreaterEquals(latestEffectiveBalance.value, effectiveBalanceLimitPerDashboard.value)
 })
 
 const hasPremiumPerkBulkAdding = computed(() => !!premium_perks.value?.bulk_adding)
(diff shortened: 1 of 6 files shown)


frontend/components/dashboard/DashboardValidatorManagementModal.vue
```diff
@@ -271,43 +271,28 @@ const totalValidators = computed(() => {
...
+const latestEffectiveBalance = computed(() => {
+  return overview.value?.balances.effective_latest
 })
 
 const effectiveBalanceLimitPerDashboard = computed(() => {
-  if (isGuestDashboard.value || !premium_perks.value?.effective_balance_per_dashboard) {
-    return effectiveBalanceLimitGuestDashboard
-  }
-  return premium_perks.value?.effective_balance_per_dashboard
+  const freeProduct = products.value?.premium_products.find(product => product.product_name === 'Free')
```
marcel-bitfly: **suggestion (non-blocking):** maybe you expose what you need from `useProductStore()`

frontend/components/dashboard/DashboardValidatorManagementModal.vue
```diff
@@ -595,15 +580,15 @@ const inputValidator = ref('')
                 >
                   <span>
                     <BcFormatAmount
-                      :value="currentEffectiveBalance ?? '0'"
+                      :value="latestEffectiveBalance ?? '0'"
                       :maximum-fraction-digits="0"
-                      :source-currency="displayCurrencyDefault.main"
+                      :target-currency="displayCurrencyDefault.main"
```
marcel-bitfly: **polish (non-blocking):** you might expose `mainDisplayCurrency` in `BcFormatAmount`

When I have a look at how `BcFormatAmount` is used, it might make sense to adapt and map `mainDisplayCurrency` to `displayCurrencyDefault.main` inside `BcFormatAmount`

frontend/pages/dashboard/[[id]]/index.vue
```diff
@@ -81,7 +81,6 @@ const {
   cookieDashboards,
   dashboards,
 } = storeToRefs(userDashboardStore)
-await useAsyncData('user_dashboards', () => refreshDashboards(), { watch: [ isLoggedIn ] })
```
Comment (marcel-bitfly) [inline frontend/pages/dashboard/[[id]]/index.vue:84]: **question:** this is not needed anymore?

frontend/components/dashboard/DashboardValidatorManagementModal.vue
enzo-bitfly: Done.
By the way, we don't have type inference for these products but it should be solved with [this API ticket](https://bitfly1.atlassian.net/browse/BEDS-1342)

frontend/components/dashboard/DashboardValidatorManagementModal.vue
enzo-bitfly: Let's discuss this out of the scope of this fix. I don't completely get what you mean.

frontend/pages/dashboard/[[id]]/index.vue
Comment (enzo-bitfly) [reply frontend/pages/dashboard/[[id]]/index.vue:84]: No. It was duplicated, see line 119 of this file.

enzo-bitfly (commented):
Thank you for the suggestions

frontend/components/dashboard/DashboardValidatorManagementModal.vue
marcel-bitfly: yeah sure I just meant the same I did with `elDisplayCurrency` for example

frontend/stores/useProductsStore.ts
```diff
@@ -45,6 +45,18 @@ export function useProductsStore() {
...
+  const premiumProductsByName = computed(() => {
+    if (!data.value?.api_products) return {}
+
+    return Object.fromEntries(data.value.premium_products.map(({
+      product_name,
+      ...rest
+    }) => [
+      product_name,
+      rest,
+    ]))
```
marcel-bitfly: **thought (non-blocking):** yeah you can do this, but there is also nothing wrong with keepin' the `product_name` part of the object (as you do not know how `premiumProductsByName` would be used.

Btw. I would keep the name `premiumProdcuts`, as this is your public api now (but just an opinion -> non-blocking)

frontend/components/dashboard/DashboardValidatorManagementModal.vue
enzo-bitfly: I see. Done

frontend/components/dashboard/DashboardValidatorManagementModal.vue
```diff
@@ -271,43 +271,28 @@ const totalValidators = computed(() => {
...
+  return overview.value?.balances.effective_latest
 })
 
 const effectiveBalanceLimitPerDashboard = computed(() => {
-  if (isGuestDashboard.value || !premium_perks.value?.effective_balance_per_dashboard) {
-    return effectiveBalanceLimitGuestDashboard
-  }
-  return premium_perks.value?.effective_balance_per_dashboard
+  const freeProduct = premiumProductsByName.value['Free']
+  const effectiveBalanceLimitFreeProduct = freeProduct?.premium_perks.effective_balance_per_dashboard
```
marcel-bitfly: **suggestion (non-blocking):** you are exposing a `product_name` which might be part of our domain now
```suggestion
  const free = premiumProductsByName.value['Free']
  const effectiveBalanceLimitFree = free?.premium_perks.effective_balance_per_dashboard
```

marcel-bitfly (approved):
nice how with every commit things get cleaner and cleaner 💪

frontend/stores/useProductsStore.ts
enzo-bitfly: I'm renaming to `premiumProducts` and including `product_name` back in

frontend/components/dashboard/DashboardValidatorManagementModal.vue
enzo-bitfly: I prefer to keep the `Product` in, or it would be hard to understand what `free` is if used some lines apart from it's definition

PR #1405, opened 2025-03-14
feat: removed network-specific premium limits

remoterami:


backend/pkg/commons/db/user.go
```diff
@@ -230,11 +230,7 @@ func GetUserInfo(ctx context.Context, userId uint64, userDbReader *sqlx.DB) (*t.
 }
 
 func premiumLimitNetworkfactor() int64 {
```
sasha-bitfly: Please, add the comment to this function to explain why does it `return 1`

sasha-bitfly (approved):

PR #1406, opened 2025-03-14
fix(notifications): use goroutines to reduce notifications delay

sasha-bitfly:
Reducing notification delays by parallelizing notifications sending and queuing with Goroutines

### DIFF
--- a/backend/pkg/notification/queuing.go
+++ b/backend/pkg/notification/queuing.go
@@ -33,27 +33,32 @@ func queueNotifications(epoch uint64, notificationsByUserID types.NotificationsP
 
 	err = QueueEmailNotifications(epoch, notificationsByUserID, tx)
 	if err != nil {
+		metrics.Errors.WithLabelValues("notifications_queuing_email").Inc()
 		return fmt.Errorf("error queuing email notifications: %w", err)
 	}
 
 	err = QueuePushNotification(epoch, notificationsByUserID, tx)
 	if err != nil {
+		metrics.Errors.WithLabelValues("notifications_queuing_push").Inc()
 		return fmt.Errorf("error queuing push notifications: %w", err)
 	}
 
 	err = QueueWebhookNotifications(notificationsByUserID, tx)
 	if err != nil {
+		metrics.Errors.WithLabelValues("notifications_queuing_webhook").Inc()
 		return fmt.Errorf("error queuing webhook notifications: %w", err)
 	}
 
 	err = tx.Commit()
 	if err != nil {
+		metrics.Errors.WithLabelValues("notifications_queuing_tx_commit").Inc()
 		return fmt.Errorf("error committing transaction: %w", err)
 	}
 
 	err = ExportNotificationHistory(epoch, notificationsByUserID)
 	if err != nil {
-		log.Error(err, "error exporting notification historyw", 0)
+		log.Error(err, "error exporting notification history", 0)
+		metrics.Errors.WithLabelValues("notifications_export_history").Inc()
 	}
 
 	subByEpoch := map[uint64][]uint64{}
--- a/backend/pkg/notification/sending.go
+++ b/backend/pkg/notification/sending.go
@@ -9,6 +9,7 @@ import (
 	"net/http"
 	"net/url"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/gobitfly/beaconchain/pkg/commons/db"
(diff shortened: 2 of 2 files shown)


backend/pkg/notification/queuing.go
```diff
@@ -31,19 +32,47 @@ func queueNotifications(epoch uint64, notificationsByUserID types.NotificationsP
...
+
+	wg.Wait()
+	close(errChan)
+
+	log.Infof("Notifications queueing finished for epoch %d in %v", epoch, time.Since(start))
+
+	// checks for errors before committing the transaction
+	for err := range errChan {
+		if err != nil {
+			log.Error(err, "rrror in queuing notifications", 0)
```
mccreedy-bitfly: rrror -> error

backend/pkg/notification/queuing.go
```diff
@@ -24,26 +25,56 @@ import (
...
+
+	// checks for errors before committing the transaction
+	for err := range errChan {
+		if err != nil {
+			log.Error(err, "error in queuing notifications", 0)
+			return err
+		}
 	}
 
 	err = tx.Commit()
```
mccreedy-bitfly, line 52: Seems like a lot to do in a single transaction, which might actually be the cause of some of our dropped notifications (i.e, notification fails to queue, transaction rolls back, queueing it isn't retried). However, I think that the change here isn't making the situation worse, mearly not improving it, so let's save that for a future CR

backend/pkg/notification/queuing.go
```diff
@@ -24,26 +25,56 @@ import (
...
+
+	log.Infof("Notifications queueing finished for epoch %d in %v", epoch, time.Since(start))
+
+	// checks for errors before committing the transaction
+	for err := range errChan {
+		if err != nil {
+			log.Error(err, "error in queuing notifications", 0)
+			return err
+		}
 	}
```
mccreedy-bitfly: Will encountering this error result in a transaction rollback? Can we record a metric here to track how often this happens?

mccreedy-bitfly: Fine if you want to include this in a future CR. My point here is that we have metrics for Sending of notifications now, but we haven't done a deep investigation on the queueing process, so I think that needs some attention from a metrics perspective next

sasha-bitfly: Yes, it makes sense to reuse existing metrics counter or add a new one

backend/pkg/notification/queuing.go
sasha-bitfly: Done

backend/pkg/notification/queuing.go
sasha-bitfly: The problem now is that we will rollback all 3 notifications queuing even if one of them fails

backend/pkg/notification/queuing.go
sasha-bitfly: Added `metrics.Errors.WithLabelValues("notifications_queuing_notifications").Inc()`

peterbitfly (changes requested):
db transactions are not safe for concurrent use as they are bound to a single db connection.

sasha-bitfly:
> db transactions are not safe for concurrent use as they are bound to a single db connection.

@peterbitfly For `queueNotifications` I started using `db.WriterDb.NamedExec` instead of `tx.NamedExec`

peterbitfly:
> > db transactions are not safe for concurrent use as they are bound to a single db connection.
> 
> @peterbitfly For `queueNotifications` I started using `db.WriterDb.NamedExec` instead of `tx.NamedExec`

the reason we have a tx here is that we can safely roll back all queued tx in the error case (db becomes unavailable while we queue notifications). this allows us to safely re-try queuing later without any chance of queuing any notification twice. removing the tx risks queueing multiple notifications twice should an error occur.

sasha-bitfly:
> > > db transactions are not safe for concurrent use as they are bound to a single db connection.
> > 
> > 
> > @peterbitfly For `queueNotifications` I started using `db.WriterDb.NamedExec` instead of `tx.NamedExec`
> 
> the reason we have a tx here is that we can safely roll back all queued tx in the error case (db becomes unavailable while we queue notifications). this allows us to safely re-try queuing later without any chance of queuing any notification twice. removing the tx risks queueing multiple notifications twice should an error occur.

@peterbitfly Got you. I have updated the PR. Please, have a look.

sasha-bitfly:
Closing this PR
We decided to redesign notifications mechanism completely instead
@peterbitfly @mccreedy-bitfly

PR #1408, opened 2025-03-17
BEDS 94/test post validators

LuccaBitfly:
Adds test for `POST /validator-dashboards/{dashboard_id}/validators` while resolving a few misc. things. Pls look at each commit individually for a clear progression.


backend/pkg/api/handlers/validator_dashboard_test.go
```diff
@@ -221,168 +222,112 @@ func validatorDashboardTestSetup(options ...vdbStubOption) (context.Context, *Ha
 }
 
 // ------------------------------------------------------------
-// POST /validator-dashboards/{dashboard_id}/groups
-
-func TestInputPostValidatorDashboardGroupsValidate(t *testing.T) {
```
LuccaBitfly, line 77: Githubs diff view is a bit weird here, all the funcs "removed" from this file in this commit just got moved

PR #1413, opened 2025-03-20
(BEDS-1238) EB-adjusted APR

remoterami:


sasha-bitfly:
@remoterami Please, write how was the changes tested

remoterami:
Via manual testing. Here you can see an example:
before:
image
after:
image

sasha-bitfly:
@remoterami The results after the changes are correct and more precise, right?

backend/pkg/api/data_access/vdb_helpers.go
```diff
@@ -310,11 +286,11 @@ func (d *DataAccessService) getElClAPR(ctx context.Context, dashboardId t.VDBId,
 }
 
 // precondition: invested amount and rewards are in the same currency
-func calcAPR(rewards, investedAmount decimal.Decimal, aprDivisor int) float64 {
+func calcAPR(rewards, investedAmount decimal.Decimal, aprDivisor uint64) float64 {
 	if rewards.IsZero() || investedAmount.IsZero() {
 		return 0
 	}
-	return (rewards.Div(decimal.NewFromInt(int64(aprDivisor))).Div(investedAmount).Mul(decimal.NewFromInt(24 * 365 * 100))).InexactFloat64()
+	return (rewards.Div(investedAmount).Sub(decimal.NewFromInt(1)).Mul(decimal.NewFromInt(100 * 365 * 24 * 60 * 60).Div(decimal.NewFromInt(int64(aprDivisor))))).InexactFloat64()
```
sasha-bitfly: Please, add the comment to explain the calculation. Should `100 * 365 * 24 * 60 * 60` be converted to a constant?

remoterami:
> @remoterami The results after the changes are correct and more precise, right?

Yeah sorry the ticket information was missing, I've included some information. In essence this PR is making use of newly introduced db columns in order to account for in-interval EB changes, which could significantly skew the numbers (e.g. for bigger consolidations)

backend/pkg/api/data_access/vdb_helpers.go
```diff
@@ -214,48 +215,32 @@ func (d *DataAccessService) getElClAPR(ctx context.Context, dashboardId t.VDBId,
...
-	}
-
-	// invested amount is wrong if the effective balance changed during the period (because of auto compound, consolidation, partial withdrawal etc.)
-	// would need to split at eb changes and weigh results
-	investedAmountDec := d.convertClToMain(decimal.NewFromUint64(investedAmount))
-
-	result.Apr.Cl = calcAPR(d.convertClToMain(decimal.NewFromInt(rewardsResultTable.Reward.Int64)), investedAmountDec, aprDivisor)
-
-	result.Rewards.Cl = decimal.NewFromInt(rewardsResultTable.Reward.Int64).Mul(decimal.NewFromInt(1e9))
+	aprDivisor := d.config.Chain.ClConfig.SecondsPerSlot * d.config.Chain.ClConfig.SlotsPerEpoch
```
LuccaBitfly: **issue:** You still need to multiply with the number of epochs you've fetched, right? Currently you have the same `aprDivisor` for all timeframes.

backend/pkg/api/data_access/vdb_helpers.go
```diff
@@ -310,11 +286,11 @@ func (d *DataAccessService) getElClAPR(ctx context.Context, dashboardId t.VDBId,
 }
 
 // precondition: invested amount and rewards are in the same currency
-func calcAPR(rewards, investedAmount decimal.Decimal, aprDivisor int) float64 {
+func calcAPR(rewards, investedAmount decimal.Decimal, aprDivisor uint64) float64 {
```
LuccaBitfly: **nitpick:** I think the term `aprDivisor` is a bit confusing here. IMO the argument should be of type `time.Duration` and be named something like `timeFrame`, which would make annualizing the ROI much more intuitive.

backend/pkg/api/data_access/vdb_helpers.go
```diff
@@ -279,28 +264,19 @@ func (d *DataAccessService) getElClAPR(ctx context.Context, dashboardId t.VDBId,
...
-	}
-
-	err = d.alloyReader.GetContext(ctx, &result.Rewards.El, query, args...)
+	result.Rewards.El, err = runQuery[decimal.Decimal](ctx, d.alloyReader, elTableDs)
 	if err != nil {
 		return IncomeInfo{}, err
 	}
 
-	result.Apr.El = calcAPR(d.convertElToMain(result.Rewards.El), investedAmountDec, aprDivisor)
+	roiDividendEl := d.convertElToMain(result.Rewards.El).Add(d.convertClToMain(decimal.NewFromBigInt(rewardsResultTable.RoiDivisor, 0)))
```
LuccaBitfly: **thought (non-blocking):** Instead of adding the invested amount to the rewards here only to subtract 1 in the division of `calcAPR`, we could rather subtract the divisor from the dividend in the CL APR calculation and not subtract 1 later. Don't know if this would make it simpler, doesn't make any difference either way in the result.

backend/pkg/api/data_access/vdb_helpers.go
```diff
@@ -182,13 +182,13 @@ func (d *DataAccessService) getElClAPR(ctx context.Context, dashboardId t.VDBId,
 	}
```
LuccaBitfly, line 172: **question:** Why do we still use the 90d table for the all time APR calculation?

backend/pkg/api/data_access/vdb_helpers.go
```diff
@@ -162,7 +162,7 @@ type IncomeInfo struct {
 	Apr t.ClElValue[float64]
 }
 
-func (d *DataAccessService) getElClAPR(ctx context.Context, dashboardId t.VDBId, groupId int64, hours int, investedAmount uint64) (rewardsApr IncomeInfo, err error) {
+func (d *DataAccessService) getElClAPR(ctx context.Context, dashboardId t.VDBId, groupId int64, hours int) (rewardsApr IncomeInfo, err error) {
```
LuccaBitfly: **suggestion:** We already have an enum for all timeframes, and we already have a helper function converting that enum to hours + its corresponding table. I also see in the code that there are a lot of places which hardcode the hours passed onto this function, which makes it difficult to add or remove timeframes in future (See [here](https://github.com/gobitfly/beaconchain/blob/150616f88247beff5879d1402f7237c41a9ab2a5/backend/pkg/api/data_access/vdb_management.go#L453-L456) or [here](https://github.com/gobitfly/beaconchain/blob/150616f88247beff5879d1402f7237c41a9ab2a5/backend/pkg/api/data_access/mobile.go#L340-L342)). Would it not be better to have the enum as a function argument, and not the hours? To make this even more attractive, note that the enum has a [duration function](https://github.com/gobitfly/beaconchain/blob/150616f88247beff5879d1402f7237c41a9ab2a5/backend/pkg/api/enums/enums.go#L144), which would tie in to my previous comment about the APR divisor.

LuccaBitfly (changes requested):

backend/pkg/api/data_access/vdb_helpers.go
remoterami: It should be correct because that information is implicitly given from the data now. The `RoiDividend` aren't the absolute rewards anymore, but a running (epoch-based) aggregate of all end balances (excl. balance changes caused by events). The `RoiDivisor` works the same, so we're basically calculating the average APR per epoch - hence the only relevant factor for extrapolation now is the chain's epoch duration

backend/pkg/api/data_access/vdb_helpers.go
remoterami: I've changed the param to a `time.Duration` and restructured the function a bit, ptal

backend/pkg/api/data_access/vdb_helpers.go
remoterami: Done, ptal

I also simplified `getTablesForPeriod` by removing the 2nd return value `hours` as it was a) redundant to that duration function and b) discarded in 95% anyways

backend/pkg/api/data_access/vdb_helpers.go
remoterami: I missed this, should be fixed with the switch to `getTablesForPeriod`

backend/pkg/api/data_access/vdb_helpers.go
```diff
@@ -216,50 +203,23 @@ func (d *DataAccessService) getElClAPR(ctx context.Context, dashboardId t.VDBId,
...
-		}
-
-		err = d.clickhouseReader.GetContext(ctx, &rewardsResultTotal, query, args...)
-		if err != nil || !rewardsResultTotal.Reward.Valid {
-			return IncomeInfo{}, err
-		}
-
-		result.Rewards.Cl = decimal.NewFromInt(rewardsResultTotal.Reward.Int64).Mul(decimal.NewFromInt(1e9))
-	}
+	epochDuration := time.Duration(d.config.Chain.ClConfig.SecondsPerSlot * d.config.Chain.ClConfig.SlotsPerEpoch * uint64(time.Second))
```
LuccaBitfly: **nitpick (non-blocking):** 
```suggestion
	epochDuration := time.Second * time.Duration(d.config.Chain.ClConfig.SecondsPerSlot*d.config.Chain.ClConfig.SlotsPerEpoch)
```
It's a bit confusing but this is the "idiomatic" way as it's done in the documentation.
image

backend/pkg/api/data_access/vdb_helpers.go
```diff
@@ -281,42 +241,24 @@ func (d *DataAccessService) getElClAPR(ctx context.Context, dashboardId t.VDBId,
...
-	}
+	result.Apr.El = calcAPR(d.convertElToMain(result.Rewards.El), cumulativeClBaseMain, epochDuration)
 
 	return result, nil
 }
 
 // precondition: invested amount and rewards are in the same currency
-func calcAPR(rewards, investedAmount decimal.Decimal, aprDivisor int) float64 {
-	if rewards.IsZero() || investedAmount.IsZero() {
+func calcAPR(rewards, cumulativeDivisor decimal.Decimal, epochDuration time.Duration) float64 {
```
LuccaBitfly: **nitpick (non-blocking):** Even though we currently always pass the epoch duration, this func works just as well for all durations, so the name `epochDuration` might be misleading.

backend/pkg/api/data_access/vdb_helpers.go
```diff
@@ -281,42 +241,24 @@ func (d *DataAccessService) getElClAPR(ctx context.Context, dashboardId t.VDBId,
...
+	if rewards.IsZero() || cumulativeDivisor.IsZero() {
 		return 0
 	}
-	return (rewards.Div(decimal.NewFromInt(int64(aprDivisor))).Div(investedAmount).Mul(decimal.NewFromInt(24 * 365 * 100))).InexactFloat64()
+	epochDurationScaleFactor := decimal.NewFromInt(int64(utils.Year.Seconds())).Div(decimal.NewFromInt(int64(epochDuration.Seconds())))
```
LuccaBitfly: **nitpick (non-blocking):** Unnecessary conversion
```suggestion
	annualizationFactor := decimal.NewFromInt(utils.Year.Nanoseconds()).Div(decimal.NewFromInt(epochDuration.Nanoseconds()))
```

backend/pkg/api/data_access/vdb_management.go
```diff
@@ -455,10 +450,10 @@ func (d *DataAccessService) GetValidatorDashboardOverview(ctx context.Context, d
...
 	}
 
-	retrieveRewardsAndEfficiency("validator_dashboard_data_rolling_24h", 24, &data.Rewards.Last24h, &data.Apr.Last24h, &data.Efficiency.Last24h)
-	retrieveRewardsAndEfficiency("validator_dashboard_data_rolling_7d", 7*24, &data.Rewards.Last7d, &data.Apr.Last7d, &data.Efficiency.Last7d)
-	retrieveRewardsAndEfficiency("validator_dashboard_data_rolling_30d", 30*24, &data.Rewards.Last30d, &data.Apr.Last30d, &data.Efficiency.Last30d)
-	retrieveRewardsAndEfficiency("validator_dashboard_data_rolling_total", -1, &data.Rewards.AllTime, &data.Apr.AllTime, &data.Efficiency.AllTime)
+	retrieveRewardsAndEfficiency("validator_dashboard_data_rolling_24h", enums.Last24h, &data.Rewards.Last24h, &data.Apr.Last24h, &data.Efficiency.Last24h)
+	retrieveRewardsAndEfficiency("validator_dashboard_data_rolling_7d", enums.Last7d, &data.Rewards.Last7d, &data.Apr.Last7d, &data.Efficiency.Last7d)
+	retrieveRewardsAndEfficiency("validator_dashboard_data_rolling_30d", enums.Last30d, &data.Rewards.Last30d, &data.Apr.Last30d, &data.Efficiency.Last30d)
+	retrieveRewardsAndEfficiency("validator_dashboard_data_rolling_total", enums.AllTime, &data.Rewards.AllTime, &data.Apr.AllTime, &data.Efficiency.AllTime)
```
LuccaBitfly: **issue:** Passing the table name here is redundant, right? I.e. you could just pass the enum and get the table inside the func with `getTablesForPeriod`

LuccaBitfly (changes requested):
This is almost perfect! I still have some small remarks if you can bear with me.

backend/pkg/api/data_access/vdb_management.go
remoterami: Same applies to usages in the `services` package.. https://github.com/gobitfly/beaconchain/blob/BEDS-1238/update-apr/backend/pkg/api/services/service_average_network_efficiency.go#L109

I'd just define this method on the type directly

LuccaBitfly (approved):
lgtm!!

PR #1414, opened 2025-03-20
(BEDS-1368) adjust deposits endpoints to new api structure, pectra changes

remoterami:


backend/pkg/api/data_access/dummy.go
```diff
@@ -369,19 +369,19 @@ func (*DummyService) GetValidatorDashboardGroupHeatmap(ctx context.Context, dash
...
 	return getDummyWithPaging[t.VDBConsensusDepositsTableRow](ctx)
 }
 
-func (*DummyService) GetValidatorDashboardTotalElDeposits(ctx context.Context, dashboardId t.VDBId) (*t.VDBTotalExecutionDepositsData, error) {
+func (*DummyService) GetValidatorDashboardTotalElDeposits(ctx context.Context, dashboardId t.VDBId, search string) (*t.VDBTotalExecutionDepositsData, error) {
 	return getDummyStruct[t.VDBTotalExecutionDepositsData](ctx)
 }
 
-func (*DummyService) GetValidatorDashboardTotalClDeposits(ctx context.Context, dashboardId t.VDBId) (*t.VDBTotalConsensusDepositsData, error) {
+func (*DummyService) GetValidatorDashboardTotalClDeposits(ctx context.Context, dashboardId t.VDBId, search string) (*t.VDBTotalConsensusDepositsData, error) {
```
LuccaBitfly, line 439: **question:** the total endpoints should be affected by search?

backend/pkg/api/handlers/input_validation.go
```diff
@@ -348,6 +348,24 @@ func (v *validationError) checkPagingParams(q url.Values) Paging {
 	return paging
 }
 
+func (v *validationError) checkPagingMap(params map[string]string) Paging {
```
LuccaBitfly: **note (non-blocking):** down the line this func should totally replace `checkPagingParams` :+1: (feel free to resolve this comment)

backend/pkg/api/handlers/validator_dashboard.go
```diff
@@ -249,3 +249,163 @@ func (h *HandlerService) GetValidatorDashboardSummaryChart(ctx context.Context,
...
+func (i *inputGetValidatorDashboardTotalConsensusLayerDeposits) Validate(params map[string]string, _ io.ReadCloser) error {
+	var v validationError
+	i.search = params["search"]
+	i.dashboardId = v.checkDashboardId(params["dashboard_id"])
+	return v.AsError()
+}
+
+type inputGetValidatorDashboardTotalConsensusLayerDeposits struct {
+	search      string
+	dashboardId interface{}
```
LuccaBitfly: **nitpick (non-blocking):** This should be `dashboardIdParam` and not `dashboardId`, but looking through the code I've been inconsistent with this myself :sweat_smile:

backend/pkg/api/types/validator_dashboard.go
```diff
@@ -240,19 +240,22 @@ type VDBExecutionDepositsTableRow struct {
...
-	Epoch                uint64          `json:"epoch"`
-	Slot                 uint64          `json:"slot"`
+	SlotQueued           uint64          `json:"slot_queued"`
+	SlotProcessed        uint64          `json:"slot_processed"`
 	WithdrawalCredential Hash            `json:"withdrawal_credential"`
 	Amount               decimal.Decimal `json:"amount"`
 	Signature            Hash            `json:"signature"`
+	Type                 string          `json:"type" tstype:"'manual' | 'auto'" faker:"oneof: manual, auto"`
+	Status               string          `json:"status" tstype:"'queued' | 'completed' | 'rejected'" faker:"oneof: queued, completed, rejected"`
+	RejectReason         *string         `json:"reject_reason,omitempty" tstype:"'invalid_signature'" faker:"oneof: invalid_signature"`
```
LuccaBitfly: **suggestion:** You could add more faker fields which the frontend team would appreciate, e.g. `faker:"eth"` to the `decimal.Decimal` fields for proper Wei values already exists. See: https://github.com/gobitfly/beaconchain/blob/538cc1e4d0e92864e20b834c3f7b21ebe61e6f40/backend/pkg/api/data_access/dummy.go#L32-L34 Could also add more custom providers for hashes, signatures etc., up to you really.

frontend/types/api/validator_dashboard.ts
```diff
@@ -207,18 +207,21 @@ export interface VDBExecutionDepositsTableRow {
...
-  slot: number /* uint64 */;
+  slot_queued: number /* uint64 */;
+  slot_processed: number /* uint64 */;
   withdrawal_credential: Hash;
   amount: string /* decimal.Decimal */;
   signature: Hash;
+  type: 'manual' | 'auto';
+  status: 'queued' | 'completed' | 'rejected';
+  reject_reason?: 'invalid_signature';
 }
```
LuccaBitfly, line 224: **nitpick:** Don't know if it's applicable in this case, but frontend could appreciate a union type when there are mutually exclusive fields (assuming it is the case here with `rejection_reason`):
```ts
export type VDBConsensusDepositsTableRow =
  | {
      public_key: PubKey;
      index: number /* uint64 */;
      group_id: number /* uint64 */;
      slot_queued: number /* uint64 */;
      slot_processed: number /* uint64 */;
      withdrawal_credential: Hash;
      amount: string /* decimal.Decimal */;
      signature: Hash;
      type: 'manual' | 'auto';
      status: 'queued' | 'completed';
    }
  | {
      public_key: PubKey;
      index: number /* uint64 */;
      group_id: number /* uint64 */;
      slot_queued: number /* uint64 */;
      slot_processed: number /* uint64 */;
      withdrawal_credential: Hash;
      amount: string /* decimal.Decimal */;
      signature: Hash;
      type: 'manual' | 'auto';
      status: 'rejected';
      rejection_reason: 'invalid_signature';
    };
```
Though this has to be coded manually into the go file, similarly to how it's done [here](https://github.com/gobitfly/beaconchain/blob/538cc1e4d0e92864e20b834c3f7b21ebe61e6f40/backend/pkg/api/types/search.go#L39) to achieve [this](https://github.com/gobitfly/beaconchain/blob/538cc1e4d0e92864e20b834c3f7b21ebe61e6f40/frontend/types/api/search.ts#L34-L36). There is probably a cleaner way to achieve this using the `tygo:emit` directive, (see docs [here](https://github.com/gzuidhof/tygo/blob/main/README.md#custom-type-mapping)) but I haven't put too much thought into it. If you decide to implement this, I'd be happy to help out.

backend/pkg/api/types/validator_dashboard.go
```diff
@@ -240,19 +240,22 @@ type VDBExecutionDepositsTableRow struct {
 	TxHash               Hash            `json:"tx_hash"`
 	WithdrawalCredential Hash            `json:"withdrawal_credential"`
 	Amount               decimal.Decimal `json:"amount"`
-	Valid                bool            `json:"valid"`
+	Validity             string          `json:"validity" tstype:"'valid' | 'invalid' | 'invalid_skipped'" faker:"oneof: valid, invalid, invalid_skipped"`
```
LuccaBitfly, line 243: **question:** 
image
This is an instance of `invalid_skipped`?

frontend/types/api/validator_dashboard.ts
```diff
@@ -207,18 +207,21 @@ export interface VDBExecutionDepositsTableRow {
...
-  epoch: number /* uint64 */;
-  slot: number /* uint64 */;
+  slot_queued: number /* uint64 */;
+  slot_processed: number /* uint64 */;
   withdrawal_credential: Hash;
   amount: string /* decimal.Decimal */;
   signature: Hash;
+  type: 'manual' | 'auto';
+  status: 'queued' | 'completed' | 'rejected';
+  reject_reason?: 'invalid_signature';
```
LuccaBitfly, line 223: **issue:** Is this correct? The reject reason can only be `'invalid_signature'` and nothing else? In this case, the field is redundant, right?

backend/pkg/api/types/validator_dashboard.go
```diff
@@ -240,19 +240,22 @@ type VDBExecutionDepositsTableRow struct {
...
 }
 type GetValidatorDashboardExecutionLayerDepositsResponse ApiPagingResponse[VDBExecutionDepositsTableRow]
 
 type VDBConsensusDepositsTableRow struct {
```
LuccaBitfly, line 247: **question:** From CL deposits design:
image
Which field will cause this text to be shown?

LuccaBitfly (changes requested):

backend/pkg/api/data_access/dummy.go
remoterami: I think so, yes, that's what we are currently doing for withdrawals at least https://github.com/gobitfly/beaconchain/blob/staging/backend/pkg/api/data_access/vdb_withdrawals.go#L440-L446

backend/pkg/api/types/validator_dashboard.go
remoterami: Thanks for the tip, played around and added a few examples. ptal

backend/pkg/api/types/validator_dashboard.go
remoterami: Yes correct.

For context: The signature only has to be correct for a validators first deposit, in all subsequent cases the signature verification is skipped. So even if the signature is technically incorrect in these cases, the deposit will still be included & accepted on chain. `invalid_skipped` was the best validity type naming for this that I came up with

frontend/types/api/validator_dashboard.ts
remoterami: > In this case, the field is redundant, right?

Technically yes, but I chose to do it like this for 2 reasons:
- compatability with future changes: similar to how pectra introduced rejection reasons for withdrawals, the same could happen to deposits later
- consistency with the other event table endpoints (withdrawals & consolidations)

backend/pkg/api/types/validator_dashboard.go
remoterami: If `type == auto`; I've added an explainer comment to the designs

frontend/types/api/validator_dashboard.ts
marcel-bitfly: dunno if it would make sense to strip out only the parts that actually are different?

```typescript
{
      public_key: PubKey;
      index: number /* uint64 */;
      group_id: number /* uint64 */;
      slot_queued: number /* uint64 */;
      slot_processed: number /* uint64 */;
      withdrawal_credential: Hash;
      amount: string /* decimal.Decimal */;
      signature: Hash;
      type: 'manual' | 'auto';
    }
  & ( 
       {
         status: 'rejected';
         rejection_reason: 'invalid_signature';
       }
       | {
       status: 'queued' | 'completed';
       }
   );
```

LuccaBitfly: after discussing with FE team, won't do the extra work on our side until NOT having union types gives them more extra work on their side.

LuccaBitfly (approved):
lgtm

backend/pkg/api/enums/validator_dashboard_enums.go
```diff
@@ -248,6 +248,76 @@ var VDBWithdrawalsColumns = struct {
...
+func (VDBDepositsElColumn) NewFromString(s string) VDBDepositsElColumn {
+	switch s {
+	case "", "block":
+		return VDBDepositElBlock
+	case "amount":
+		return VDBDepositElAmount
+	default:
+		return VDBDepositsElColumn(-1)
+	}
+}
```
enzo-bitfly: question: could you add the alias to sort by `timestamp`?

remoterami: right sure, done

PR #1416, opened 2025-03-20
feat: add hoodi

guybrush:


backend/pkg/commons/config/hoodi.chain.yml
InvisibleSymbol, line 1: maybe a stretch but can we consider not adding this file and instead tell everyone to load it from the node? That's what we do in prod already anyways, no?

PR #1418, opened 2025-03-21
(BEDS-1234) API: Consolidations tabs

remoterami:


frontend/types/api/validator_dashboard.ts
```diff
@@ -247,6 +247,31 @@ export interface VDBTotalWithdrawalsData {
   total_amount: string /* decimal.Decimal */;
 }
 export type GetValidatorDashboardTotalWithdrawalsResponse = ApiDataResponse<VDBTotalWithdrawalsData>;
+/**
+ * ------------------------------------------------------------
+ * Consolidations Tab
+ */
+export interface VDBConsolidationsElTableRow {
```
enzo-bitfly, line 254: `timestamp` field is missing, right? same for CL

frontend/types/api/validator_dashboard.ts
```diff
@@ -247,6 +247,31 @@ export interface VDBTotalWithdrawalsData {
...
+  block_processed: number /* uint64 */;
+  tx_hash: Hash;
+  fee: string /* decimal.Decimal */;
+}
+export type GetValidatorDashboardExecutionLayerConsolidationsResponse = ApiPagingResponse<VDBConsolidationsElTableRow>;
+export interface VDBConsolidationsClTableRow {
+  source: number /* uint64 */;
+  target: number /* uint64 */;
+  slot_queued: number /* uint64 */;
+  slot_processed: number /* uint64 */;
```
enzo-bitfly, line 269: These two are `Block` and not `Slot` in the designs for the `CL consolidations`. Should I ask Philip to change them?

enzo-bitfly (commented):
Thank you for the mocks :)
Reminder: `Age` and `Amount` fields need to be sortable.

frontend/types/api/validator_dashboard.ts
remoterami: Timestamp is implicitly given by the slot number, that's how we usually did it. But that is for CL only (`slot_queued` / `slot_processed`) - for EL you're right, can't calculate that, will add it :+1:

frontend/types/api/validator_dashboard.ts
enzo-bitfly: The comment belongs in the `EL` object above.

PR #1419, opened 2025-03-21
(BEDS-1370) API: Consolidation tab endpoints

remoterami:
duplicated by #1418 (see comments there)

Based on #1414

PR #1421, opened 2025-03-21
(BEDS-1371) API: pectra withdrawals endpoint adjustments

remoterami:
Based on #1419


backend/pkg/api/types/validator_dashboard.go
LuccaBitfly, line 1: **suggestion:** You could add more faker fields which the frontend team would appreciate. See https://github.com/gobitfly/beaconchain/pull/1414#discussion_r2009717438

frontend/types/api/validator_dashboard.ts
```diff
@@ -233,20 +233,46 @@ export type GetValidatorDashboardTotalConsensusDepositsResponse = ApiDataRespons
  * ------------------------------------------------------------
  * Withdrawals Tab
  */
-export interface VDBWithdrawalsTableRow {
-  epoch: number /* uint64 */;
-  slot: number /* uint64 */;
+export interface VDBWithdrawalsElTableRow {
```
LuccaBitfly, line 238: **question:** This type seemingly has more fields than I can find in the design. Where do these come from?

- `executer`
- `is_missing_estimate`

frontend/types/api/validator_dashboard.ts
```diff
@@ -233,20 +233,46 @@ export type GetValidatorDashboardTotalConsensusDepositsResponse = ApiDataRespons
...
+  block_queued: number /* uint64 */;
+  timestamp_queued: number /* int64 */;
+  block_processed: number /* uint64 */;
+  timestamp_processed: number /* int64 */;
+  index: number /* uint64 */;
+  tx_hash: Hash;
+  group_id: number /* uint64 */;
+  sender: Address;
+  executer: Address;
```
LuccaBitfly: **typo:** executer -> executor?

frontend/types/api/validator_dashboard.ts
```diff
@@ -233,20 +233,46 @@ export type GetValidatorDashboardTotalConsensusDepositsResponse = ApiDataRespons
...
+  amount: string /* decimal.Decimal */;
+  status: 'queued' | 'processed';
+  fee: string /* decimal.Decimal */;
+  is_missing_estimate: boolean;
+}
+export type GetValidatorDashboardExecutionLayerWithdrawalsResponse = ApiPagingResponse<VDBWithdrawalsElTableRow>;
```
LuccaBitfly, line 251: **nitpick:** Don't know if it's applicable in this case, but frontend would appreciate a union type when there are mutually exclusive fields. See https://github.com/gobitfly/beaconchain/pull/1414#discussion_r2009730547.

backend/pkg/api/handlers/input_validation.go
```diff
@@ -373,19 +373,19 @@ func (v *validationError) parseSortOrder(order string) bool {
 	}
 }
 
-func checkSort[T enums.EnumFactory[T]](v *validationError, sortString string) *types.Sort[T] {
+func checkSort[T enums.EnumFactory[T]](v *validationError, sortString string) types.Sort[T] {
```
LuccaBitfly, line 364: **praise:** good change (feel free to resolve)

LuccaBitfly (changes requested):

backend/pkg/api/types/validator_dashboard.go
remoterami: Okay I'll do that once approved in the linked PR, then I can rebase this branch and make use of the approach

frontend/types/api/validator_dashboard.ts
remoterami: Correct, executer is not present yet but should prob be added. I think it's analog to the deposits table where we split the EL transaction into `From` and `Depositor`. Will further discuss this
`is_missing_estimate` shouldn't be there for the EL withdrawals endpoint, I'll remove it.
For CL it was present in the past already, so I kept it. After looking into it it seems like it was not displayed anywhere directly but used as a logic entity instead. By convention the first row (on the first page) would always be the static "estimation" row, and this flag would indicate if it's not available for some reason. Doesn't feel too clean.. Maybe this should also be changed :sweat_smile:

frontend/types/api/validator_dashboard.ts
LuccaBitfly: won't do

backend/pkg/api/types/validator_dashboard.go
remoterami: Done, ptal

frontend/types/api/validator_dashboard.ts
remoterami: `Executer` was renamed to `Withdrawer` & tx sender was (temporarily) removed, see design discussions for details. ptal

LuccaBitfly (approved):
lgtm

remoterami:
merged via #1480

PR #1425, opened 2025-03-25
feat(execution layer): logging and better architecture

Tangui-Bitfly:
* add proper structured logging (remove depguard on logrus)
* simplify `eth1indexer/main.go`
* create `executionlayer.IndexerService` to centralize all indexing tasks
* fix some issues with indexing

Tested indexer on mainnet syncing new blocks

### DIFF
--- a/backend/cmd/eth1indexer/main.go
+++ b/backend/cmd/eth1indexer/main.go
@@ -2,7 +2,6 @@ package eth1indexer
 
 import (
 	"context"
-	"encoding/json"
 	"flag"
 	"fmt"
 	"net/http"
@@ -13,11 +12,9 @@ import (
 	"github.com/gobitfly/beaconchain/pkg/commons/db"
 	"github.com/gobitfly/beaconchain/pkg/commons/db2"
 	"github.com/gobitfly/beaconchain/pkg/commons/db2/database"
-	"github.com/gobitfly/beaconchain/pkg/commons/erc20"
 	"github.com/gobitfly/beaconchain/pkg/commons/log"
 	"github.com/gobitfly/beaconchain/pkg/commons/metrics"
 	"github.com/gobitfly/beaconchain/pkg/commons/rpc"
-	"github.com/gobitfly/beaconchain/pkg/commons/services"
 	"github.com/gobitfly/beaconchain/pkg/commons/types"
 	"github.com/gobitfly/beaconchain/pkg/commons/utils"
 	"github.com/gobitfly/beaconchain/pkg/commons/version"
@@ -40,26 +37,22 @@ func Run() {
 
 	reorgDepth := fs.Uint64("reorg.depth", 20, "Lookback to check and handle chain reorgs")
 
-	concurrencyBlocks := fs.Uint64("blocks.concurrency", 30, "Concurrency to use when indexing blocks from erigon")
-	startBlocks := fs.Uint64("blocks.start", 0, "Block to start indexing")
-	endBlocks := fs.Uint64("blocks.end", 0, "Block to finish indexing")
-	bulkBlocks := fs.Uint64("blocks.bulk", 8000, "Maximum number of blocks to be processed before saving")
-	offsetBlocks := fs.Uint64("blocks.offset", 100, "Blocks offset")
-	checkBlocksGaps := fs.Bool("blocks.gaps", false, "Check for gaps in the blocks table")
-	checkBlocksGapsLookback := fs.Int("blocks.gaps.lookback", 1000000, "Lookback for gaps check of the blocks table")
+	bulk := fs.Uint64("bulk", 8000, "Maximum number of blocks to be processed before saving")
+	concurrency := fs.Uint64("concurrency", 30, "Concurrency to use when indexing blocks from erigon")
 	traceMode := fs.String("blocks.tracemode", "parity/geth", "Trace mode to use, can bei either 'parity', 'geth' or 'parity/geth' for both")
 
-	concurrencyData := fs.Uint64("data.concurrency", 30, "Concurrency to use when indexing data from bigtable")
-	startData := fs.Uint64("data.start", 0, "Block to start indexing")
-	endData := fs.Uint64("data.end", 0, "Block to finish indexing")
-	bulkData := fs.Uint64("data.bulk", 8000, "Maximum number of blocks to be processed before saving")
-	offsetData := fs.Uint64("data.offset", 1000, "Data offset")
+	startBlocks := fs.Uint64("resync.start", 0, "Block to start indexing")
+	endBlocks := fs.Uint64("resync.end", 0, "Block to finish indexing")
+	skipBlocks := fs.Bool("resync.skipBlocks", false, "Skip resync for blocks table")
+	skipData := fs.Bool("resync.skipData", false, "Skip resync for data table")
+
+	checkBlocksGaps := fs.Bool("blocks.gaps", false, "Check for gaps in the blocks table")
+	checkBlocksGapsLookback := fs.Int("blocks.gaps.lookback", 1000000, "Lookback for gaps check of the blocks table")
+
 	checkDataGaps := fs.Bool("data.gaps", false, "Check for gaps in the data table")
 	checkDataGapsLookback := fs.Int("data.gaps.lookback", 1000000, "Lookback for gaps check of the blocks table")
 
-	enableBalanceUpdater := fs.Bool("balances.enabled", false, "Enable balance update process")
-	enableFullBalanceUpdater := fs.Bool("balances.full.enabled", false, "Enable full balance update process")
-	balanceUpdaterBatchSize := fs.Int("balances.batch", 1000, "Batch size for balance updates")
+	balanceUpdaterBatchSize := fs.Int64("balances.batch", 1000, "Batch size for balance updates")
 
 	tokenPriceExport := fs.Bool("token.price.enabled", false, "Enable token export process")
 	tokenPriceExportList := fs.String("token.price.list", "", "Tokenlist path to use for the token price export")
@@ -141,18 +134,18 @@ func Run() {
 		log.Fatal(err, "erigon client creation error", 0)
 	}
 
-	chainId := strconv.FormatUint(utils.Config.Chain.ClConfig.DepositChainID, 10)
+	chainID := strconv.FormatUint(utils.Config.Chain.ClConfig.DepositChainID, 10)
 
-	nodeChainId, err := client.GetNativeClient().ChainID(context.Background())
+	nodeChainID, err := client.GetNativeClient().ChainID(context.Background())
 	if err != nil {
 		log.Fatal(err, "node chain id error", 0)
 	}
 
-	if nodeChainId.String() != chainId {
-		log.Fatal(fmt.Errorf("node chain id mismatch, wanted %v got %v", chainId, nodeChainId.String()), "", 0)
+	if nodeChainID.String() != chainID {
+		log.Fatal(fmt.Errorf("node chain id mismatch, wanted %v got %v", chainID, nodeChainID.String()), "", 0)
 	}
 
-	bt, err := db.InitBigtable(utils.Config.Bigtable.Project, utils.Config.Bigtable.Instance, chainId, utils.Config.RedisCacheEndpoint)
+	bt, err := db.InitBigtable(utils.Config.Bigtable.Project, utils.Config.Bigtable.Instance, chainID, utils.Config.RedisCacheEndpoint)
 	if err != nil {
 		log.Fatal(err, "error connecting to bigtable", 0)
 	}
@@ -164,7 +157,7 @@ func Run() {
 	}
 
 	cache := freecache.NewCache(100 * 1024 * 1024) // 100 MB limit
-	store := db2.NewStoreV1FromBigtable(bigtable, database.FreeCache{Cache: cache})
+	store := db2.NewStoreV1FromBigtable(bigtable, db2.CachedBalanceUpdates{RemoteCache: database.FreeCache{Cache: cache}})
 
 	batcherConfig := evm.BatcherConfig{
 		Limit: utils.Config.Indexer.BatchLimit,
@@ -173,61 +166,70 @@ func Run() {
 		parsed := common.HexToAddress(utils.Config.Indexer.MulticallAddresses)
 		batcherConfig.MulticallAddress = &parsed
 	}
-	batcher := evm.NewBatcher(nodeChainId, client.GetNativeClient(), batcherConfig)
-
-	if *tokenPriceExport {
-		go func() {
-			for {
-				tokenList, err := readTokenListFile(*tokenPriceExportList)
-				if err != nil {
-					log.Error(err, "error reading token list file", 0)
-				}
-				pricer := executionlayer.NewTokenPricer(
-					store,
-					chainId,
-					executionlayer.NewLlamaClient(),
-					tokenList,
-					batcher,
-				)
-				if err := pricer.UpdateTokens(); err != nil {
-					log.Error(err, "error updating tokens", 0)
-				}
-				time.Sleep(*tokenPriceExportFrequency)
-			}
-		}()
-	}
-
-	if *enableEnsUpdater {
-		importer := executionlayer.NewENSImporter(store, db2.NewENSStore(db.WriterDb), executionlayer.NewEnsContracts(client.GetNativeClient()))
-		go ImportEnsUpdatesLoop(chainId, importer, *ensBatchSize)
-	}
+	batcher := evm.NewBatcher(nodeChainID, client.GetNativeClient(), batcherConfig)
 
 	lastBlockStore := db2.NewCachedLastBlocks(database.Redis{Client: redisClient}, store)
-	indexer := executionlayer.NewIndexer(store, lastBlockStore, executionlayer.AllTransformers...)
-	balanceUpdater := executionlayer.NewBalanceUpdater(chainId, store, store, batcher)
-	reorgWatcher := executionlayer.NewReorgWatcher(client.GetNativeClient(), store, *reorgDepth, chainId, lastBlockStore)
+	indexer := executionlayer.NewIndexer(store, lastBlockStore, executionlayer.IndexerConfig{
+		Concurrency: *concurrency,
+		TraceMode:   *traceMode,
+	}, client, executionlayer.AllTransformers...)
+	balanceUpdater := executionlayer.NewBalanceUpdater(chainID, store, store, batcher)
+	reorgWatcher := executionlayer.NewReorgWatcher(client.GetNativeClient(), store, *reorgDepth, chainID, lastBlockStore)
 
-	if *enableFullBalanceUpdater {
-		ProcessBalanceUpdates(balanceUpdater, *balanceUpdaterBatchSize, -1)
-		return
+	pricer := executionlayer.NewTokenPricer(
+		store,
+		chainID,
+		executionlayer.NewLlamaClient(),
+		batcher,
+	)
+
+	var importer *executionlayer.ENSImporter
+	if *enableEnsUpdater {
+		importer = executionlayer.NewENSImporter(store, db2.NewENSStore(db.WriterDb), executionlayer.NewEnsContracts(client.GetNativeClient()))
 	}
 
+	start, end := uint64(0), uint64(0)
 	if *block != 0 {
-		if err := indexer.IndexNode(chainId, client, *block, *block, *concurrencyBlocks, *traceMode); err != nil {
-			log.Fatal(err, "error indexing from node", 0, map[string]interface{}{"block": *block, "concurrency": *concurrencyBlocks})
-		}
-		if err := indexer.IndexEvents(chainId, *block, *block, *concurrencyData); err != nil {
-			log.Fatal(err, "error indexing from bigtable", 0)
-		}
-		cache.Clear()
+		start, end = *block, *block
+	}
+	if startBlocks != nil {
+		start = *startBlocks
+	}
+	if endBlocks != nil {
+		end = *endBlocks
+	}
 
-		log.Infof("indexing of block %v completed", *block)
+	service := executionlayer.NewIndexerService(
+		nodeChainID.String(),
+		client.GetNativeClient(),
+		&balanceUpdater,
+		reorgWatcher,
+		lastBlockStore,
+		pricer,
+		indexer,
+		db2.CachedBalanceUpdates{RemoteCache: database.FreeCache{Cache: cache}},
+		store,
+		importer,
+		executionlayer.Config{
+			TokenPriceExportFrequency: *tokenPriceExportFrequency,
+			BalanceUpdaterBatchSize:   *balanceUpdaterBatchSize,
+			ENSImportBatchSize:        *ensBatchSize,
+			Bulk:                      *bulk,
+		},
+	)
+	if *tokenPriceExport {
+		go service.SyncTokenPrice(*tokenPriceExportList)
+	}
+
+	if end != 0 {
+		if err := service.SyncRange(start, end, *skipBlocks, *skipData); err != nil {
+			log.Fatal(err, "error indexing from node", 0, map[string]interface{}{"start": *startBlocks, "end": *endBlocks, "concurrency": *concurrency})
+		}
 		return
 	}
 
 	if *checkBlocksGaps {
 		_, _, _, err := bt.CheckForGapsInBlocksTable(*checkBlocksGapsLookback)
-
 		if err != nil {
 			log.Fatal(err, "error checking for gaps in blocks table", 0)
 		}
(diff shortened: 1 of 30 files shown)


backend/pkg/commons/db2/database/cache.go
```diff
@@ -49,6 +54,15 @@ func (m *MemCache) Get(ctx context.Context, key string) ([]byte, error) {
...
 
+func (m *MemCache) Clear(ctx context.Context, prefix string) error {
+	for v := range maps.Keys(m.values) {
+		if strings.HasPrefix(v, prefix) {
+			delete(m.values, v)
+		}
+	}
+	return nil
+}
+
```
mccreedy-bitfly, line 65: Can you provide some context/comments on these Clear functions, and why we have 3 different ones?

backend/pkg/executionlayer/executionlayer.go
```diff
@@ -0,0 +1,344 @@
...
+}
+
+func (config *Config) validate() {
+	if config.TokenPriceFrequency == 0 {
+		config.TokenPriceFrequency = defaultConfig.TokenPriceFrequency
+	}
+	if config.BlockFrequency == 0 {
+		config.BlockFrequency = defaultConfig.BlockFrequency
+	}
+}
```
mccreedy-bitfly: nit: I would expect a validate function to check values and return an error if they are invalid. This is doing something more like setting default values. Maybe it's even more of an `init()` function

mccreedy-bitfly (commented):
Please include a comment on how you've tested this

backend/pkg/commons/db2/database/cache.go
Tangui-Bitfly: It's a way to clear the cache. It's needed for example for balance indexing, we store in cache each address with a balance change in the block before writing the update to the db, if an address has two or more changes we don't re-write it to the db. But we need need to invalidate the cache once switch to a new block.

There is three different implementations of the `RemoteCache` interface (probably not well named): `MemCache`, `Redis` and `FreeCache`. Thus three functions to implement.

https://github.com/gobitfly/beaconchain/blob/dce3162c69f02740cf4ce413662726069c36f59e/backend/pkg/commons/db2/database/cache.go#L23

Monika-Bitfly: would be good to drop short comment in the code to explain that ^

backend/cmd/eth1indexer/main.go
```diff
@@ -40,26 +37,22 @@ func Run() {
...
-	concurrencyBlocks := fs.Uint64("blocks.concurrency", 30, "Concurrency to use when indexing blocks from erigon")
-	startBlocks := fs.Uint64("blocks.start", 0, "Block to start indexing")
-	endBlocks := fs.Uint64("blocks.end", 0, "Block to finish indexing")
-	bulkBlocks := fs.Uint64("blocks.bulk", 8000, "Maximum number of blocks to be processed before saving")
-	offsetBlocks := fs.Uint64("blocks.offset", 100, "Blocks offset")
+	bulk := fs.Uint64("bulk", 8000, "Maximum number of blocks to be processed before saving")
+	concurrency := fs.Uint64("concurrency", 30, "Concurrency to use when indexing blocks from erigon")
+	traceMode := fs.String("blocks.tracemode", "parity/geth", "Trace mode to use, can bei either 'parity', 'geth' or 'parity/geth' for both")
+
+	startBlocks := fs.Uint64("resync.start", 0, "Block to start indexing")
```
Monika-Bitfly, line 45: imo the old flags were more straightforward and easier to understand

backend/pkg/commons/db2/store_test.go
```diff
@@ -0,0 +1,42 @@
...
+import (
+	"context"
+	"testing"
+
+	"github.com/gobitfly/beaconchain/pkg/commons/db2/database"
+	"github.com/gobitfly/beaconchain/pkg/commons/db2/database/databasetest"
+	"github.com/gobitfly/beaconchain/pkg/commons/types"
+)
+
+func TestStoreV1(t *testing.T) {
```
Monika-Bitfly, line 12: This only tests the block range. Since there are more functions inside store.go, it would be good to improve test coverage by adding tests for the other functions as well

backend/pkg/commons/db2/store.go
```diff
@@ -191,6 +191,19 @@ func (store StoreV1) addIndexedBlockInMetadata(block IndexedBlock) error {
 	return nil
 }
 
+func (store StoreV1) CountBalanceUpdates(chainID string) (int64, error) {
+	key := fmt.Sprintf("%s:%s:", chainID, balanceKey)
+	rows, err := store.updates.Read(key, database.WithoutValue(), database.WithLimit(0))
```
Monika-Bitfly, line 196: why limit set to 0?

backend/cmd/eth1indexer/main.go
Tangui-Bitfly: I want to simplify our cmd flags. Yes the name is probably not the perfect one but we can't continue with all those almost similar flags.
The difference between those are too low to justify a flag:
- `startBlocks` and `startData`
- `concurrencyBlocks` and `concurrencyData`
- `bulkBlocks ` and `bulkData`
...

backend/pkg/commons/db2/store.go
Tangui-Bitfly: Limit 0 is no limit, it allows to override the default limit

backend/pkg/commons/db2/store.go
```diff
@@ -428,6 +441,15 @@ func (store StoreV1) GetLastBlockInBlocksTable(chainID string) (uint64, error) {
 	return lastBlock, nil
 }
 
+func (store StoreV1) CountEnsUpdates(chainID string) (int64, error) {
+	key := fmt.Sprintf("%s:ENS:V", chainID)
```
Monika-Bitfly, line 445: why not storing this `ENS` and other keys eg `BLOCK` in a variable, similar to how we store the `B` key for balance ?

backend/pkg/commons/log/log.go
```diff
@@ -146,3 +148,11 @@ func logErrorInfo(err error, callerSkip int, isWarning bool, additionalInfos ...
 }
 
 type Fields = logrus.Fields
+
+var Logger = logrus.StandardLogger()
+
+var Empty = func() *logrus.Entry {
```
Monika-Bitfly, line 154: why do we need this?

backend/pkg/executionlayer/executionlayer.go
```diff
@@ -0,0 +1,346 @@
...
+	service.muBalance.Lock()
+	defer service.muBalance.Unlock()
+
+	for {
+		logger := logger.WithFields(state.Fields())
+		start := time.Now()
+
+		total, err := service.store.CountBalanceUpdates(state.chainID)
+		if err != nil {
+			logger.WithField("error", err).Error("error while updating balances")
```
Monika-Bitfly:
```suggestion
			logger.WithField("error", err).Error("error while counting balance updates")
```

backend/pkg/executionlayer/executionlayer.go
```diff
@@ -0,0 +1,346 @@
...
+			continue
+		}
+		logger = logger.WithField("pending", total)
+		if total == 0 {
+			logger.Info("finished updating balances")
```
Monika-Bitfly, line 268: I think we're not updating balances here yet?

backend/pkg/executionlayer/executionlayer.go
```diff
@@ -0,0 +1,346 @@
...
+	service.muENS.Lock()
+	defer service.muENS.Unlock()
+
+	for {
+		logger := logger.WithFields(state.Fields())
+		start := time.Now()
+
+		total, err := service.store.CountEnsUpdates(state.chainID)
+		if err != nil {
+			logger.WithField("error", err).Error("error while importing ens")
```
Monika-Bitfly:
```suggestion
			logger.WithField("error", err).Error("error while counting total ens updates")
```

backend/pkg/executionlayer/executionlayer.go
```diff
@@ -0,0 +1,346 @@
...
+			continue
+		}
+		logger = logger.WithField("pending", total)
+		if total == 0 {
+			logger.Info("finished importing ens")
+			return
+		}
+		res, err := service.ensImporter.Import(state.chainID, service.config.ENSImportBatchSize)
+		if err != nil {
+			logger.WithField("error", err).Error("error while importing balances")
```
Monika-Bitfly:
```suggestion
			logger.WithField("error", err).Error("error while importing ens")
```

backend/pkg/executionlayer/executionlayer.go
```diff
@@ -0,0 +1,346 @@
...
+func NewStateReader(chainID string, client *ethclient.Client, lastBlockStore db2.LastBlocksStore) StateReader {
+	return StateReader{
+		chainID:        chainID,
+		client:         client,
+		lastBlockStore: lastBlockStore,
+	}
+}
+
+func (r StateReader) state() (syncState, error) {
+	lastBlock, err := r.client.BlockNumber(context.Background())
```
Monika-Bitfly, line 160: we should abstract this behind an interface

backend/pkg/executionlayer/executionlayer.go
```diff
@@ -0,0 +1,346 @@
...
+	if err != nil {
+		return syncState{}, fmt.Errorf("get chain head: %w", err)
```
Monika-Bitfly, line 162:
```suggestion
		return syncState{}, fmt.Errorf("get block number: %w", err)
```

backend/pkg/executionlayer/executionlayer.go
```diff
@@ -0,0 +1,346 @@
...
+		store:          store,
+		ensImporter:    ensImporter,
+		config:         config,
+	}
+}
+
+func (service *Indexer) FromHead(state syncState) error {
+	start := time.Now()
+	logger := logger.WithFields(logrus.Fields{
+		"blocksRange": fmt.Sprintf("%d-%d", state.blocks+1, state.node),
```
Monika-Bitfly, line 233: if the table doesn't contain any data, the value of state.blocks will be 0 meaning block 0 will not be indexed. Is that correct?

backend/pkg/executionlayer/executionlayer.go
```diff
@@ -0,0 +1,346 @@
...
+		"node":    state.node,
+		"blocks":  state.blocks,
+		"data":    state.data,
+	}
+}
+
+type IndexerConfig struct {
+	BalanceUpdaterBatchSize int64
+	ENSImportBatchSize      int64
+	Bulk                    uint64
```
Monika-Bitfly, line 200:
```suggestion
	BulkSize                    uint64
```

backend/pkg/executionlayer/executionlayer.go
```diff
@@ -0,0 +1,346 @@
...
+		chainID: r.chainID,
+		node:    lastBlock,
+		blocks:  lastBlockFromBlocksTable,
+		data:    lastBlockFromDataTable,
+	}, nil
+}
+
+type syncState struct {
+	chainID string
+	node    uint64
```
Monika-Bitfly, line 182:
```suggestion
	lastBlockFromNode    uint64
```

backend/pkg/executionlayer/executionlayer.go
```diff
@@ -0,0 +1,346 @@
...
+	blocks  uint64
```
Monika-Bitfly, line 183:
```suggestion
	lastBlockFromBlocksTable  uint64
```

backend/pkg/executionlayer/executionlayer.go
```diff
@@ -0,0 +1,346 @@
...
+	data    uint64
```
Monika-Bitfly, line 184:
```suggestion
	lastBlockFromDataTable    uint64
```

backend/pkg/executionlayer/executionlayer.go
```diff
@@ -0,0 +1,346 @@
...
+	tokenPricer  *TokenPricer
+	indexer      *Indexer
+	stateReader  StateReader
+	reorgWatcher *ReorgWatcher
+
+	config Config
+}
+
+func NewIndexerService(stateReader StateReader, indexer *Indexer, reorgWatcher *ReorgWatcher, tokenPricer *TokenPricer, config Config) *IndexerService {
+	config.init()
```
Monika-Bitfly: We don't need to initialize the config again since it's already being initialized inside NewIndexer and passed to NewIndexerService

backend/pkg/executionlayer/executionlayer.go
```diff
@@ -0,0 +1,346 @@
...
+		config.TokenPriceFrequency = defaultConfig.TokenPriceFrequency
+	}
+	if config.BlockFrequency == 0 {
+		config.BlockFrequency = defaultConfig.BlockFrequency
+	}
+}
+
+type IndexerService struct {
+	tokenPricer  *TokenPricer
+	indexer      *Indexer
```
Monika-Bitfly, line 34: to reduce coupling between IndexerService and Indexer we can introduce an interface eg

```
type BlockIndexer interface {  
    FromHead(state syncState) error  
    Balances(state syncState)  
    ENS(state syncState)  
    Range(state syncState, start, end uint64, skipNode, skipData bool) error  
}  
```

backend/pkg/executionlayer/executionlayer.go
```diff
@@ -0,0 +1,346 @@
...
+func (service *IndexerService) SyncLive() {
+	for ; ; time.Sleep(service.config.BlockFrequency) {
+		state, err := service.stateReader.state()
+		if err != nil {
+			logger.WithField("error", err).Error("cannot get state")
+			continue
+		}
+		logger.WithFields(state.Fields()).Info("last blocks")
+
+		if err := service.reorgWatcher.LookForReorg(); err != nil {
```
Monika-Bitfly, line 105: it might be a good idea to use separate goroutines for `LookForReorg` and `FromHead` to improve concurrency

backend/pkg/executionlayer/ens.go
```diff
@@ -41,89 +40,108 @@ type ENSImporter struct {
...
 
 		for _, ensLog := range batch {
 			g.Go(func() error {
-				return importer.importENS(ensLog, checked)
+				res, err := importer.importENS(ensLog, checked)
+				if err != nil {
+					return err
+				}
+				mu.Lock()
+				all.updated = append(all.updated, res.updated...)
```
Monika-Bitfly, line 85: Do we need to log all the ENS updates? Could there be cases where logging hundreds of ENS updates might  clutter the logs?

backend/pkg/executionlayer/blockindexer.go
```diff
@@ -23,24 +23,49 @@ type Store interface {
...
 }
 
-type Indexer struct {
+type BlockIndexerConfig struct {
+	Concurrency uint64
+	TraceMode   string
+}
+
+var defaultBlockIndexerConfig = BlockIndexerConfig{
+	Concurrency: 30,
```
Monika-Bitfly: imo this value might be too high for some cases, would be better to set the default concurrency based on system resources

backend/pkg/executionlayer/blockindexer.go
```diff
@@ -147,17 +170,17 @@ func (indexer *Indexer) IndexEvents(chainID string, start, end uint64, concurren
 }
 
 // Index retrieve the types.Eth1Block from the client and apply the transformers to them
-func (indexer *Indexer) Index(chainID string, client Client, start, end uint64, concurrency uint64, traceMode string) error {
-	if err := indexer.IndexNode(chainID, client, start, end, concurrency, traceMode); err != nil {
-		return err
+func (indexer *BlockIndexer) Index(chainID string, start, end uint64) error {
+	if err := indexer.IndexNode(chainID, start, end); err != nil {
+		return fmt.Errorf("blocks from node: %w", err)
```
Monika-Bitfly, line 159:
```suggestion
		return fmt.Errorf("error while indexing block range %d - %d from node: %w", start, end, err)
```

backend/pkg/executionlayer/blockindexer.go
```diff
@@ -147,17 +170,17 @@ func (indexer *Indexer) IndexEvents(chainID string, start, end uint64, concurren
...
 	}
-	if err := indexer.IndexEvents(chainID, start, end, concurrency); err != nil {
-		return err
+	if err := indexer.IndexEvents(chainID, start, end); err != nil {
+		return fmt.Errorf("events: %w", err)
```
Monika-Bitfly, line 162:
```suggestion
		return fmt.Errorf("error while indexing events for block range %d - %d: %w", start, end, err)
```

backend/pkg/commons/db2/store.go
Tangui-Bitfly: `B` was not very descriptive, `balanceKey` was better.
`ENS` is already descriptive, I'm fine with replacing by `ENSKey` that was just not the main purpose of this code change.

backend/pkg/executionlayer/balance_test.go
```diff
@@ -41,9 +41,9 @@ func TestBalanceUpdater(t *testing.T) {
 		Value: expectedBalance,
```
Monika-Bitfly, line 41: Value is currently unused, should update [this line](https://github.com/gobitfly/beaconchain/blob/5d61711b609538659d556b873f2941186601b0eb/backend/pkg/executionlayer/balance_test.go#L56) to:

```
if got, want := balances[0].Value.String(), expected.Value.String(); got != want {
		t.Errorf("got %v, want %v", got, want)
}
```

backend/pkg/commons/db2/store_test.go
Tangui-Bitfly: Yes that would be good, but if we ditch bigtable in the near future I don't want to invest time in it.
Moreover those are functions that basically store and retrieve data, from my point of view testing business logic had a higher priority.

backend/pkg/commons/log/log.go
Tangui-Bitfly: This not used, true.
It's useful for removing noise in test. If you run a test with `-v` the result will print the logs.
`log.Logger = log.Empty` removes that noise.

backend/cmd/eth1indexer/main.go
```diff
@@ -242,175 +247,5 @@ func Run() {
...
-		log.InfoWithFields(log.Fields{
-			"node":   lastBlockFromNode,
-			"blocks": lastBlockFromBlocksTable,
-			"data":   lastBlockFromDataTable,
-		}, "last blocks")
-
-		continueAfterError := false
-		if lastBlockFromNode > 0 {
-			if lastBlockFromBlocksTable < lastBlockFromNode {
-				log.Infof("missing blocks %v to %v in blocks table, indexing ...", lastBlockFromBlocksTable+1, lastBlockFromNode)
```
Monika-Bitfly, line 294: this log was pretty useful imo, can we bring it back?

backend/pkg/executionlayer/blockindexer.go
Tangui-Bitfly: That's the current default
https://github.com/gobitfly/beaconchain/blob/staging/backend/cmd/eth1indexer/main.go#L43

backend/cmd/eth1indexer/main.go
```diff
@@ -242,175 +247,5 @@ func Run() {
...
-				}
-			}
-		}
-
-		if *enableBalanceUpdater {
-			ProcessBalanceUpdates(balanceUpdater, *balanceUpdaterBatchSize, 10)
-		}
-
-		log.Infof("index run completed")
-		services.ReportStatus("eth1indexer", "Running", nil)
```
Monika-Bitfly, line 375: missing this status report

backend/pkg/executionlayer/blockindexer.go
Tangui-Bitfly: Start and end are known by the caller, adding them to the error message here would risk them to be present two times in the error message.
Here for example they resulting log message would be
```
"level"="error" "startBlock"=100 "endBlock"=101 "message"="indexing from head" "error"="blocks from node: XXXX"
```
we don't need to repeat ourself with `error ...` or start and end

backend/pkg/executionlayer/blockindexer.go
Tangui-Bitfly: same

backend/pkg/executionlayer/blockindexer.go
```diff
@@ -23,24 +23,49 @@ type Store interface {
...
 		lastBlockStore: lastBlockStore,
 		transformers:   transformers,
+		client:         client,
+		config:         config,
 	}
 }
 
 // IndexNode retrieve types.Eth1Block from the client and save them into the store
-func (indexer *Indexer) IndexNode(chainID string, client Client, start, end uint64, concurrency uint64, traceMode string) error {
+func (indexer *BlockIndexer) IndexNode(chainID string, start, end uint64) error {
```
Monika-Bitfly, line 50: would be great to look into implementing retry mechanism for handling errors, rather than aborting the entire batch when an error occurs

backend/pkg/executionlayer/blockindexer.go
```diff
@@ -88,14 +113,13 @@ func (indexer *Indexer) IndexNode(chainID string, client Client, start, end uint
...
 // IndexEvents retrieve read the types.Eth1Block from the store and apply the transformers to them
-func (indexer *Indexer) IndexEvents(chainID string, start, end uint64, concurrency uint64) error {
+func (indexer *BlockIndexer) IndexEvents(chainID string, start, end uint64) error {
 	retrieval := new(errgroup.Group)
-	retrieval.SetLimit(int(concurrency))
+	retrieval.SetLimit(int(indexer.config.Concurrency))
 
 	indexing := new(errgroup.Group)
-	indexing.SetLimit(int(concurrency * concurrency))
+	indexing.SetLimit(int(indexer.config.Concurrency * indexer.config.Concurrency))
```
Monika-Bitfly, line 105: this could be risky, if values are set too high it might nuke the system

backend/pkg/executionlayer/ens.go
Tangui-Bitfly: I don't think that is concern since blocks have a gas limit. We cannot have a "spam" attack of ens updates.
But I see your point, why would we need it ? I choose to add it because we already had the deleted ens logged.

backend/pkg/executionlayer/executionlayer.go
Tangui-Bitfly: I prefer to introduce abstraction when it's needed.

backend/pkg/executionlayer/executionlayer.go
Tangui-Bitfly: `NewIndexer` use type `IndexerConfig` for config and `IndexerService` use type `Config`.
It's a bit confusing I agree I will change that.

backend/pkg/executionlayer/executionlayer.go
Tangui-Bitfly: Yes and no, I do agree we could have a better indexing flow, but I tried to stay as close as possible to the current version.
In the separate PR in the future, definitely.

backend/pkg/executionlayer/executionlayer.go
Tangui-Bitfly: IMO `syncState.node` is more readable than `syncState.lastBlockFromNode` for fewer characters.

backend/pkg/executionlayer/executionlayer.go
Tangui-Bitfly: same

backend/pkg/executionlayer/executionlayer.go
Tangui-Bitfly: same

backend/pkg/executionlayer/executionlayer.go
Tangui-Bitfly: correct, I will add proper handling of the genesis block

backend/cmd/eth1indexer/main.go
Tangui-Bitfly: It was replaced by this log
https://github.com/gobitfly/beaconchain/blob/5d61711b609538659d556b873f2941186601b0eb/backend/pkg/executionlayer/executionlayer.go#L77

backend/pkg/executionlayer/blockindexer.go
Tangui-Bitfly: We already talked about it in an old PR
https://github.com/gobitfly/beaconchain/pull/1339#discussion_r1979817920
This is just the migration of old (and working) logic but more explicit. Yes it's probably not great, but if it's working I don't want to rework it for now. But at least now it's not hidden in the code.

backend/pkg/executionlayer/blockindexer.go
Tangui-Bitfly: Yep that would be a great new feature to work later on.

backend/pkg/executionlayer/executionlayer.go
Tangui-Bitfly: see for fix https://github.com/gobitfly/beaconchain/pull/1452

invis-bitfly (approved):
lgtm™

PR #1426, opened 2025-03-25
feat(executionlayer): metrics

Tangui-Bitfly:
* prometheus metrics
* refactor transformer to allow higher level metric
This PR allows such dashboard
image

### DIFF
--- a/backend/pkg/commons/metrics/executionlayer.go
+++ b/backend/pkg/commons/metrics/executionlayer.go
@@ -1,0 +1,38 @@
+package metrics
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+)
+
+var (
+	IndexingBlockDifference = promauto.NewGaugeVec(prometheus.GaugeOpts{
+		Name: "indexing_block_difference",
+		Help: "Difference between the latest on-chain block and the last indexed block",
+	}, []string{"chainID"})
+
+	IndexingTime = promauto.NewHistogramVec(prometheus.HistogramOpts{
+		Name: "indexing_time_seconds",
+		Help: "Time taken to index the new blocks",
+	}, []string{"chainID"})
+
+	IndexingTransformerProcessingTime = promauto.NewHistogramVec(prometheus.HistogramOpts{
+		Name: "indexing_transformer_processing_time_seconds",
+		Help: "Time taken by a transformer to index a block",
+	}, []string{"chainID", "transformer"})
+
+	IndexingPendingBalanceUpdate = promauto.NewGaugeVec(prometheus.GaugeOpts{
+		Name: "indexing_pending_balance_update",
+		Help: "Pending balances to update",
+	}, []string{"chainID"})
+
+	IndexingPendingENSUpdate = promauto.NewGaugeVec(prometheus.GaugeOpts{
+		Name: "indexing_pending_ens_update",
+		Help: "Pending ENS to update",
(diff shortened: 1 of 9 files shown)


backend/pkg/commons/metrics/executionlayer.go
```diff
@@ -0,0 +1,38 @@
...
+
+var (
+	IndexingBlockDifference = promauto.NewGaugeVec(prometheus.GaugeOpts{
+		Name: "indexing_block_difference",
+		Help: "Difference between the latest on-chain block and the last indexed block",
+	}, []string{"chainID"})
+
+	IndexingTime = promauto.NewHistogramVec(prometheus.HistogramOpts{
+		Name: "indexing_time_seconds",
+		Help: "Time taken to index the new blocks since last tick",
```
mccreedy-bitfly: What's a tick? Can we use terminology which is more obvious about what is happening

backend/pkg/executionlayer/executionlayer.go
```diff
@@ -241,9 +255,11 @@ func (service *Indexer) FromHead(state syncState) error {
 			"elapsed": time.Since(start),
 		}).Info("indexed blocks")
 	}
+	end := time.Since(start) // save end to unify log and metric value
+	indexingMetrics.Time(state.chainID, end)
```
Monika-Bitfly: bad naming, `Time` doesn't provide enough context to understand what it does

backend/pkg/executionlayer/transformer.go
```diff
@@ -29,23 +29,73 @@ import (
...
-	"TransformBlock":             TransformBlock,
-	"TransformBlobTx":            TransformBlob,
-	"TransformContract":          TransformContract,
-	"TransformItx":               TransformITx,
-	"TransformERC721":            TransformERC721,
-	"TransformERC1155":           TransformERC1155,
-	"TransformUncle":             TransformUncle,
-	"TransformWithdrawals":       TransformWithdrawal,
-	"TransformEnsNameRegistered": TransformEnsNameRegistered,
+type Transformer struct {
```
Monika-Bitfly, line 31: These changes should be in a separate PR since it's not related to metrics

backend/pkg/executionlayer/reorg.go
```diff
@@ -42,10 +42,10 @@ func NewReorgWatcher(client EthClient, store ReorgStore, depth uint64, chainID s
 	}
 }
 
-func (r *ReorgWatcher) LookForReorg() error {
+func (r *ReorgWatcher) LookForReorg() (uint64, error) {
```
Monika-Bitfly, line 49: same here

backend/pkg/executionlayer/transformer.go
Tangui-Bitfly: This change is related to metrics.
In the previous version, we don't have access to the name of the transformer applied. This prevent us to make good metrics. This refactoring is needed inside this pr.

backend/pkg/executionlayer/reorg.go
Tangui-Bitfly: Here we don't want to add complexity to the already quite complex function `LookForReorg`.
I prefer to have the functions report their execution and letting their caller handle the logging / metrics.

Monika-Bitfly:
Would also be good to add database and client metrics to track their durations

mccreedy-bitfly (approved):

invis-bitfly (approved):
lgtm™

PR #1432, opened 2025-03-26
Beds 97/fix notifications table age button

marcel-bitfly:


frontend/assets/css/main.scss
```diff
@@ -43,8 +43,8 @@ a {
 }
 
 .link {
-  color: var(--link-color);
   cursor: pointer;
+  color: var(--link-color) !important;
```
marcel-bitfly, line 47: **thought:** we should never use `!important` but I see no other quick way to solve the `specificity issue` in this case

A global class applying a color that is less specific than `bc-button-icon` as a `scoped style`

enzo-bitfly (approved):

PR #1433, opened 2025-03-26
feat(slot_exporter) add metrics

Monika-Bitfly:
Changes:
- introduced Metrics Repository
- added task duration metrics for slot and epoch export 
- added client call duration and error metrics for lighthouse and node client


backend/pkg/consapi/client_node.go
```diff
@@ -36,57 +37,135 @@ func NewClientWithConfig(endpoint string, httpClient *http.Client) Client {
...
+	}(timeStart)
+
 	requestURL := fmt.Sprintf("%s/eth/v1/beacon/states/%v/validator_balances", r.Endpoint, stateID)
-	return network.Get[types.StandardValidatorBalancesResponse](r.httpClient, requestURL)
+	resp, err := network.Get[types.StandardValidatorBalancesResponse](r.httpClient, requestURL)
+	if err != nil {
+		r.metrics.Error("node_get_validator_balances")
+		return nil, err
+	}
+	return resp, nil
```
Tangui-Bitfly: We could use a wrapper, here it adds complexity. And maybe in the future we want a `NodeClient` without metrics

```go
// implements ClientInt
type NodeClientWithMetrics struct {
	client ClientInt
	metrics metrics.MetricsRepository
}

func (c *NodeClient) GetValidatorBalances(stateID any) (*types.StandardValidatorBalancesResponse, error) {
	timeStart := time.Now()
	defer func(timeStart time.Time) {
		r.metrics.ObserveClientCallDuration("node", "get_validator_balances", time.Since(timeStart))
	}(timeStart)
	resp, err := c.client.GetValidatorBalances(stateID)
	if err != nil {
		r.metrics.Error("node_get_validator_balances")
		return nil, err
	}
	return resp, nil
}
// and so on ...
```

backend/pkg/commons/rpc/lighthouse.go
```diff
@@ -39,6 +40,7 @@ type LighthouseClient struct {
 	slotsCache          *lru.Cache
 	slotsCacheMux       *sync.Mutex
 	signer              gethtypes.Signer
+	metrics             metrics.MetricsRepository
```
Tangui-Bitfly: Same comment, a wrapper would be better

backend/pkg/exporter/db/bigtable.go
```diff
@@ -17,29 +18,51 @@ type SlotExporterBTRepository interface {
...
+		client:  btClient,
+		metrics: metrics,
 	}
 }
 
 func (s *SlotExporterBT) GetValidatorBalanceHistory(validators []uint64, startEpoch uint64, endEpoch uint64) (map[uint64][]*types.ValidatorBalance, error) {
-	return s.client.GetValidatorBalanceHistory(validators, startEpoch, endEpoch)
+	balanceHistory, err := s.client.GetValidatorBalanceHistory(validators, startEpoch, endEpoch)
+	if err != nil {
+		s.metrics.Error("slot_exporter_get_validator_balance_history")
```
Tangui-Bitfly: I'm not sure it's the good way of handling error metrics for db interaction.
I don't have necessarily a better solution but this seems too specific.
We basically only read and write to the db, maybe that should be the scope for our errors metrics.

backend/pkg/exporter/db/cache.go
```diff
@@ -23,26 +24,35 @@ type SlotExporterCacheRepository interface {
...
 }
 
 func (c *SlotExporterCache) setValue(key string, value []byte, expiration time.Duration) error {
 	ctx, cancel := context.WithTimeout(context.Background(), time.Second*30)
 	defer cancel()
-	return c.cache.Set(ctx, key, value, expiration)
+
+	err := c.cache.Set(ctx, key, value, expiration)
+	if err != nil {
+		c.metrics.Error("slot_exporter_cache_set_value")
```
Tangui-Bitfly: Same, too specific. We could put those metric one level down, inside a `database.RemoteCacheWithMetrics`

Tangui-Bitfly (commented):
We need to have a discussion on how to report error metrics.
I'm little concerned by our forever growing structs and enormous line count.

PR #1440, opened 2025-03-27
Beds 1362/el deposits table pectra

enzo-bitfly:
Based on #1414


marcel-bitfly:
**nitpick**: commit message grammar

```
refactor: rename <BcTable> prop to `is-loading`
```

**nitpick**: commit message form

```
feat: adapt EL deposits table for Pectra
<Blank>
Move fetching logic to `dashboard` root page component
<Blank>
See: BEDS-1362
```

Both would be important for `changelog` automation

frontend/components/bc/table/BcTable.vue
marcel-bitfly, line 1: **praise:** nice catch thx 💪

frontend/components/bc/BcBadge.vue
```diff
@@ -0,0 +1,54 @@
+<script setup lang="ts">
+export type BadgeColor = 'gray' | 'green' | 'orange' | 'red'
+
+const {
+  color,
+  width = 'min-content',
```
marcel-bitfly: **issue (blocking):** your `type definition` and your `default value` do not match

(this should error and might in the future, but I guess `typescript` interprets your `with prop` as `string` for now)

frontend/components/bc/BcBadge.vue
```diff
@@ -0,0 +1,54 @@
...
+} = defineProps<{
+  color: BadgeColor,
+  width?: `${number}px` | `${number}rem`,
```
marcel-bitfly: **issue:** the user should not be able to put in `px values` as this is not `accessible`

I think `BcBadge` should now even need a `with prop` but should be as big as the text.
You also do not handle the case if the `text` is longer than the provided text (e.g. `ellipsis`)

frontend/components/bc/BcBadge.vue
```diff
@@ -0,0 +1,54 @@
...
+  &--gray {
+    background-color: var(--grey);
+  }
+
+  &--green {
+      background-color: var(--positive-color);
+  }
+
+  &--orange {
+    background-color: var(--partial-color);
```
marcel-bitfly: **question (blocking):** what's `--partial-color` 😅

frontend/components/bc/tab/BcTabList.vue
```diff
@@ -12,6 +12,17 @@ const props = defineProps<Props>()
...
+const emit = defineEmits<{
+  (e: 'changed-tab', value: string): void,
+}>()
+
+watch(
+  activeTab,
+  (value) => {
+    emit('changed-tab', value)
+  },
+)
```
marcel-bitfly: **issue (blocking):** `watchers` are for `side effects`

`v-model` is already emitting an event

```vue
...
const hello = ($event) => {
console.log("currentTabName:", $event)
}
...
<Tabs
    v-model:value="activeTab"
    lazy
    scrollable
    class="dashboard-tab-view"
    @update:value="hello"
  >

```

marcel-bitfly: **suggestion**: do a [defineModel](https://vuejs.org/guide/components/v-model.html#basic-usage) in `BcTabList` and use `v-model` in the parent

frontend/components/dashboard/table/DashboardTableElDeposits.vue
```diff
@@ -124,56 +102,72 @@ const {
   displayCurrencyDefault,
   selectedCurrencyMain,
 } = useCurrency()
+
+onMounted(() => {
```
marcel-bitfly: **todo (blocking):** remove `debug code` plz

frontend/components/dashboard/table/DashboardTableElDeposits.vue
```diff
@@ -124,56 +102,72 @@ const {
...
 
 <template>
   <div>
-    <BcTableControl :title="$t('dashboard.validator.el_deposits.title')">
+    <BcTableControl
+      :title="$t('dashboard.validator.el_deposits.title')"
+      :search-placeholder="
+        $t(
+          isGuestDashboard
+            ? 'dashboard.validator.el_deposits.search_placeholder_public'
```
marcel-bitfly: **suggestion (non-blocking):** be precise in both cases

```
'dashboard.validator.el_deposits.search_placeholder_public_dashboard'
'dashboard.validator.el_deposits.search_placeholder_guest_dashboard'
```

frontend/components/dashboard/table/DashboardTableElDeposits.vue
```diff
@@ -124,56 +102,72 @@ const {
...
             @set-cursor="setCursor"
             @sort="onSort"
             @set-page-size="setPageSize"
           >
             <Column
-              v-if="colsVisible.publicKey"
-              field="public_key"
-              :header="$t('dashboard.validator.col.public_key')"
+              field="timestamp"
+              body-class="age-field"
```
marcel-bitfly: **question:** is this needed?

frontend/components/dashboard/table/DashboardTableElDeposits.vue
```diff
@@ -250,47 +216,48 @@ const {
...
-                  type="withdrawal_credentials"
-                />
+                  class="status-cell-content"
+                >
+                  <BcBadge
+                    :color="getStatusColor(slotProps.data.validity)"
+                    width="75px"
+                  >
+                    {{ getStatusText(slotProps.data.validity) }}
+                  </BcBadge>
```
marcel-bitfly: **issue (blocking):** logic that is part of the component is in a composable

- useStatus() does not return `stateful logic` -> no composable
- logic is only used once 
- useStatus does not give you status from anything else than a `dashboard.validator.el_deposits.validity`

**suggestion**: put the logic in a component
**suggestion**: when `BcBadge` should be generic, wrap `BcBadge` in another component solving your `el deposits issues`

frontend/components/dashboard/table/DashboardTableElDeposits.vue
```diff
@@ -311,38 +278,75 @@ const {
...
                     {{ groupNameLabel(slotProps.data.group_id) }}
                   </div>
                 </div>
+
                 <div class="row">
+                  <div class="label">
+                    {{ $t("block.col.transaction_hash") }}
+                  </div>
+                  <BcFormatHash
+                    v-if="slotProps.data.index !== undefined"
```
marcel-bitfly: **polish:** v-if already checks for `falsiness`
(index will probably not be `negative` or `0`)
```v-if="slotProps.data.index"```

File: frontend/components/dashboard/table/DashboardTableElDeposits.vue [hunk 10]
```diff
@@ -421,9 +410,10 @@ const {
 
 <style lang="scss" scoped>
 @use "~/assets/css/utils.scss";
+@use '~/assets/css/breakpoints' as *;
 
 :deep(.el_deposits_table) {
-  > .p-datatable-wrapper {
+  >.p-datatable-wrapper {
```
marcel-bitfly: **nitpick:** whitespace (this can happen due to your workflow hack with your `ide`)

frontend/components/dashboard/table/DashboardTableElDeposits.vue
```diff
@@ -448,34 +438,55 @@ const {
...
     padding: 0 7px;
     @include utils.set-all-width(110px);
   }
 }
 
 .expansion {
-  color: var(--container-color);
+  display: grid;
+  grid-template-columns: repeat(2, max-content);
+  grid-template-rows: unset;
```
marcel-bitfly: **question:** did you try leaving this out?

frontend/composables/useApi.ts
marcel-bitfly, line 1: **issue:** this is no composable as it does simply a wrapper for specific fetch functions 

I get how you want to make our codebase cleaner this way, but this introduces another pattern, which is similar to a `repository pattern` or `service layer pattern` which will not solve our actual problem at the moment

But very nice that you try stuff out, that helps 👍

frontend/composables/useStatus.ts
```diff
@@ -0,0 +1,41 @@
...
+  }
+  const statusNameMap: Record<Status, string> = {
+    invalid: $t('dashboard.validator.el_deposits.validity.invalid'),
+    invalid_skipped: $t('dashboard.validator.el_deposits.validity.valid'),
+    valid: $t('dashboard.validator.el_deposits.validity.valid'),
+  }
+  const statusIconMap: Record<StatusColor, StatusIcon> = {
+    green: 'check',
+    red: 'xmark',
+  }
```
marcel-bitfly: **thought:** implementation details that are hard to generalize and `deeply couples` ourselves to the icon names which have nothing todo with `useStatus()`

frontend/i18n/locales/en.json
```diff
@@ -4,7 +4,7 @@
         "col": {
             "graffiti": "Graffiti",
             "proposer": "Proposer",
-            "tx_hash": "Tx Hash"
+            "transaction_hash": "Transaction Hash"
```
marcel-bitfly, line 7: **praise:** makes it easier to read ❤️

frontend/i18n/locales/en.json
```diff
@@ -259,7 +260,15 @@
...
             },
             "el_deposits": {
-                "title": "Execution Layer"
+                "info_valid_with_invalid_signature": "Signature invalid, but will still be accepted due to a prior deposit with a valid signature.",
+                "search_placeholder": "Index, Block, Address, Group",
+                "search_placeholder_public": "Index, Block, Address",
+                "title": "Execution Layer",
+                "validity": {
+                    "invalid": "Invalid",
+                    "screenreader_text": "Validity is: {validity}",
```
marcel-bitfly: **thought:** this is repetitive

Maybe telling the `sr user` what is valid or invalid would be of better help.
Eg: `Signature invalid`

frontend/pages/dashboard/[[id]]/index.vue
```diff
@@ -199,6 +200,51 @@ watch(
   },
   { immediate: true },
 )
+
+const api = useApi()
```
Comment (marcel-bitfly) [inline frontend/pages/dashboard/[[id]]/index.vue]: **thought:** fyi this is how `useCustomFetch` should have been implemented https://nuxt.com/docs/guide/recipes/custom-usefetch#custom-usefetchuseasyncdata

frontend/composables/useCustomFetch.ts
marcel-bitfly, line 1: **chore (if-minor):** then you could also remove the whole `public/mock` directory I guess 😅

frontend/.vscode/settings.json
```diff
@@ -65,6 +65,7 @@
         "useFeatureFlag",
         "useNotificationsOverviewStore",
         "useWindowSize",
+        "variables",
```
marcel-bitfly: **chore:** I already introduced `main.scss` to be specific

I think we should change your scope and add `.scss` as well, what do you think?

enzo-bitfly: I missed that, I thought `.scss` would be better too. Changed

frontend/components/bc/BcBadge.vue
enzo-bitfly: Done

frontend/components/bc/BcBadge.vue
enzo-bitfly: Removed `px`.

Using `width` comes from a design specification to have a fixed width for the `status badges` so they all have the same width

frontend/components/bc/BcBadge.vue
enzo-bitfly: I changed it for `--orange`

frontend/components/bc/tab/BcTabList.vue
enzo-bitfly: I changed this but to use the route `hash`.
thought: as mentioned before, I will change this to use `router query params` as soon as I'm done with all the tables.

frontend/components/dashboard/table/DashboardTableElDeposits.vue
enzo-bitfly: Done

frontend/components/dashboard/table/DashboardTableElDeposits.vue
enzo-bitfly: Chaned to `'dashboard.validator.el_deposits.search_placeholder_guest_dashboard'` and `'dashboard.validator.el_deposits.search_placeholder_private_dashboard'`

frontend/components/dashboard/table/DashboardTableElDeposits.vue
enzo-bitfly: I only moved this column. But yes, the date can break lines otherwise.

frontend/components/dashboard/table/DashboardTableElDeposits.vue
enzo-bitfly: - This is a composable because it uses another composable (`useTranslation`)
- It will be reused in the next tables
- The next tables have different status (`queued`, `completed`, `rejected`, `processed`), which I will add here as needed

I had decided to have it as a `composable` to prevent having specific types in the `base components`, **but I created `BcStatusBadge` now to follow your suggestion**

frontend/components/dashboard/table/DashboardTableElDeposits.vue
enzo-bitfly: Index can be 0 (for validator 0). Basically this checks if the `index` property is not there, meaning we're on the first row, which shows the total amount.

**I changed this now by adding a `isTotaAmountRow` flag to the first row.**

frontend/components/dashboard/table/DashboardTableElDeposits.vue
enzo-bitfly: Yep. Fixed

frontend/components/dashboard/table/DashboardTableElDeposits.vue
enzo-bitfly: Good catch, I removed it

frontend/composables/useStatus.ts
enzo-bitfly: Well this composable was meant to be an interface between the `status` we get from the api and the way we want to show them in the `badge`. Now it all belongs to `BcStatusBadge.vue` and this composable is gone.

frontend/i18n/locales/en.json
enzo-bitfly: I changed it to `Deposit Invalid`

frontend/pages/dashboard/[[id]]/index.vue
Comment (enzo-bitfly) [reply frontend/pages/dashboard/[[id]]/index.vue]: Nice, we'll do that.

frontend/composables/useCustomFetch.ts
enzo-bitfly: I cannot quickly estimate right now if that would involve changing something else, but good to know.

enzo-bitfly (commented):
Thank you for the thorough review 💪

frontend/composables/useCustomFetch.ts
marcel-bitfly: Wait I remember now that this`flag` was introduced recently by @LuccaBitfly and has nothing to do with the `mock stuff in frontend` as I thought before.
It is possible to get Mocked data, which was helping during development before.

@LuccaBitfly I guess this still works and might come in handy in the future?

Also this is connected for example: https://github.com/gobitfly/beaconchain/blob/BEDS-1362/el-deposits-table-pectra/frontend/package.json#L52

frontend/assets/css/variables.scss
```diff
@@ -4,13 +4,13 @@
...
-  --padding-large: 20px;
-  --padding-medium: 15px;
-  --padding: 10px;
-  --padding-small: 5px;
-  --padding-tiny: 2px;
-  --padding-line: 1px;
+  --padding-xl: 2rem;
+  --padding-large: 1.25rem;
+  --padding-medium: 1rem;
+  --padding: 0.5rem;
```
marcel-bitfly, line 10: **question (blocking):** did you check visually if something looks off now? And did you asked philip if this was ok?

As for example `10px` !== `0.5rem` you are changing the `design system` which might introduce inconsistencies

frontend/components/bc/BcBadge.vue
```diff
@@ -0,0 +1,54 @@
+<script setup lang="ts">
+export type BadgeColor = 'gray' | 'green' | 'orange' | 'red'
+
+const {
+  color,
+  width = 'min-content',
+} = defineProps<{
+  color: BadgeColor,
+  width?: 'min-content' | `${number}px` | `${number}rem`,
```
marcel-bitfly: **issue:** still `px` values possible

See: https://github.com/gobitfly/beaconchain/pull/1440#discussion_r2016347897

frontend/components/bc/BcStatusBadge.vue
```diff
@@ -0,0 +1,61 @@
...
+  >
+    <span v-if="iconOnly">
+      <BcScreenreaderOnly
+        :screenreader-text="{
+          key: 'dashboard.validator.el_deposits.validity.screenreader_text',
+          interpolation: { validity: statusNameMap[status] },
+        }"
+      />
+      <BcIcon
+        :name="statusIconMap[statusDisplayTypeMap[status]]"
```
marcel-bitfly: **issue (blocking):** this can be simpler by putting display logic into `markup` instead of `<script>`

frontend/components/bc/BcStatusBadge.vue
```diff
@@ -0,0 +1,61 @@
...
+type Status =
+  'invalid'
+  | 'invalid_skipped'
+  | 'valid'
+type StatusColor = Extract<BadgeColor, 'green' | 'red'>
+type StatusDisplayType = 'error' | 'success'
+type StatusIcon = Extract<Icon, 'check' | 'xmark'>
+
+const { status } = defineProps<{
+  iconOnly?: boolean,
```
marcel-bitfly: **issue:** our convention is to prefix boolean with `is` or `has` 

**suggestion**: `hasText = true`

frontend/components/bc/BcStatusBadge.vue
marcel-bitfly, line 1: **issue:** rename generic component to `use case`

this component is not generalized but contains specific `knowlege` about `dashboard.validator.el_deposits`.
Thus should be prefixed with `DashbaordTableElDeposits` I guess.

frontend/composables/useApi.ts
marcel-bitfly, line 1: **thought:** so like mentioned before it might be best not to introduce `useApi` for `el deposits` now

frontend/i18n/locales/en.json
```diff
@@ -259,7 +260,15 @@
                 "title": "Epoch"
             },
             "el_deposits": {
-                "title": "Execution Layer"
+                "info_valid_with_invalid_signature": "Signature invalid, but will still be accepted due to a prior deposit with a valid signature.",
```
marcel-bitfly, line 264: **question:** is this grammatically correct?

`Signature is invalid, but will still be...`

frontend/assets/css/variables.scss
enzo-bitfly: The design system is already based on `1 rem`, it was us who were not, so we're actually closer to the design now than before.
I checked and couldn't find anything that looks off because of this change.

frontend/components/bc/BcBadge.vue
enzo-bitfly: Forgot to remove it, done.

frontend/components/bc/BcStatusBadge.vue
enzo-bitfly: Done. Also added `hasIcon`.
Only problem now is that you can in theory set `hasText` and `hasIcon` to false, in which case the component won't render any content.

frontend/components/bc/BcStatusBadge.vue
enzo-bitfly: Done

frontend/i18n/locales/en.json
enzo-bitfly: Yes. It could have more clarity using `The signature is invalid` but it's correct.

frontend/composables/useCustomFetch.ts
marcel-bitfly: **question**: @enzo-bitfly ok when I remove yr commit changing this file?

enzo-bitfly: Go ahead

PR #1442, opened 2025-03-27
(BEDS-1395) pectra deposits tab business logic

remoterami:
Based on #1414 and #1423


backend/pkg/api/data_access/vdb_deposits.go
```diff
@@ -32,16 +34,20 @@ func (d *DataAccessService) GetValidatorDashboardElDeposits(ctx context.Context,
...
-	// Resolve validator indices to pubkeys
-	byteaArray, err := d.getValidatorPubkeys(dashboardId)
-	if err != nil {
-		return nil, nil, err
+	// filters
+	searchPubkey := t.ReValidatorPublicKeyWithPrefix.MatchString(search)
+	searchGroup := len(search) > 0 && !dashboardId.AggregateGroups && t.ReName.MatchString(search)
+	searchIndexOrBlock := t.ReInteger.MatchString(search)
+
+	if search != "" && !searchPubkey && !searchGroup && !searchIndexOrBlock {
```
LuccaBitfly: **thought:** Maybe we can add a helper func for this, e.g.
```go
func isInvalidSearch(search string, checks ...bool) bool {
	if search == "" {
		return false
	}
	for _, check := range checks {
		if check {
			return false
		}
	}
	return true
}
```
and then
```go
if isInvalidSearch(search,
	searchPubkey,
	searchGroup,
	searchIndexOrBlock
) {
	// early return...
```

backend/pkg/api/data_access/vdb_deposits.go
```diff
@@ -32,16 +34,20 @@ func (d *DataAccessService) GetValidatorDashboardElDeposits(ctx context.Context,
...
 	}
 
-	// Resolve validator indices to pubkeys
-	byteaArray, err := d.getValidatorPubkeys(dashboardId)
-	if err != nil {
-		return nil, nil, err
+	// filters
+	searchPubkey := t.ReValidatorPublicKeyWithPrefix.MatchString(search)
+	searchGroup := len(search) > 0 && !dashboardId.AggregateGroups && t.ReName.MatchString(search)
+	searchIndexOrBlock := t.ReInteger.MatchString(search)
```
LuccaBitfly: **nitpick (non-blocking):** should prefix bool vars with `is` or `has`. E.g. `isValidSearchPubkey` or something.

backend/pkg/api/data_access/vdb_deposits.go
```diff
@@ -50,27 +56,69 @@ func (d *DataAccessService) GetValidatorDashboardElDeposits(ctx context.Context,
...
+						goqu.Or(
+							goqu.I("ed2.block_number").Lt(goqu.I("ed.block_number")),
+							goqu.And(
+								goqu.I("ed2.block_number").Eq(goqu.I("ed.block_number")),
+								goqu.I("ed2.log_index").Lt(goqu.I("ed.log_index")),
+							),
+						),
+					)), "invalid_skipped").
+				Else("invalid").
+				As("valid_signature"),
```
LuccaBitfly: **question:** The subquery only gets executed when `ed.valid_signature` is `false` right? I.e. this "case" expression is more efficient than leaving `Valid` as a bool and fetching "`has_previous_valid_signature`" for every validator.

backend/pkg/api/data_access/vdb_deposits.go
```diff
@@ -84,13 +132,65 @@ func (d *DataAccessService) GetValidatorDashboardElDeposits(ctx context.Context,
...
+		if searchIndexOrBlock || searchPubkey {
+			var index uint64
+			if searchIndexOrBlock {
+				idx, err := strconv.Atoi(search)
+				if err != nil {
+					return nil, nil, err
+				}
+				index = uint64(idx)
+			} else {
+				validatorMapping, err := d.services.GetCurrentValidatorMapping()
```
LuccaBitfly: **issue:** The validator mapping is fetched both here and in https://github.com/gobitfly/beaconchain/blob/ed3b4553bef6395a53c2e4bac27102c4e1c57a4e/backend/pkg/api/data_access/vdb_deposits.go#L216

Should rather fetch the validatorMapping once at the beginning.

backend/pkg/api/data_access/vdb_deposits.go
```diff
@@ -312,19 +419,64 @@ func (d *DataAccessService) GetValidatorDashboardClDeposits(ctx context.Context,
...
 	depositsDs := depositsBridgeDs
 	if d.config.ClConfig.ElectraForkEpoch < utils.MaxForkEpoch {
 		depositsDs = depositsDs.
 			UnionAll(depositRequestsDs)
 		if currentCursor.IsValid() {
-			postElectra := uint64(currentCursor.Slot)/d.config.ClConfig.SlotsPerEpoch > d.config.ClConfig.ElectraForkEpoch
+			postElectra := uint64(currentCursor.SlotProcessed)/d.config.ClConfig.SlotsPerEpoch > d.config.ClConfig.ElectraForkEpoch
 			if postElectra && currentCursor.Reverse {
 				depositsDs = depositRequestsDs
 			} else if !postElectra && !currentCursor.Reverse {
```
LuccaBitfly: **suggestion:** Pls add in a comment what happens in this whole if-block and why.

backend/pkg/api/data_access/vdb_deposits.go
```diff
@@ -312,19 +419,64 @@ func (d *DataAccessService) GetValidatorDashboardClDeposits(ctx context.Context,
...
+		)
+	}
+
+	if len(searchesBridge) > 0 {
+		depositsBridgeDs = depositsBridgeDs.Where(goqu.Or(searchesBridge...))
+		depositRequestsDs = depositRequestsDs.Where(goqu.Or(searchesRequests...))
 	}
 
 	depositsDs := depositsBridgeDs
 	if d.config.ClConfig.ElectraForkEpoch < utils.MaxForkEpoch {
```
LuccaBitfly, line 349: **question:** What is "max fork epoch"? Should maybe be commented at the place where it's defined.

backend/pkg/api/data_access/vdb_deposits.go
```diff
@@ -233,32 +326,47 @@ func (d *DataAccessService) GetValidatorDashboardClDeposits(ctx context.Context,
...
+		WithdrawalCredential []byte          `db:"withdrawalcredentials"`
+		Amount               decimal.Decimal `db:"amount"`
+		Signature            []byte          `db:"signature"`
+		SlotQueued           int64           `db:"slot_queued"`   // BEDS-1399
+		Type                 string          `db:"type"`          // BEDS-1399
+		Status               string          `db:"status"`        // BEDS-1399
+		RejectReason         sql.NullString  `db:"reject_reason"` // BEDS-1399
 	}
 
 	depositsBridgeDs := goqu.Dialect("postgres").
```
LuccaBitfly: **nitpick:** It was not clear to me why we need two dataset vars until reading the end of the function. Should have a comment above explaining this, e.g. `// fetches pre-pectra rows` above declaring `depositsBridgeDs` and an analogous comment for `depositRequestsDs` (if it's accurate?). Or maybe simply rename the whole var into `prePectraDepositsDs`.

backend/pkg/api/data_access/vdb_deposits.go
LuccaBitfly, line 349: **nitpick:** It appears that you first build your complete datasets and then choose in this if-block which ones you use. Wouldn't it be better the other way around? E.g. first you declare bool vars `hasPrePectraRows` and `hasPostPectraRows` with the logic here, and afterwards you build the datasets through helper funcs if their respective bool var is true, and finally you combine the data sets, if needed.

LuccaBitfly (changes requested):

backend/pkg/api/data_access/vdb_deposits.go
remoterami: It was basically meant for fork-independent compatibility, ie. post-pectra tables are not assumed to be present / queried for pre-pectra networks

backend/pkg/api/data_access/vdb_deposits.go
remoterami: > if it's accurate?

Well.. the old deposit mechanism will be disabled post pectra, but not necessarily with the fork epoch. It takes until all bridge deposits have been processed.

backend/pkg/api/data_access/vdb_deposits.go
remoterami: Yes true, that would probably be cleaner, I'll change it. Figured building the queries isn't that expensive :smile:

backend/pkg/api/data_access/vdb_deposits.go
remoterami: That's a good question, thanks for pointing it out!
> The subquery only gets executed when ed.valid_signature is false right?

Yes, but for each row so in DB land this isn't more efficient if it applies to too many rows :sweat_smile: A CTE should be the better choice. At least if we want to optimize for the more expensive cases, it may boil down to a tradeoff.. needs testing

The current impl is mostly concerned about logic+data correctness and schema compatibility, I didn't do too much performance optimization yet. A quick test on hoodi confirms that a CTE is faster, will change to that

LuccaBitfly (approved):
lgtm

remoterami:
resolved via #1480

PR #1446, opened 2025-03-28
feat(executionlayer): metrics for bigtable + client GetBlock

Tangui-Bitfly:
* metric for bigtable, contains the table and the action done
* metric for GetBlock from the client

**Tested on**: mainnet and gnosis with live indexing

Example result on mainnet and with poor performance laptop
Screenshot 2025-03-28 at 12 41 40

### DIFF
--- a/backend/pkg/commons/metrics/executionlayer.go
+++ b/backend/pkg/commons/metrics/executionlayer.go
@@ -35,4 +35,9 @@ var (
 		Name: "indexing_reorg_total",
 		Help: "Total number of block impacted by chain reorganizations",
 	}, []string{"chainID"})
+
+	ClientGetBlock = promauto.NewHistogramVec(prometheus.HistogramOpts{
+		Name: "client_get_block_time_seconds",
+		Help: "Time taken to retrieve a block from the blockchain and parse it",
+	}, []string{"chainID"})
 )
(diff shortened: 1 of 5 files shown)


backend/pkg/commons/metrics/executionlayer.go
```diff
@@ -35,4 +35,9 @@ var (
 		Name: "indexing_reorg_total",
 		Help: "Total number of block impacted by chain reorganizations",
 	}, []string{"chainID"})
+
+	ClientGetBlock = promauto.NewHistogramVec(prometheus.HistogramOpts{
```
Monika-Bitfly, line 39: instead of creating a metric specifically for GetBlock, it would be better to have a more general metric like ClientCallDuration with a method label - this way we can reuse the metric for other client calls in the future

backend/pkg/commons/metrics/executionlayer.go
```diff
@@ -35,4 +35,9 @@ var (
...
+		Name: "client_get_block_time_seconds",
+		Help: "Time taken to retrieve a block from the blockchain and parse it",
+	}, []string{"chainID"})
```
Monika-Bitfly: should be snake case "chain_id"

backend/pkg/commons/metrics/executionlayer.go
Tangui-Bitfly: The "client" i'm referring to is not the ethclient but the erigonclient (our wrapper around ethclient).
What you suggest is more applicable for the ethclient

invis-bitfly (approved):
lgtm™

PR #1452, opened 2025-03-31
feat(execution layer): handle genesis block

Tangui-Bitfly:
### Bigtable issue
Genesis block (number = 0) cause a problem with how we encode it in bigtable
* block 0 has key chainID:1000000000
* block 1 has key chainID:999999999
* block 2 has key chainID:999999998

Since Bigtable keys are lexicographically ordered, blocks are ordered like the following:
* chainID:1000000000 (block 0)
* ...
* chainID:(1000000000-n) (block n)
* ...
* chainID:999999998 (block 2)
* chainID:999999999 (block 1)

This means we cannot request the block range [0,1] because this requires that block 0 and 1 are adjacent in the table.
We must handle genesis block separately and then add it the rest of the range results.

### Last processed block issue
We also need to differentiate between no last block processed (fresh start) and last block processed equal to 0 (block 0 was indexed and we aim to index from block 1).

### DIFF
--- a/backend/pkg/commons/db2/store.go
+++ b/backend/pkg/commons/db2/store.go
@@ -313,27 +313,29 @@ func (store StoreV1) GetBlocksRange(chainID string, start, end uint64) ([]*types
 		return nil, fmt.Errorf("invalid block range provided (high: %v, low: %v)", end, start)
 	}
 
-	var row *database.Row
-	var err error
+	var rows []database.Row
 	// we need to index genesis block (number=0) alone because there is an issue with how we generate block key
 	// block 0 has key chainID:1000000000
 	// block 1 has key chainID:999999999
 	// both are not adjacent in database, thus preventing to retrieve the range [0,1]
 	if start == 0 {
-		row, err = store.blocks.GetRow(blockKey(chainID, start))
+		row, err := store.blocks.GetRow(blockKey(chainID, start))
 		if err != nil {
 			return nil, err
 		}
+		if row != nil {
+			rows = append([]database.Row{*row}, rows...)
+		}
 		start++
 	}
-	rows, err := store.blocks.GetRowsRange(blockKey(chainID, start), blockKey(chainID, end))
-	if err != nil {
-		return nil, err
+	if start <= end {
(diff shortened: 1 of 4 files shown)


backend/pkg/commons/db2/store.go
```diff
@@ -412,7 +428,7 @@ func (store StoreV1) GetLastBlockInDataTable(chainID string) (uint64, error) {
 		return 0, err
 	}
 	if len(rows) != 1 {
```
Monika-Bitfly, line 432: I think it's better to check if the `len(rows) == 0` ?

backend/pkg/commons/db2/store.go
```diff
@@ -313,11 +313,27 @@ func (store StoreV1) GetBlocksRange(chainID string, start, end uint64) ([]*types
...
+	// block 1 has key chainID:999999999
+	// both are not adjacent in database, thus preventing to retrieve the range [0,1]
+	if start == 0 {
+		row, err = store.blocks.GetRow(blockKey(chainID, start))
+		if err != nil {
+			return nil, err
+		}
+		start++
+	}
 	rows, err := store.blocks.GetRowsRange(blockKey(chainID, start), blockKey(chainID, end))
```
Monika-Bitfly: suggestion for edge case with `start = 0` and `end = 0`: check here again if the `start` value is now bigger than `end` value 
```
if start > end {
   return blocks, nil
}
```

backend/pkg/commons/db2/store.go
```diff
@@ -313,11 +313,27 @@ func (store StoreV1) GetBlocksRange(chainID string, start, end uint64) ([]*types
...
 	if err != nil {
 		return nil, err
 	}
 	var blocks []*types.Eth1Block
+	if row != nil {
```
Monika-Bitfly: I think it would better to first unmarshal the genesis block and append it to the `blocks` and then proceed with the remaining blocks parsing

backend/pkg/commons/db2/store.go
Tangui-Bitfly: good catch

invis-bitfly (approved):
lgtm™

PR #1459, opened 2025-04-01
refactor: introduce `statusReport` interface for `newStatusReport` in /monitoring/services

Monika-Bitfly:
Changes:
- Replace direct `services.NewStatusReport` usage with injectable `statusReport` interface
- Add `stubStatusReporter` for testing and `statusReporter` for production
- Make `newStatusReport` function internal
- Eliminate race condition in tests by removing global config access

Testing:
- Update unit tests to use mock status reporter
- Verify no more race detector warnings

### DIFF
--- a/backend/pkg/exporter/modules/base.go
+++ b/backend/pkg/exporter/modules/base.go
@@ -44,6 +44,7 @@ func StartAll(moduleCtx ModuleContext, modules []ModuleInterface, justV2 bool) {
 	if !justV2 {
 		ctx := context.Background()
 		consDB := db2.NewConsensusRepository(db.ReaderDb, db.WriterDb)
+		InitStatusReport()
 
 		networkLivenessUpdater := newNetworkLivenessUpdater(ctx, moduleCtx.ConsClient, consDB)
 		go networkLivenessUpdater.Export()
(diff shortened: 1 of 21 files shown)


backend/pkg/exporter/modules/base.go
```diff
@@ -44,6 +44,7 @@ func StartAll(moduleCtx ModuleContext, modules []ModuleInterface, justV2 bool) {
 	if !justV2 {
 		ctx := context.Background()
 		consDB := db2.NewConsensusRepository(db.ReaderDb, db.WriterDb)
+		services.InitStatusReport()
```
invis-bitfly: this needs to also be triggered when `justV2` is true. maybe move it to a higher level?

invis-bitfly (changes requested):

backend/pkg/exporter/modules/base.go
Monika-Bitfly: true! I've updated it now

backend/pkg/monitoring/services/base.go
```diff
@@ -133,3 +133,27 @@ func GetRequiredEvents() []constants.Event {
...
+
+type statusReport interface {
+	NewStatusReport(id constants.Event, timeout time.Duration, checkInterval time.Duration) func(status constants.StatusType, metadata map[string]string)
+}
+
+type stubStatusReporter struct{}
+
+func (sr stubStatusReporter) NewStatusReport(id constants.Event, timeout time.Duration, checkInterval time.Duration) func(status constants.StatusType, metadata map[string]string) {
+	return func(status constants.StatusType, metadata map[string]string) {
+		// No-op implementation
```
invis-bitfly: maybe a bit overkill but can you make it log a warning message if the `utils.Config` is set and the deployment type isnt `development`? so we notice when we accidentally use the stub in staging/prod

invis-bitfly (changes requested):

invis-bitfly (approved):
lgtm

PR #1467, opened 2025-04-03
feat(DashboardValidatorManagmentModal): add CTA to update `subscripti…

marcel-bitfly:
…on plan` when MEB is exceeded

When adding validators to a dashboard, e.g. via `withdrawal address` the `maximum effective balance` (MEB) could exceed the current `dashboard limit`.

We show the <BcPremiumModal> to inform the user about the possibility to upgrade the current plan.

See: BEDS-1402

### DIFF
--- a/frontend/components/dashboard/DashboardValidatorManagementModal.vue
+++ b/frontend/components/dashboard/DashboardValidatorManagementModal.vue
@@ -112,10 +112,10 @@ const changeGroup = async (body: PostValidatorDashboardValidatorsRequest, groupI
       method: 'POST',
     },
     { dashboardKey: dashboardKey.value },
-  )
-
-  loadData()
-  refreshOverview(dashboardKey.value)
+  ).then(() => {
+    loadData()
+    refreshOverview(dashboardKey.value)
+  })
 }
 
 const removeValidators = async (validators?: NumberOrString[]) => {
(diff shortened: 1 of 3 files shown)


frontend/components/dashboard/DashboardValidatorManagementModal.vue
```diff
@@ -337,8 +337,17 @@ const handleSubmit = (item: InternalPostSearchResponse['data'][number] | undefin
...
   selectedGroup.value,
-  )
-  resetInput()
+  ).then(() => {
+    resetInput()
+  }).catch((error) => {
+    if (error.statusCode === 403) {
+      dialog.open(BcPremiumModal, {
+        data: {
+          description: $t('dashboard.validator.management.validators_limit_exceeded'),
```
enzo-bitfly, line 346: question: the dialog shows different messages depending on the user having exceeded the EB before trying to add a validator or not. Would it make sense to pass this description to the dialog in `handleInvalidSubmit` in this file too?

enzo-bitfly (approved):
Nice job 👍

frontend/components/dashboard/DashboardValidatorManagementModal.vue
marcel-bitfly: See the video in the ticket.

Actually it is already if I did not get your question wrong: https://github.com/gobitfly/beaconchain/blob/BEDS-1402/show-error-when-validator-cannot-be-added-to-db/frontend/components/dashboard/DashboardValidatorManagementModal.vue#L293

PR #1469, opened 2025-04-03
fix(execution layer): make v2 iso with v1

Tangui-Bitfly:
Fix multiple little issues
* TransferBatch log topic for erc1155
* internal tx parser properly increase current trace index
* fix execution layer reporter log

**Tested on**: mainnet with live indexing for 7k blocks (1 day)


Monika-Bitfly (approved):

invis-bitfly (approved):
lgtm™

PR #1471, opened 2025-04-03
feat(execution layer): add Pectra new transformers

Tangui-Bitfly:


backend/pkg/executionlayer/transformer.go
```diff
@@ -642,6 +658,60 @@ func transformEnsNameRegistered(chainID string, block *types.Eth1Block, res *db2
...
 }
 
+func transformConsolidationRequests(chainID string, block *types.Eth1Block, res *db2.IndexedBlock) error {
+	var requests []db2.ConsolidationRequest
+	for txIndex, tx := range block.GetTransactions() {
+		for _, log := range tx.GetLogs() {
+			if !bytes.Equal(log.Address, consolidationContractAddress.Bytes()) {
+				continue
+			}
+			// we have found a consolidation event
```
Monika-Bitfly, line 671: we should validate the length of `log.Data` to ensure it is at least 116 bytes long for consolidation requests and 76 bytes long for withdrawal requests to avoid any potential issues

backend/pkg/executionlayer/transformer.go
```diff
@@ -642,6 +658,60 @@ func transformEnsNameRegistered(chainID string, block *types.Eth1Block, res *db2
...
+				TargetPubKey:   log.Data[68:116],
+				TxHash:         tx.GetHash(),
+				TxIndex:        txIndex,
+				BlockNumber:    block.GetNumber(),
+				BlockTimestamp: block.GetTime().AsTime(),
+			})
+		}
+	}
+	res.ConsolidationRequests = requests
+	return nil
```
Monika-Bitfly, line 688: atm this function never returns an error but if we add a validation check we can return an error from there

backend/pkg/executionlayer/transformer.go
```diff
@@ -642,6 +658,60 @@ func transformEnsNameRegistered(chainID string, block *types.Eth1Block, res *db2
...
+}
+
+func transformWithdrawalRequests(chainID string, block *types.Eth1Block, res *db2.IndexedBlock) error {
+	var requests []db2.WithdrawalRequest
+	for txIndex, tx := range block.GetTransactions() {
+		for _, log := range tx.GetLogs() {
+			if !bytes.Equal(log.Address, withdrawalContractAddress.Bytes()) {
+				continue
+			}
+			// we have found a consolidation event
```
Monika-Bitfly:
```suggestion
			// we have found a withdrawal event
```

backend/pkg/executionlayer/transformer.go
```diff
@@ -642,6 +658,66 @@ func transformEnsNameRegistered(chainID string, block *types.Eth1Block, res *db2
 	return nil
 }
 
+func transformConsolidationRequests(chainID string, block *types.Eth1Block, res *db2.IndexedBlock) error {
+	var requests []db2.ConsolidationRequest
```
Monika-Bitfly, line 662: missing metrics for task duration with `bt_transform_consolidation_requests` and  `bt_transform_withdrawal_requests` labels

backend/pkg/commons/db2/bridge.go
```diff
@@ -0,0 +1,44 @@
...
+
+func NewBridgeStore(db *sqlx.DB) *BridgeSQLStore {
+	return &BridgeSQLStore{
+		db: db,
+	}
+}
+
+func (store BridgeSQLStore) SetConsolidationRequests(request ConsolidationRequest) error {
+	_, err := store.db.Exec(`
+		INSERT INTO eth1_consolidation_requests (tx_hash, tx_index, block_number, block_ts, source_address, source_pubkey, target_pubkey) VALUES ($1, $2, $3, $4, $5, $6, $7)
```
Monika-Bitfly, line 24: can you format those two queries pls?

backend/pkg/executionlayer/transformer.go
Tangui-Bitfly: metrics are injected at higher level
https://github.com/gobitfly/beaconchain/blob/63a9435663b10f257d59a201d27ed1ddf9f672e6/backend/pkg/executionlayer/blockindexer.go#L179

backend/pkg/commons/db2/bridge.go
Tangui-Bitfly: That's a linter job. Since we don't have a defined format, both are acceptable.

Monika-Bitfly: True, both work - but messy SQL bites us later. A few simple line breaks and indents make it way easier to scan and debug, IMO worth the small effort now to save headaches later

Monika-Bitfly (approved):

backend/pkg/executionlayer/transformer.go
```diff
@@ -23,6 +24,11 @@ import (
 	"github.com/gobitfly/beaconchain/pkg/commons/types"
 )
 
+var (
+	consolidationContractAddress = common.HexToAddress("0x00431F263cE400f4455c2dCf564e53007Ca4bbBb")
+	withdrawalContractAddress    = common.HexToAddress("0x0c15F14308530b7CDB8460094BbB9cC28b9AaaAA")
```
invis-bitfly, line 29: would prefer having it configurable over some kind of config. 
the deposit contract address for example isnt the same on gnosis/sepolia as it is on mainnet, and while the request contracts should match mainnet on sepolia it wouldnt surprise me if gnosis ends up modifying (and therefore changing the address of) the request contract somehow.

Tangui-Bitfly: Addressed in https://github.com/gobitfly/beaconchain/pull/1490.
I prefer to have my constant hardcoded in the go code rather than in some configuration files.
Ideally we should be able to run the indexer with only the clients endpoints (for the chain, obviously we also have db, metrics, ...). All the rest can be retrieved by the chain id.

PR #1487, opened 2025-04-08
fix(dashboard_exporter): invalid deposits breaking everything

invis-bitfly:


backend/pkg/consapi/types/debug.go
```diff
@@ -5,8 +5,9 @@ import (
 )
 
 type ElectraDeposit struct {
-	Pubkey hexutil.Bytes `json:"pubkey" db:"pubkey"`
-	Amount uint64        `json:"amount,string" db:"amount"`
+	Pubkey         hexutil.Bytes `json:"pubkey" db:"pubkey"`
+	Amount         int64         `json:"amount,string" db:"amount"`
+	SignatureValid bool          `json:"signature_valid" db:"signature_valid"`
```
remoterami: **issue**: Seems like this new field is never written to. Looks like this should happen in `ElectraGetProcessedDeposits`

backend/pkg/exporter/db/db.go
```diff
@@ -1597,21 +1597,24 @@ func ElectraGetEpochProcessedHashes(epoch uint64) ([][]byte, error) {
...
-		Amount uint64 `db:"amount"`
+		Amount int64  `db:"amount"`
 		Pubkey string `db:"pubkey"`
 	}
 	q := goqu.Dialect("postgres").Select(
 		goqu.L("data->>'amount'").As("amount"),
 		goqu.L("data->>'pubkey'").As("pubkey"),
+		// SignatureValid field, assume true if not present
+		goqu.COALESCE(goqu.L("data->>'signature_valid'"), goqu.L("true")).As("signature_valid"),
 	).
```
remoterami: **issue**: the struct definition doesn't include the new selected `signature_valid` field. Maybe this definition could be removed entirely, and just use `[]constypes.ElectraDeposit` instead?

backend/pkg/exporter/modules/dashboard_data_process.go
```diff
@@ -482,16 +495,60 @@ func (d *dashboardData) processElectraDeposits(data *MultiEpochData, tar *[]type
...
-				d.log.Tracef("processing electra deposit of pubkey %s (len %d) in epoch %d", deposit.Pubkey, len(deposit.Pubkey), epoch)
-				index, indexExists := data.validatorBasedData.validatorIndices[string(deposit.Pubkey)]
+				if _, ok := deposits[string(deposit.Pubkey)]; !ok {
+					deposits[string(deposit.Pubkey)] = make([]constypes.ElectraDeposit, 0)
+				}
+				deposits[string(deposit.Pubkey)] = append(deposits[string(deposit.Pubkey)], deposit)
+			}
+			// now we have a map of pubkey => deposits
+			// loop over the map and process the deposits
+			epochEnd := int64(epoch) - 1
```
remoterami: **question**: Shouldn't the next epoch be `int64(epoch) + 1`? If not, why?

remoterami (changes requested):
left some comments, ptal

**praise**: The detailed explainer comments for `processElectraDeposits` really help with understanding what's happening and why :+1:

backend/pkg/exporter/modules/dashboard_data_process.go
invis-bitfly: `epochEnd` should be `epoch` while `epochStart` should be `epoch - 1`, no idea how that mixup happened. the `validatorStates` is keyed so `n-1` gives you the data at the start of the epoch and `n` gives you the data at the end.
i wrote down examples but not sure if it makes things clearer or even more confusing :sweat_smile: 

image

backend/pkg/consapi/types/debug.go
invis-bitfly: fixed

backend/pkg/exporter/db/db.go
invis-bitfly: fixed. was originally using a different struct because of the base64 stuff, but found a cleaner way to resolve that

remoterami (approved):
utAck

PR #1490, opened 2025-04-08
fix(transformer): updates dependencies and adds EL data for pectra

Tangui-Bitfly:
* update go-ethereum dependency to import correct system contract address
* change db schema: new unique id + new fields
* update pectra transformers to retrieve the new fields
* open the door for configurable system contract


backend/pkg/executionlayer/transformer.go
```diff
@@ -804,7 +852,13 @@ func isBlobTx(txType uint32) bool {
 	return txType == gethtypes.BlobTxType
 }
 
+// isValidItx filter unwanted internal transactions
 func isValidItx(itx *types.Eth1InternalTransaction) bool {
+	// always process internals going to system contracts
+	if bytes.Equal(itx.To, params.ConsolidationQueueAddress.Bytes()) ||
+		bytes.Equal(itx.To, params.WithdrawalQueueAddress.Bytes()) {
+		return true
+	}
```
invis-bitfly: with this change to `isValidItx`, do we have to reindex all internal transactions for post-pectra blocks? i.e also re-run `transformITx` (and `transformBlock`?), and not just `transformConsolidationRequests` & `transformWithdrawalRequests`?

Tangui-Bitfly: Hum... good point maybe I went to quick with this condition. I will update.

Tangui-Bitfly: https://github.com/gobitfly/beaconchain/pull/1490/commits/7fb3a43b3e9c592539d2ea1a002453a662f911af make things better and only requires to run `transformConsolidationRequests` and `transformWithdrawalRequests`

backend/pkg/commons/db/migrations/postgres/20250407121321_add_el_requests_columns.sql
```diff
@@ -0,0 +1,28 @@
+-- +goose Up
+-- +goose StatementBegin
+ALTER TABLE eth1_consolidation_requests ADD COLUMN IF NOT EXISTS itx_index int4 not null;
+ALTER TABLE eth1_consolidation_requests ADD COLUMN IF NOT EXISTS from_address bytea NOT NULL;
+ALTER TABLE eth1_consolidation_requests ADD COLUMN IF NOT EXISTS fee bytea NOT NULL;
```
remoterami: **issue**: Please make `fee` an appropriate numeric type like `bigint` or `decimal`

remoterami: **suggestion**: Please also add an unique `id` column that can be used to reference table rows.

remoterami (changes requested):
Left some comments about the schema changes, ptal

remoterami (approved):
utAck

PR #1499, opened 2025-04-09
(BEDS-1453) cl events schema mocks

remoterami:
also relates to #1490


backend/pkg/commons/db/migrations/postgres/20250409114238_extend_pectra_tables.sql
```diff
@@ -0,0 +1,93 @@
...
+
+ALTER TABLE eth1_consolidation_requests ADD COLUMN IF NOT EXISTS id SERIAL UNIQUE;
+ALTER TABLE eth1_withdrawal_requests ADD COLUMN IF NOT EXISTS id SERIAL UNIQUE;
+
+CREATE TABLE IF NOT EXISTS blocks_deposit_requests (
+	id SERIAL PRIMARY KEY UNIQUE,                               -- would ideally match eth1_id if possible ?
+	eth1_id bytea REFERENCES eth1_deposits(merkletree_index),   -- if available (= genesis not hardcoded)
+	-- null if never queued bcus rejected
+	slot_queued int4,
+	index_queued int4,                                          -- max MAX_DEPOSIT_REQUESTS_PER_PAYLOAD (8,192)
```
invis-bitfly: re: comment, fork epoch can create >10k queued deposits

backend/pkg/commons/db/migrations/postgres/20250409114238_extend_pectra_tables.sql
```diff
@@ -0,0 +1,93 @@
...
+	block_queued_root bytea,                                    -- can't fk blocks.blockroot bcus field isn't unique,
+	-- null if not processed yet bcus queued
+	slot_processed int4,                                        -- really an epoch property, but it points to the slot containing the transition
+	index_processed int2,                                       -- max MAX_PENDING_DEPOSITS_PER_EPOCH (16)
+	block_processed_root bytea,
+	
+	type text not null,                                         -- account, system_excess
```
invis-bitfly: re: comment, potentially `genesis` as well

backend/pkg/commons/db/migrations/postgres/20250409114238_extend_pectra_tables.sql
```diff
@@ -0,0 +1,93 @@
+-- +goose Up
+-- +goose StatementBegin
+
+ALTER TABLE eth1_consolidation_requests ADD COLUMN IF NOT EXISTS id SERIAL UNIQUE;
+ALTER TABLE eth1_withdrawal_requests ADD COLUMN IF NOT EXISTS id SERIAL UNIQUE;
+
+CREATE TABLE IF NOT EXISTS blocks_deposit_requests (
+	id SERIAL PRIMARY KEY UNIQUE,                               -- would ideally match eth1_id if possible ?
+	eth1_id bytea REFERENCES eth1_deposits(merkletree_index),   -- if available (= genesis not hardcoded)
```
invis-bitfly: re: comment, `type=auto` would also require this to be null, fyi

backend/pkg/commons/db/migrations/postgres/20250409114238_extend_pectra_tables.sql
```diff
@@ -0,0 +1,93 @@
...
+	
+	status text not null,                                                       -- queued, completed, rejected
+	reject_reason text,                                                         -- ~15
+
+	validator_pubkey bytea NOT NULL,
+	amount int8 NOT NULL
+);
+
+
+CREATE TABLE IF NOT EXISTS blocks_switch_to_compounding_requests (
```
invis-bitfly, line 75: if we add this table then we also need a table for exit requests

backend/pkg/commons/db/migrations/postgres/20250409114238_extend_pectra_tables.sql
```diff
@@ -0,0 +1,93 @@
...
+	pubkey bytea NOT NULL,
+	withdrawal_credentials bytea NOT NULL,
+	amount int8 NOT NULL,
+	signature bytea NOT NULL
+);
+
+
+
+CREATE TABLE IF NOT EXISTS blocks_consolidation_requests (
+	id integer REFERENCES eth1_consolidation_requests(id) NOT NULL PRIMARY KEY, -- no genesis hardcoding here, should always have an eth1 event
```
invis-bitfly: can we match the deposit setup? it would be nice if i can have it be null during initial transform and then have a second step that resolves it to the el events, instead of having to do everything at once

backend/pkg/commons/db/migrations/postgres/20250409114238_extend_pectra_tables.sql
```diff
@@ -0,0 +1,93 @@
...
+	
+	amount_consolidated int8 NULL,
+	source_index int4 NOT NULL,
+	target_index int4 NOT NULL
+);
+
+
+
+CREATE TABLE IF NOT EXISTS blocks_withdrawal_requests (
+	id integer REFERENCES eth1_withdrawal_requests(id) NOT NULL PRIMARY KEY,    -- no genesis hardcoding here, should always have an eth1 event
```
invis-bitfly: ditto

backend/pkg/commons/db/migrations/postgres/20250409114238_extend_pectra_tables.sql
```diff
@@ -0,0 +1,93 @@
...
+	status text not null,                                                       -- queued, completed, rejected
+	reject_reason text,                                                         -- ~15
+
+	validator_pubkey bytea NOT NULL,
+	amount int8 NOT NULL
+);
+
+
+CREATE TABLE IF NOT EXISTS blocks_switch_to_compounding_requests (
+    id integer REFERENCES eth1_consolidation_requests(id) NOT NULL PRIMARY KEY, -- no genesis hardcoding here, should always have an eth1 event
```
invis-bitfly: ditto EX

backend/pkg/commons/db/migrations/postgres/20250409114238_extend_pectra_tables.sql
remoterami: Well sure, the epoch can have 32 * 8,192 in total then. But you're right, each slot can have > 8k with excess events from consolidations, so the comment isn't entirely correct

backend/pkg/commons/db/migrations/postgres/20250409114238_extend_pectra_tables.sql
remoterami: I don't insist. It did already exist and is used in v1 already, so I just included the mod, too.

I'd prefer having less tables to manage while also minimizing the v1 changes even if that's an inconsistency, but fine with either adding exit requests or removing both

backend/pkg/commons/db/migrations/postgres/20250409114238_extend_pectra_tables.sql
InvisibleSymbol: Sorry, meant the fork upgrade state transition specifically. That's in a single slot and can create >10k events

backend/pkg/commons/db/migrations/postgres/20250409114238_extend_pectra_tables.sql
remoterami: Ok sure will change, so the assumption isn't true anymore `should always have an eth1 event`

backend/pkg/commons/db/migrations/postgres/20250409114238_extend_pectra_tables.sql
remoterami: Ah right thanks, non-active validators getting special treatment.. as long as `int4` holds we're fine :smile: 
comments are just for additional context anyways

backend/pkg/commons/db/migrations/postgres/20250409114238_extend_pectra_tables.sql
invis-bitfly: thinking of it. switch to compounding requests never hit the queue, as they get processed pre-queue. do we still want to retain the split queued / processed slot columns?

remoterami: No shouldn't be needed, missed that, can remove the columns

PR #1503, opened 2025-04-10
feat: add `CL consolidations table`

enzo-bitfly:
See: BEDS-1351

### DIFF
--- a/frontend/components/dashboard/table/DashboardTableClConsolidations.vue
+++ b/frontend/components/dashboard/table/DashboardTableClConsolidations.vue
@@ -1,0 +1,317 @@
+<script setup lang="ts">
+import type { DataTableSortEvent } from 'primevue/datatable'
+import type {
+  GetValidatorDashboardConsensusLayerConsolidationsResponse,
+  VDBConsolidationsClTableRow,
+} from '~/types/api/validator_dashboard'
+import type {
+  Cursor, TableQueryParams,
+} from '~/types/datatable'
+import BcTableControl from '~/components/bc/table/BcTableControl.vue'
+
+const {
+  clConsolidations,
+} = defineProps<{
+  clConsolidations?: GetValidatorDashboardConsensusLayerConsolidationsResponse,
+  isLoading: boolean,
+}>()
+
+const { width } = useWindowSize()
+const isMobile = computed(() => {
+  return width.value < 768
+})
+
+const query = defineModel<TableQueryParams>('query')
+
+const onSort = (sort: DataTableSortEvent) => {
+  query.value = setQuerySort(sort, query.value)
+}
+const setCursor = (cursor: Cursor) => {
+  query.value = setQueryCursor(cursor, query.value)
+}
+const setPageSize = (limit: number) => {
+  query.value = setQueryPageSize(limit, query.value)
+}
+const setSearch = (value?: string) => {
+  query.value = {
+    ...query.value,
+    search: value,
+  }
+}
+</script>
+
+<template>
+  <BcTableControl
+    :title="$t('dashboard.validator.cl_consolidations.title')"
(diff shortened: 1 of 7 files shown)


frontend/components/dashboard/table/DashboardTableClConsolidations.vue
```diff
@@ -0,0 +1,316 @@
...
+  query.value = setQueryCursor(cursor, query.value)
+}
+const setPageSize = (limit: number) => {
+  query.value = setQueryPageSize(limit, query.value)
+}
+const setSearch = (value?: string) => {
+  query.value = {
+    ...query.value, search: value,
+  }
+}
```
marcel-bitfly: **polish (non-blocking):** this is several times in yr codebase
```suggestion
const setSearch = (value?: string) => {
  query.value = {
    ...query.value,
    search: value,
  }
}
```

frontend/i18n/locales/en.json
```diff
@@ -225,6 +225,30 @@
                 "search_placeholder_public": "Index, Public Key",
                 "title": "Blocks"
             },
+            "cl_consolidations": {
+                "info_slot_processed_estimated": "This is an estimation since this consolidation is still queued.",
+                "reject_reason": {
+                    "full_queue": "The consolidation queue is full.",
+                    "insufficient_consolidation_churn": "Not enough churn capacity available to consolidate.",
```
marcel-bitfly, line 232: **question:** @Buttaa should we link to an explanation what `churn capacity` means?

frontend/i18n/locales/en.json
```diff
@@ -274,8 +298,8 @@
                 "block_processed": "Block Processed",
                 "block_queued": "Block Queued",
                 "info_block_processed_estimated": "This is an estimation since this consolidation is still queued.",
-                "search_placeholder_guest_dashboard": "Index, Slot, Address",
-                "search_placeholder_private_dashboard": "Index, Slot, Address, Group",
+                "search_placeholder_guest_dashboard": "Index, Block, Address",
+                "search_placeholder_private_dashboard": "Index, Block, Address, Group",
```
marcel-bitfly, line 302: **thought:** can this key be reused (since it occurs 3 times)?

frontend/pages/dashboard/[[id]]/index.vue
```diff
@@ -282,6 +282,28 @@ const elConsolidations = computed(() => {
   return elConsolidationsData.value || undefined
 })
 
+// Consolidations consensus layer deposits data
+const clConsolidationsQueryParams = ref<TableQueryParams>({
+  limit: 5, sort: 'timestamp:desc',
```
Comment (marcel-bitfly) [inline frontend/pages/dashboard/[[id]]/index.vue:288]: **polish (non-blocking):** 
```suggestion
  limit: 5, 
  sort: 'timestamp:desc',
```

marcel-bitfly:
**issue**: `dashboard id` of `guest dashboards` are stil removed when changing the `tab`

frontend/pages/dashboard/[[id]]/index.vue
Comment (enzo-bitfly) [reply frontend/pages/dashboard/[[id]]/index.vue:288]: Done

frontend/i18n/locales/en.json
enzo-bitfly: It's a coincidence that the same string happens for different tables, but if we want to avoid repetition, maybe something like this could make sense:
```js
:placeholder="${`$t('xxx.index'), $t('xxx.block'), $t('xxx.group')`}"
```
But I'm ok with having a `placeholder` translation for each table too. I can change this in the next table if you find that better.

frontend/components/dashboard/table/DashboardTableClConsolidations.vue
enzo-bitfly: Yeah, we only started catching this in the previous code review. Fixed for this table

PR #1505, opened 2025-04-10
dashboard_exporter: use aligned first epoch of sync period

invis-bitfly:


backend/pkg/exporter/db/db.go
```diff
@@ -1088,18 +1088,31 @@ func BackfillRoi(epochs []BackfillMetadata) error {
...
+		GROUP BY epoch_timestamp
+		Settings use_skip_indexes_if_final = 1
+	`, FinalEpochsTableName), timestamps)
+	if err != nil {
+		return fmt.Errorf("error fetching epoch count: %w", err)
+	}
+	if len(result) != len(epochs) {
+		return fmt.Errorf("epoch count mismatch: expected %d, got %d", len(epochs), len(result))
+	}
+	for i, e := range epochs {
```
Monika-Bitfly: is iterating over `epochs` correct here or should we be ranging over `result` instead?

InvisibleSymbol: whoops, ye should iterate over `result` not `epochs`, nice catch

Monika-Bitfly (approved):

PR #1513, opened 2025-04-16
(BEDS-1462) maxEB Banner for GuestDB

remoterami:


frontend/types/api/validator_dashboard.ts
```diff
@@ -26,6 +26,7 @@ export interface VDBOverviewData {
   apr: PeriodicValues<ClElValue<number /* float64 */>>;
   chart_history_seconds: ChartHistorySeconds;
   balances: ValidatorBalances;
+  is_above_eb_limit: boolean; // refers to owner; relevant for shared dashboards
```
marcel-bitfly: **suggestion (if-minor):** without context it is really hard to guess what `eb` stands for
```suggestion
  hasExceededEffectiveBalance: boolean;
```

marcel-bitfly: **nitpick:** imo the context why this change is necessary should be explained in the body of your `commit msg`

frontend/pages/dashboard/[[id]]/index.vue
```diff
@@ -220,7 +205,7 @@ watch(
     <BcPageWrapper>
       <template #banner>
         <BcNotificationBanner
-          v-if="hasReachedLimit"
+          v-if=" overview?.is_above_effective_balance_limit"
```
Comment (enzo-bitfly) [inline frontend/pages/dashboard/[[id]]/index.vue:208]: had we started with this...  :')

question: would it make sense to use this flag now also in the `DashboardValidatorManagementModal.vue` too?

frontend/components/dashboard/DashboardValidatorManagementModal.vue
```diff
@@ -313,21 +314,49 @@ const handleSubmit = (item: InternalPostSearchResponse['data'][number] | undefin
...
+    let hasError = false
+    const currentValidators = decodeBase64Url(dashboardKey.value).split(',')
     if (item.type === 'validator') {
-      addEntities([ `${item.value.index}` ])
-      resetInput()
+      const newValidators = [
+        ...currentValidators,
+        item.value.index,
+      ]
+      await loadData(encodeBase64Url(newValidators.join(',')))
```
marcel-bitfly, line 324: **thought:** this all adding to the mess which is our `dashboard`

But refactoring PR is on its way 🫣

See: #1525

frontend/pages/dashboard/[[id]]/index.vue
Comment (marcel-bitfly) [reply frontend/pages/dashboard/[[id]]/index.vue:208]: no as it is part of the overview call

enzo-bitfly (approved):
It is cleaner than before though 🤷

PR #1518, opened 2025-04-23
refactor: remove `queue position` from `DashboardValidatorManagementModal`

marcel-bitfly:
This feature showed information like `Pending #2` before,
which indicated the number in the `activation queue`.

See: BEDS-1468


LuccaBitfly (approved):
lgtm

PR #1519, opened 2025-04-23
feat: add `CL withdrawals table` for Pectra

enzo-bitfly:
-Remove obsolete `DashboardTableWithdrawals.vue`

See: BEDS-1352


marcel-bitfly (approved):
nice 💪

frontend/components/dashboard/table/DashboardTableClWithdrawals.vue
```diff
@@ -0,0 +1,467 @@
...
+                  <div class="dashboard-table-cl-withdrawals__details-label">
+                    {{ $t("table.status") }}
+                  </div>
+                  <div class="dashboard-table-cl-withdrawals__details-value">
+                    <DashboardTableClWithdrawalsStatus
+                      :status="slotProps.data.status"
+                      :reject-reason="slotProps.data.reject_reason"
+                    />
+                  </div>
+                </div>
```
marcel-bitfly, line 305: **thought (non-blocking):** the thing that comes to my mind if I see the `design` is, if this should rather be a `table` for better `a11y`
Could be hard complicated to work with because of the second `column` though.

enzo-bitfly: I want to refactor this into a reusable component in the next days, so that's a good point. I had also thought of using `dl`, `dt` and `dd`

marcel-bitfly: I am not so happy with the design to be honest. As the layout looks too busy with dynamic column width.
But let's see

PR #1522, opened 2025-04-24
fix: account for consolidations when writing validator_stats

LuccaBitfly:
- Fixes consolidations being counted towards CL Rewards
- Treats incoming and outgoing consolidations as deposits and withdrawals respectively

### DIFF
--- a/backend/pkg/commons/db/statistics.go
+++ b/backend/pkg/commons/db/statistics.go
@@ -134,6 +134,12 @@ func WriteValidatorStatisticsForDay(day uint64, client rpc.Client) error {
 		}
 		return nil
 	})
+	g.Go(func() error {
+		if err := gatherValidatorConsolidations(day, validatorData, validatorDataMux); err != nil {
+			return fmt.Errorf("error in gatherValidatorConsolidations: %w", err)
+		}
+		return nil
+	})
 
 	var statisticsData1d []*types.ValidatorStatsTableDbRow
 	g.Go(func() error {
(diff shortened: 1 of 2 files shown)


backend/pkg/commons/db/statistics.go
```diff
@@ -212,12 +218,12 @@ func WriteValidatorStatisticsForDay(day uint64, client rpc.Client) error {
...
-		data.WithdrawalsTotal = previousDayData.WithdrawalsTotal + data.Withdrawals
-		data.WithdrawalsAmountTotal = previousDayData.WithdrawalsAmountTotal + data.WithdrawalsAmount
+		data.WithdrawalsTotal = previousDayData.WithdrawalsTotal + data.Withdrawals + data.OutgoingConsolidationCount
+		data.WithdrawalsAmountTotal = previousDayData.WithdrawalsAmountTotal + data.WithdrawalsAmount + data.OutgoingConsolidationAmount
 
 		// update deposits total
-		data.DepositsTotal = previousDayData.DepositsTotal + data.Deposits
-		data.DepositsAmountTotal = previousDayData.DepositsAmountTotal + data.DepositsAmount
+		data.DepositsTotal = previousDayData.DepositsTotal + data.Deposits + data.IncomingConsolidationCount
+		data.DepositsAmountTotal = previousDayData.DepositsAmountTotal + data.DepositsAmount + data.IncomingConsolidationAmount
```
remoterami: **question**: Why should the withdrawal/deposit counts be adjusted by consolidations? I see why that could be done, but I don't think we should as it's not the same thing. For example I wouldn't consider an incoming consolidation a deposit as those are usually associated with new ETH entering the beacon chain through the deposit contract. Mixing those things up might seem simple for the user at first glance but we shouldn't compromise on correctness for that. Or is this an intended v1 workaround decision in order to avoid extending all the table schemas? 

The withdrawals count is also used here for example, so that number would be skewed imo. The deposits count is sourced from somewhere else afaics though (and not adjusted by the consolidation count), so it's also inconsistent
image

remoterami (changes requested):

backend/pkg/commons/db/statistics.go
invis-bitfly: sure, lets not include it in the count for now

LuccaBitfly: fixed, ptal

remoterami (approved):
utAck

PR #1527, opened 2025-04-28
fix: correct db type casting

remoterami:


invis-bitfly (approved):
`Int256` is prob overkill but also cant really hurt, this query doesnt run that often. lgtm

PR #1532, opened 2025-04-29
Pectra api fixes

remoterami:


backend/pkg/api/data_access/vdb_withdrawals.go
```diff
@@ -915,15 +928,21 @@ func (d *DataAccessService) getNextWithdrawalRow(queryValidators []validatorGrou
...
 			Hash:       t.Hash(address.String()),
 			Ens:        ens_name,
 			IsContract: contractStatus[0] == types.CONTRACT_CREATION || contractStatus[0] == types.CONTRACT_PRESENT,
 		},
-		Amount: utils.GWeiToWei(big.NewInt(int64(withdrawalAmount))),
+		Amount:    utils.GWeiToWei(big.NewInt(int64(withdrawalAmount))),
+		Type:      "auto",
+		Status:    "queued",
+		Slot:      nextWithdrawalSlot,
+		SlotIndex: 0, // ?
```
mccreedy-bitfly, line 945: ?

manuelsc (approved):
utACK

backend/pkg/api/data_access/vdb_withdrawals.go
remoterami: The slot estimate isn't very precise already, no point in further estimation of the index atm. In a rare corner case this may lead to a non-unique row id, but negligible

PR #1549, opened 2025-05-02
fix(api): pectra

remoterami:


backend/pkg/api/data_access/vdb_withdrawals.go
```diff
@@ -238,7 +238,7 @@ func (d *DataAccessService) GetValidatorDashboardElWithdrawals(ctx context.Conte
 			TxIndexQueued:   res.TxIndex,
 			ITxIndexQueued:  res.ITxIndex,
 			TxHash:          t.Hash(hexutil.Encode(res.TxHash)),
-			Amount:          res.Amount,
+			Amount:          res.Amount.Mul(decimal.NewFromInt(1e9)),
```
sasha-bitfly: Can we use util function for better readability? Like `GWeiToWei`?

backend/pkg/api/data_access/vdb_withdrawals.go
```diff
@@ -565,16 +566,11 @@ func (d *DataAccessService) GetValidatorDashboardClWithdrawals(ctx context.Conte
...
-			nextData = &t.VDBWithdrawalsClTableRow{
-				IsMissingEstimate: true,
-			}
-		}
+		// if nextData != nil {
+		// Complete the next data TODO
+		// TODO integrate label/ens data for "next" row
+		// nextData.Recipient.Ens = addressEns[string(nextData.Recipient.Hash)]
+		// }
 		responseData = append([]t.VDBWithdrawalsClTableRow{*nextData}, responseData...)
```
sasha-bitfly: Are we sure nextData is not `nil`?

backend/pkg/api/data_access/vdb_withdrawals.go
remoterami: It's not that clean because of the required data types, but changed it

backend/pkg/api/data_access/vdb_consolidations.go
```diff
@@ -204,10 +208,13 @@ func (d *DataAccessService) GetValidatorDashboardExecutionLayerConsolidations(ct
...
 
+		row.TimestampQueued = res.BlockQueuedTime.Unix()
+
 		if res.BlockProcessedTime.Valid {
 			row.Status = "processed"
-			row.BlockProcessed = uint64(res.BlockProcessed.Int64)
-			row.TimestampProcessed = res.BlockProcessedTime.Int64
+			blockProcessed := uint64(res.BlockProcessed.Int64)
+			row.BlockProcessed = &blockProcessed
+			row.TimestampProcessed = &res.BlockProcessed.Int64
```
sasha-bitfly: Currently, `row.TimestampProcessed = &res.BlockProcessed.Int64` should it be `row.TimestampProcessed = &res.BlockProcessedTime.Int64` @remoterami

remoterami: Oh yes it should, great catch!

PR #1557, opened 2025-05-05
chore: improve log message when no relay payloads are found

Monika-Bitfly:
Changes:
- change log level from Error to Trace when no relay payloads are found


invis-bitfly (approved):
lgtm

PR #1561, opened 2025-05-06
Beds 1498/use filter on total amounts

enzo-bitfly:


frontend/components/dashboard/table/DashboardTableClConsolidations.vue
```diff
@@ -75,8 +75,8 @@ const setSearch = (value?: string) => {
             </template>
             <template #body="slotProps">
               <BcTableDateTime
-                v-if="slotProps.data.slot_processed !== undefined"
-                :unix-timestamp="getTimestampFromSlot(slotProps.data.slot_processed)"
+                v-if="slotProps.data.slot !== undefined"
```
marcel-bitfly, line 78: **suggestion:** 
```suggestion
                v-if="slotProps.data.slot"
```

frontend/components/dashboard/table/DashboardTableClWithdrawals.vue
```diff
@@ -95,25 +95,25 @@ const getGroupName = (groupId: number) => {
           :selected-sort="query?.sort"
           :cursor="query?.cursor"
           :page-size="query?.limit"
-          :row-class="(row: VDBWithdrawalsElTableRow) => row.status === 'queued' ? 'grayed-out-row' : ''"
-          :is-row-expandable="(row: VDBWithdrawalsElTableRow) => row.index !== undefined"
+          :row-class="(row: VDBWithdrawalsClTableRow) => row.status === 'queued' ? 'grayed-out-row' : ''"
+          :is-row-expandable="(row: VDBWithdrawalsClTableRow) => row.index !== undefined"
```
marcel-bitfly, line 99: **suggestion:** 
```suggestion
          :is-row-expandable="(row: VDBWithdrawalsClTableRow) => !!row.index"
```

frontend/components/dashboard/table/DashboardTableClConsolidations.vue
enzo-bitfly: I want to specifically see if the value is not there, since `0`, although not possible considering that it's a slot in the past, would be falsy. I find `!== undefined` more expressive as `slot` is an optional value.

frontend/components/dashboard/table/DashboardTableClWithdrawals.vue
enzo-bitfly: Same here, `index` is a property that could potentially be `0` and that would make the row not expandable.

PR #1562, opened 2025-05-06
(BEDS-1489) apply filter to withdrawals sums

remoterami:


backend/pkg/api/data_access/vdb_management.go
```diff
@@ -905,7 +905,7 @@ func (d *DataAccessService) UpdateValidatorDashboardPublicId(ctx context.Context
 	}{}
 
 	// Update the name and settings of the public validator dashboard
-	err := d.alloyWriter.GetContext(ctx, &dbReturn, `
+	err := d.writerDb.GetContext(ctx, &dbReturn, `
```
LuccaBitfly, line 908: **issue:** Using `writerDb` for a get. Is this intentional?

backend/pkg/api/data_access/vdb_withdrawals.go
```diff
@@ -335,6 +225,123 @@ func (d *DataAccessService) GetValidatorDashboardElWithdrawals(ctx context.Conte
...
+func getElWithdrawalsDs(dashboardId t.VDBId, search string) (*goqu.SelectDataset, error) {
+	isValidSearchPubkey := t.ReValidatorPublicKeyWithPrefix.MatchString(search)
+	isValidSearchSenderOrWithdrawer := t.ReEthereumAddress.MatchString(search)
+	isValidSearchIndexOrBlock := t.ReInteger.MatchString(search)
+	isValidSearchGroup := len(search) > 0 && !dashboardId.AggregateGroups && t.ReName.MatchString(search)
+	isValidSearchTxHash := t.ReTransactionHash.MatchString(search)
+
+	if isInvalidSearch(search, isValidSearchPubkey, isValidSearchSenderOrWithdrawer, isValidSearchIndexOrBlock, isValidSearchGroup, isValidSearchTxHash) {
+		return nil, nil
+	}
```
LuccaBitfly, line 237: **issue:** 
! We prob don't have time for this comment to be implemented, I'd do it in another PR. Feel free to resolve!

This is business logic and should not be part of the query-building helper func. Ideally, this code block would be part of the handler. (Goes for all query-building funcs)

backend/pkg/api/data_access/vdb_withdrawals.go
```diff
@@ -335,6 +225,123 @@ func (d *DataAccessService) GetValidatorDashboardElWithdrawals(ctx context.Conte
 	return responseData, p, nil
 }
 
+func getElWithdrawalsDs(dashboardId t.VDBId, search string) (*goqu.SelectDataset, error) {
```
LuccaBitfly, line 228: **praise:**  it's good that we're extracting this func :pray:

backend/pkg/api/data_access/vdb_withdrawals.go
LuccaBitfly: I can now see why you did it this way (i.e. to early exit in the total funcs aswell + reuse the query w/o select there). This seems a bit hacky to me. I'd probably doing in some way like this:

```go
type ElWithdrawalSearches struct {
	isValidSearchWithdrawalAddress, isValidSearchIndexOrSlot, isValidSearchGroup, isValidSearchPublicKey bool
}

func getValidElWithdrawalSearches(search string) ElWithdrawalSearches {
	return ElWithdrawalSearches{false, false, false, false} // TODO
}

func buildElWithdrawalQueryRelations(ardId t.VDBId, search string, validSearches ElWithdrawalSearches) (*goqu.SelectDataset, error) {
	// TODO this is where you put all of the query building without the select
	return nil, nil
}
func getTotalElWithdrawalsDs(dashboardId t.VDBId, search string, validSearches ElWithdrawalSearches) (*goqu.SelectDataset, error) {
	ds, err := buildElWithdrawalQueryRelations(dashboardId, search, validSearches)
	if err != nil {
		return nil, err
	}
	ds = ds.
		Select( /* TODO: select stuff relevant for Total EL Withdrawals */ )
	return nil, nil
}

// TODO: create similar DS func for normal el withdrawals

func (d *DataAccessService) GetValidatorDashboardTtotalElWithdrawals(ctx context.Context, dashboardId t.VDBId, search string, protocolModes t.VDBProtocolModes) (*t.VDBTotalExecutionWithdrawalsData, error) {
	result := &t.VDBTotalExecutionWithdrawalsData{
		TotalAmount: decimal.NewFromBigInt(big.NewInt(0), 0),
	}

	validSearches := getValidElWithdrawalSearches(search)
	if isInvalidSearch(search, validSearches.isValidSearchWithdrawalAddress, validSearches.isValidSearchIndexOrSlot, validSearches.isValidSearchGroup, validSearches.isValidSearchPublicKey) {
		return result, nil
	}
	withdrawalsDs, err := getTotalElWithdrawalsDs(dashboardId, search, validSearches)
	if err != nil {
		return nil, err
	}

	sum, err := runQuery[int64](ctx, d.readerDb, withdrawalsDs)
	if err != nil && !errors.Is(err, sql.ErrNoRows) {
		return nil, fmt.Errorf("error getting total cl withdrawals: %w", err)
	}

	result.TotalAmount = utils.GWeiToWei(big.NewInt(sum))
	return result, nil
}

// TODO: create func for normal el withdrawals
```

backend/pkg/api/data_access/vdb_management.go
LuccaBitfly, line 1: **issue:** Getting error
```
{
    "error": "error getting total cl withdrawals: error executing query: ERROR: column request.slot does not exist (SQLSTATE 42703)"
}
```
with `GET /api/i/validator-dashboards/50/total-consensus-layer-withdrawals`

LuccaBitfly (changes requested):

backend/pkg/api/data_access/vdb_management.go
remoterami: The query contains an `UPDATE` so it's definitely writing something, db is correct imo. Since it's also returning a row it's probably correct to be a call to `GetContext`

LuccaBitfly: right, didn't read in which func this was executed in, mb

LuccaBitfly (approved):
lgtm

PR #1566, opened 2025-05-07
feat: reset `cursor` on `page size change`

marcel-bitfly:
Due to a bigger refactoring for `backend` it was
decided, that we should reset the cursor on `page size` change events.
Otherwise the following scenario is misleading:
1. user sets `page size`: 5
2. user gets to page 3
3. user sets `page size` to 10
4. user clicks `previous`
result:  user will get back to last `result` with `page size` 5

See: BEDS-1483


enzo-bitfly (approved):
Nice, thank you 💪

PR #1568, opened 2025-05-07
refactor: move ens and contract fetching into helper func

LuccaBitfly:


backend/pkg/api/data_access/vdb_blocks.go
```diff
@@ -326,12 +324,7 @@ func (d *DataAccessService) GetValidatorDashboardBlocks(ctx context.Context, das
 		Reward decimal.Decimal
 	}
 	startTime := time.Now()
-	query, args, err := blocksDs.Prepared(true).ToSQL()
-	if err != nil {
-		return nil, nil, err
-	}
-
-	err = d.alloyReader.SelectContext(ctx, &proposals, query, args...)
+	proposals, err := runQueryRows[[]dbResult](ctx, d.alloyReader, blocksDs)
```
remoterami:
```suggestion
	proposals, err := runQueryRows[[]dbResult](ctx, d.readerDb, blocksDs)
```

backend/pkg/api/data_access/vdb_helpers.go
```diff
@@ -702,3 +705,60 @@ func getViewAndDateColumn(aggregation enums.ChartAggregation) (string, string, e
...
+	if dbBlock.Valid {
+		block = dbBlock.Int64
+	}
+	return block
+}
+
+// getElInfo is a generic function to get EL info for addresses and their contract status.
+// Returns a map of address + block to address info.
+// Use the func `getElInfoKey` to get the key for the map.
+func getElInfo[In any](ctx context.Context, d *DataAccessService, input []In, buildReqs func(In) []db.ContractInteractionAtRequest) (map[string] /* address + ':' + block /*/ t.Address, error) {
```
remoterami: **issue**: the block number alone is not enough to uniquely identify a request in all cases. This is essentially because a contract can be created and destroyed at an address multiple times per block (or transaction even)

Basically the full request is instead, ie. address + block number + tx index + itx index. Seems like it's not possible to have the request struct itself being the map key with some custom comparator in golang, but mabye you can come up with something close. If not just concat all 4 properties I guess

backend/pkg/api/data_access/vdb_consolidations.go
```diff
@@ -172,23 +171,19 @@ func (d *DataAccessService) GetValidatorDashboardExecutionLayerConsolidations(ct
...
-			Block:    -1,
-			TxIdx:    -1,
-			TraceIdx: -1,
+	buildReqs := func(row elDbResult) []db.ContractInteractionAtRequest {
+		return []db.ContractInteractionAtRequest{
+			{
+				Address:  fmt.Sprintf("%x", row.Consolidator),
+				Block:    getInteractionRequestBlock(row.BlockProcessed),
+				TxIdx:    -1,
+				TraceIdx: -1,
```
remoterami: **suggestion**: maybe this should just use the queued state instead (like the age column does in FE now), which are always present. I.e. `row.BlockQueued`, `row.TxIndex` and `row.ITxIndex`

backend/pkg/api/data_access/vdb_deposits.go
```diff
@@ -203,10 +202,34 @@ func (d *DataAccessService) GetValidatorDashboardElDeposits(ctx context.Context,
...
+		if len(row.Depositor) > 0 {
+			reqs = append(reqs, db.ContractInteractionAtRequest{
+				Address:  fmt.Sprintf("%x", row.Depositor),
+				Block:    row.BlockNumber,
+				TxIdx:    -1,
+				TraceIdx: -1,
+			})
+		}
+		return reqs
+	}
```
remoterami: **praise**: Nice to see how the individual funcs can make custom use of this

backend/pkg/api/data_access/vdb_deposits.go
```diff
@@ -203,10 +202,34 @@ func (d *DataAccessService) GetValidatorDashboardElDeposits(ctx context.Context,
...
+	elInfos, err := getElInfo(ctx, d, data, buildReqs)
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to get el info: %w", err)
+	}
 	for i, row := range data {
+		if len(row.Depositor) == 0 {
+			responseData[i].Depositor = responseData[i].From
+		}
```
remoterami: **issue**: Didn't run this, but looks like it should be moved below the main `responseData[i] = t.VDBExecutionDepositsTableRow{` assignment. Otherwise it's just always empty, no?

backend/pkg/api/data_access/vdb_withdrawals.go
```diff
@@ -102,23 +102,19 @@ func (d *DataAccessService) GetValidatorDashboardElWithdrawals(ctx context.Conte
...
-		addrStr := hexutil.Encode(addr)
-		addressMapping[addrStr] = nil
-		return t.Address{Hash: t.Hash(addrStr)}
+	buildReqs := func(row dbResult) []db.ContractInteractionAtRequest {
+		return []db.ContractInteractionAtRequest{
+			db.ContractInteractionAtRequest{
+				Address:  fmt.Sprintf("%x", row.Withdrawer),
+				Block:    getInteractionRequestBlock(row.BlockProcessed),
+				TxIdx:    -1,
+				TraceIdx: -1,
```
remoterami: Same here as above, could just switch from processed to queued state here using block + tx + itx (btw I designed the db schema in part to be able to do this). As it's the EL table I think it's more correct to refer to the EL event itself anyways

remoterami (changes requested):

backend/pkg/api/data_access/vdb_deposits.go
LuccaBitfly: You're right that it will get overwritten, good catch. Should be `row.Depositor = row.From` instead, which needs to happen before creating the response data because we need a map lookup for `row.Depositor`

LuccaBitfly:
Also added goroutines when fetching address and contract info, ptal

LuccaBitfly:
Added EL info fetch for CL-withdrawal recipient also.

remoterami (approved):
lgtm

LuccaBitfly:
rebase

PR #1569, opened 2025-05-07
fix(api): show skimming estimate

remoterami:


backend/pkg/api/data_access/vdb_withdrawals.go
```diff
@@ -538,17 +544,17 @@ func (d *DataAccessService) GetValidatorDashboardClWithdrawals(ctx context.Conte
...
-	// Remove the last entry from data as it is only required for the check
-	if moreDataFlag {
-		responseData = responseData[:len(responseData)-1]
-		queryResult = queryResult[:len(queryResult)-1]
+		// Remove the last entry from data as it is only required for the check
+		if moreDataWithEstimateFlag {
+			responseData = responseData[:len(responseData)-1]
+			queryResult = queryResult[:len(queryResult)-1]
+		}
 	}
```
LuccaBitfly: **question:** Instead of potentially removing the last element twice (once before the "next row" block and once at the end of it), would it work if we only had something like this after that block?
```go
/*
<code block adding the next withdrawal row>
*/
// checking response instead of `moreDataFlag` because response may have been modified in prev block
if len(responseData) > int(limit) { 
	responseData = responseData[:limit]
	queryResult = queryResult[:limit]
}
```

remoterami: good question, should be possible. I added this and also adjusted the `moreDataFlag` in a similar fashion, ptal

backend/pkg/api/data_access/vdb_withdrawals.go
```diff
@@ -535,20 +534,18 @@ func (d *DataAccessService) GetValidatorDashboardClWithdrawals(ctx context.Conte
...
-	if moreDataFlag {
-		responseData = responseData[:len(responseData)-1]
-		queryResult = queryResult[:len(queryResult)-1]
+	moreDataFlag := len(responseData) > int(limit)
+	if !moreDataFlag && !currentCursor.IsValid() {
+		// No paging required
+		return responseData, &t.Paging{}, nil
+	} else if moreDataFlag {
+		responseData = responseData[:limit]
+		queryResult = queryResult[:limit]
```
LuccaBitfly: **thought (non-blocking):** Possible with fewer conditions, don't know if that's better tho.
```suggestion
	if moreDataFlag {
		responseData = responseData[:limit]
		queryResult = queryResult[:limit]
	} else if !currentCursor.IsValid() {
		// No paging required
		return responseData, &t.Paging{}, nil
```

LuccaBitfly (approved):
lgtm, one thing:

backend/pkg/api/data_access/vdb_withdrawals.go
LuccaBitfly: nvm, still would need to check `currentCursor.IsValid()`

PR #1571, opened 2025-05-08
Reject reasons changes

enzo-bitfly:


frontend/i18n/locales/en.json
```diff
@@ -263,15 +263,15 @@
...
+                    "excess_balance": "Validator had no excess balance available for withdrawal.",
                     "exiting": "The validator was exiting when this request was processed.",
                     "full_queue": "The withdrawal queue was full when this request was processed.",
                     "inactive": "The validator was inactive when this request was processed.",
-                    "insufficient_effective_balance": "Effective balance is below the withdrawal threshold.",
-                    "no_execution_withdrawal_credentials": "Validator has unsupported withdrawal credentials.\nUpdate withdrawal credentials to 0x01- or 0x02-prefixed.",
-                    "not_compounding": "Validator is not compounding.\nUpdate withdrawal credentials to 0x02-prefixed to enable compounding.",
-                    "pending_withdrawals": "Wait for current withdrawals to process before starting another.",
-                    "too_young": "Validator is too new to withdraw.",
+                    "insufficient_effective_balance": "Effective balance below the withdrawal threshold.",
```
marcel-bitfly, line 273: **question:** are you sure this is better english now?

frontend/components/dashboard/table/DashboardTableClDepositsStatus.vue
```diff
@@ -57,6 +58,23 @@ defineProps<{
         {{ $t("dashboard.validator.table.status_text.rejected") }}
       </template>
     </BcBadge>
+    <BcTooltip
+      v-if="status === 'rejected' && !isMobile && !!rejectReason"
```
marcel-bitfly: **suggestion:** 

```vue
      v-if="rejectReason && !isMobile && status === 'rejected'"

```

frontend/components/dashboard/table/DashboardTableClDepositsStatus.vue
```diff
@@ -57,6 +58,23 @@ defineProps<{
...
+      class="status-tooltip-trigger"
+    >
+      <BcIcon
+        name="circle-info"
+      />
+      <template #tooltip>
+        <slot name="tooltip">
+          <span v-if="rejectReason === 'invalid_signature'">
+            {{ $t('dashboard.validator.cl_consolidations.reject_reason.invalid_signature') }}
+          </span>
```
marcel-bitfly, line 74: **issue:** it could easily happen that the `tooltip` has no message when you wire the code together like that

 may be check for `invalid_signature` in line 62?

frontend/i18n/locales/en.json
enzo-bitfly: @philip-bitfly

frontend/components/dashboard/table/DashboardTableClDepositsStatus.vue
enzo-bitfly: why? 🤔 
also, I prefer boolean-casting `rejectReason` for expressiveness.

frontend/components/dashboard/table/DashboardTableClDepositsStatus.vue
enzo-bitfly: true, but the same applies to any other tooltip that gets a new `reject_reason` added. Would we check in `cl_withdrawals` that it's any of the 11 handled reasons?

Would a default reject reason text for the tooltips make sense? @philip-bitfly

frontend/components/dashboard/table/DashboardTableClDepositsStatus.vue
marcel-bitfly: 1. you should have a reason for `casting`, otherwise you just make the code more complex (more mental capacity)

one situation I can think of would be return types, e.g.:
```vue
const hasRejectReason = computed(() => !!rejectReason)
```

2. it is easier to see that you need a `reject reason` otherwise this will not render
3. putting `rejectReason` upfront, the code might run more optimized (but this is a weak argument).

PR #1573, opened 2025-05-12
fix: handle type 4/EIP-7702 txs in statistics.WriteExecutionChartSeri…

guybrush:
…esForDay


backend/pkg/commons/db/statistics.go
```diff
@@ -1748,6 +1748,15 @@ func WriteExecutionChartSeriesForDay(day int64) error {
 				totalBurnedBlob = blobGasUsed.Mul(decimal.NewFromBigInt(new(big.Int).SetBytes(tx.BlobGasPrice), 0))
 				totalBlobCount = totalBlobCount.Add(decimal.NewFromInt(int64(len(tx.BlobVersionedHashes))))
 
+			case 4:
+				// EIP-7702
+				// priority fee is capped because the base fee is filled first
+				tipFee = decimal.Min(prioFee, maxFee.Sub(baseFee))
+				blobTxCount += 1
```
remoterami: **suggestion**: This value is not used currently, but it's still not correct semantically. Should rather introduce a new counter `contractAuthorizationTxCount` or so

remoterami (changes requested):

remoterami (approved):
lgtm

PR #1578, opened 2025-05-14
refactor: remove csrf middleware

LuccaBitfly:
- Since the API uses no HTML forms and validates the Content-Type of all unsafe requests, the middleware is unnecessary.
- Added linter rule preventing any form parsing being added in the future.
- Proper CORS config has to be maintained with or without CSRF middleware anyways.


remoterami (approved):
lgtm

PR #1584, opened 2025-05-15
(BEDS-1516) CL deposits estimate

remoterami:


backend/pkg/api/data_access/vdb_deposits.go
```diff
@@ -439,17 +462,112 @@ func (d *DataAccessService) GetValidatorDashboardClDeposits(ctx context.Context,
...
+			requestsDs = requestsDs.
+				Where(goqu.I("status").Eq("queued")).
+				Order(order...)
+			queuedData, err := runQueryRows[[]dbResult](ctx, d.readerDb, requestsDs)
+			if err != nil {
+				return nil, nil, err
+			}
+			for _, queuedRow := range queuedData {
+				for i, dataRow := range data { // find validator
+					if reflect.DeepEqual(dataRow.PublicKey, queuedRow.PublicKey) {
```
LuccaBitfly: **issue:** Please use `slices.Equal` for byte slice comparison

backend/pkg/api/data_access/vdb_deposits.go
```diff
@@ -439,17 +462,112 @@ func (d *DataAccessService) GetValidatorDashboardClDeposits(ctx context.Context,
...
+									break
+								}
+							}
+							if found {
+								continue
+							}
+						}
+						event.ResultRefs = append(event.ResultRefs, newRef)
+						break
+					}
```
LuccaBitfly: **issue:** Please use the guard clause pattern when the entire logic falls within an if statement instead, i.e.
```
if !Equals(dataRow.PublicKey, queuedRow.PublicKey) {
	continue
}
<your code>
```

backend/pkg/api/data_access/vdb_deposits.go
```diff
@@ -409,6 +414,24 @@ func (d *DataAccessService) GetValidatorDashboardClDeposits(ctx context.Context,
...
 
+	moreDataFlag := len(data) > int(limit)
+	hasFetchedAllQueued := false
+	type queueEventRef struct {
+		ResultRefs []*t.VDBConsensusDepositsTableRow
+		Counter    int
+	}
+	queuedEvents := make(map[uint64]*queueEventRef, 0)
+	queuedValidators := make([]uint64, 0)
+	if !moreDataFlag && !currentCursor.IsValid() { // could extend
```
LuccaBitfly: **question:** What's the rationale here? Only if we have no more data AND our cursor is invalid we set the flag to true? How can cursors become invalid?

backend/pkg/api/data_access/vdb_deposits.go
```diff
@@ -409,6 +414,24 @@ func (d *DataAccessService) GetValidatorDashboardClDeposits(ctx context.Context,
...
+		hasFetchedAllQueued = true
+	} else if colSort.Column == enums.VDBDepositsClColumns.Slot {
```
LuccaBitfly: **suggestion:** Could be simplified with:
```go
	hasFetchedAllQueued := true
	if colSort.Column == enums.VDBDepositsClColumns.Slot && (moreDataFlag || currentCursor.IsValid()) {
```

backend/pkg/api/data_access/vdb_deposits.go
```diff
@@ -409,6 +414,24 @@ func (d *DataAccessService) GetValidatorDashboardClDeposits(ctx context.Context,
 		return nil, nil, fmt.Errorf("failed to recover indices after query: %w", err)
 	}
 
+	moreDataFlag := len(data) > int(limit)
+	hasFetchedAllQueued := false
+	type queueEventRef struct {
+		ResultRefs []*t.VDBConsensusDepositsTableRow
+		Counter    int
+	}
+	queuedEvents := make(map[uint64]*queueEventRef, 0)
```
LuccaBitfly: **question:** Why does this need to be a map to a pointer? Why not just `map[uint64]queueEventRef`

backend/pkg/api/data_access/vdb_deposits.go
```diff
@@ -481,9 +485,53 @@ func (d *DataAccessService) GetValidatorDashboardClDeposits(ctx context.Context,
...
+			requestsDs = requestsDs.
+				Where(goqu.I("status").Eq("queued")).
+				Order(order...)
+			queuedData, err := runQueryRows[[]dbResult](ctx, d.readerDb, requestsDs)
+			if err != nil {
+				return nil, nil, err
+			}
+			for _, queuedRow := range queuedData {
+				for i, dataRow := range data { // find validator
+					if reflect.DeepEqual(dataRow.PublicKey, queuedRow.PublicKey) {
```
LuccaBitfly: **suggestion:** Instead iterating over `data` here and breaking at the end of the first equal pubkey, consider doing
```go
				i := slices.IndexFunc(data, func(row dbResult) bool {
					return slices.Equal(row.PublicKey, queuedRow.PublicKey)
				})
				dataRow := data[i]
```

backend/pkg/api/data_access/vdb_deposits.go
```diff
@@ -481,9 +485,53 @@ func (d *DataAccessService) GetValidatorDashboardClDeposits(ctx context.Context,
...
+									found = *event.ResultRefs[j].SlotQueued > uint64(queuedRow.SlotQueued.Int64)
+								}
+								if found {
+									break
+								}
+							}
+							if found {
+								continue
+							}
+						}
```
LuccaBitfly: **question:** What's happening here? A `found` var is being set, but I don't see what's being done with it.

backend/pkg/api/data_access/vdb_deposits.go
```diff
@@ -439,17 +461,112 @@ func (d *DataAccessService) GetValidatorDashboardClDeposits(ctx context.Context,
...
+			if hasFetchedAllQueued {
+				if _, ok := queuedEvents[responseData[i].Index]; !ok {
+					queuedEvents[responseData[i].Index] = &queueEventRef{
+						ResultRefs: make([]*t.VDBConsensusDepositsTableRow, 0),
+					}
+				}
+				refsArr := queuedEvents[responseData[i].Index]
+				// avoid array copies
+				(*refsArr).ResultRefs = append((*refsArr).ResultRefs, &responseData[i])
+			}
```
LuccaBitfly: **issue:** Looking at this from a general design pattern perspective, I don't like that we're preparing additional fetches while in the process of creating the response, especially when incorporating the extra data happens through pointer/counter magic later on. I'd much rather we iterate through the initial query result multiple times to get what we need beforehand and build the complete response at the end with all available data.

backend/pkg/api/data_access/vdb_deposits.go
LuccaBitfly, line 1: **issue:** After looking through the code for a bit I honestly don't understand what's happening 😅 May I ask you to rewrite it according the the [comment above](https://github.com/gobitfly/beaconchain/pull/1584#discussion_r2095380011)?  I can't wrap my head around all of the pointer stuff.

backend/pkg/api/data_access/vdb_deposits.go
LuccaBitfly: On that note, what is `newRef`? I have no idea what the entire `queuedData` loop is supposed to do.

backend/pkg/api/data_access/vdb_deposits.go
```diff
@@ -439,17 +461,112 @@ func (d *DataAccessService) GetValidatorDashboardClDeposits(ctx context.Context,
...
+			return nil, nil, err
+		}
+		for _, row := range queueData {
+			event := queuedEvents[row.ValidatorIndex]
+
+			slotProcessed := row.ClearEpoch * d.config.ClConfig.SlotsPerEpoch
+			if event.Counter < len(event.ResultRefs) && event.ResultRefs[event.Counter] != nil {
+				event.ResultRefs[event.Counter].SlotProcessed = &slotProcessed
+			}
+			event.Counter++
```
LuccaBitfly: **issue:** Please explain in a comment what this counter stuff is doing. Is there any way of doing all this without counters?

LuccaBitfly (changes requested):

backend/pkg/api/data_access/vdb_deposits.go
LuccaBitfly: Also please add the answers as comments.

backend/pkg/api/data_access/vdb_deposits.go
LuccaBitfly: Also please add the answer as a comment.

backend/pkg/api/data_access/vdb_deposits.go
remoterami: > How can cursors become invalid?

Simply if it's not present because the user didn't provide a cursor, then it's "invalid" because the object is not initialized

> What's the rationale here? Only if we have no more data AND our cursor is invalid we set the flag to true?

Correct, having just one condition alone is not enough. E.g. no more data but a valid cursor means there's "prev" data, so for amount sort or `desc == reverse` it could be possible to not have fetched all queued events for a validator.
As the comment indicates you could think of multiple similar combinations which would allow setting `hasFetchedAllQueued` to true. Would be an optimization but I'd assume they are rather rare

backend/pkg/api/data_access/vdb_deposits.go
remoterami: Not sure if I misunderstood, but this is missing the `false` cases, no? Unless we want to do this, which isn't really more readable than the current version imo:
```
	hasFetchedAllQueued := true
	if colSort.Column == enums.VDBDepositsClColumns.Slot && (moreDataFlag || currentCursor.IsValid()) {
		// ...
	} else {
		hasFetchedAllQueued = false
	}
```

And in that case I'd prefer my version, as that's easier to extend with more special cases

backend/pkg/api/data_access/vdb_deposits.go
remoterami: It was meant as a memory optimization, but thinking about it again I was wrong here (forgot slice copies are not really overhead as it only copies the slice header, not all elements; see comment in line 476). Changed

backend/pkg/api/data_access/vdb_deposits.go
remoterami: It's not always the first occurence, this was meant to just look at the newest results, see https://github.com/gobitfly/beaconchain/blob/c42285579da899615ad83715375b5eb21f09a133/backend/pkg/api/data_access/vdb_deposits.go#L526-L528 But this might actually not work for amount sort - I'll try to rework this, was hoping I could avoid having an per-validator-match-counter :sweat_smile:

backend/pkg/api/data_access/vdb_deposits.go
remoterami: So first of all, **all** of this is essentially handling a set of corner cases: It's just about validators which have more than one queued deposit at the same time. More specifically even, about those multiple queued deposits of which not all of them are returned by the current request (because of pagination / filtering). So this is just about ensuring correctness for the remaining, idk, 2%; and it could be argued that their relevance also depends on total network conditions, if the queue is almost 0 anyways it would just return wrong data for a short amount of time.

Anyways: The basic goal was to map the entries from the `blocks_deposit_requests_v2` table to the correct entries in `pending_deposits_queue`, so we can return the correct estimate. Heuristics like "join on amount / slot number" could may suffice to return another 80% of the remainder, but they don't ensure correctness in all cases.`queueEventRef` is the central component which implements that mapping (for each validator): each `ResultRefs` entry refers to one row in the total `responseData` list. Ie. the n-th entry in the `ResultRefs` slice refers to the n-th occurence of the respective validator in the result. So `Counter` is just used as an index here, to keep track of the slice position which has been used last tiem. The counter could be removed by removing the first element of the slice on iteration ig, but I doubt that'd make it all more understandable.

After all of this the simplest approach might be to just extend `pending_deposits_queue` by a foreign key to the other, then everything here in data access boils down to a single join. I can research how much effort that may be. Worth pointing out that this issue will become relevant to subsequent consolidation & withdrawal queue estimations, too.

remoterami:
Reworked, data is filled via gobitfly/eth2-beaconchain-explorer#3082. For context and reasoning see https://github.com/gobitfly/beaconchain/pull/1584#discussion_r2095780571

LuccaBitfly (approved):
utAck

PR #1588, opened 2025-05-20
fix: use v2-pectra-request-tables in statistics

guybrush:


backend/pkg/commons/db/statistics.go
```diff
@@ -841,34 +852,42 @@ func gatherValidatorDepositWithdrawals(day uint64, data []*types.ValidatorStatsT
...
+				blocks_deposits.publickey,
+				blocks_deposits.block_slot,
+				blocks_deposits.block_index
 			from
 				blocks_deposits
+			inner join blocks on
+				blocks_deposits.block_root = blocks.blockroot and (blocks.status = '1' or blocks.slot = 0)
 			where
-				valid_signature
-				and publickey in (select bd.publickey from blocks_deposits bd where bd.valid_signature and bd.block_slot >= $1 and bd.block_slot <= $2)
```
guybrush, line 852: note: valid_signature check here is not needed since this query is only to get performance-gains (like noted in the comment in the patch)

backend/pkg/commons/db/statistics.go
```diff
@@ -841,34 +852,42 @@ func gatherValidatorDepositWithdrawals(day uint64, data []*types.ValidatorStatsT
...
-				request_index
+				distinct on (blocks_deposit_requests_v2.pubkey)
+				blocks_deposit_requests_v2.pubkey,
+				blocks_deposit_requests_v2.slot_processed as block_slot,
+				blocks_deposit_requests_v2.index_processed as request_index
 			from
-				blocks_deposit_requests
+				blocks_deposit_requests_v2
+			inner join blocks on
+				blocks_deposit_requests_v2.block_processed_root = blocks.blockroot and (blocks.status = '1' or blocks.slot = 0)
```
guybrush, line 881: this check may be redundant since we check for `blocks_deposit_requests_v2.status = 'completed'` anyway, but it doesnt hurt too much perf-wise and is good to be more explicit i guess

LuccaBitfly (approved):
utAck

PR #1589, opened 2025-05-20
BEDS-1508: [2/2] add backfill job to fix missing electra fork events in ch

invis-bitfly:


backend/pkg/commons/db/migrations/clickhouse/20250520135931_finalize_electra_fork_epoch_backfill.sql
```diff
@@ -0,0 +1,17 @@
+-- +goose Up
+-- +goose StatementBegin
+-- +goose ENVSUB ON
+ALTER TABLE _final_validator_dashboard_data_epoch
+UPDATE 
+    withdrawal_count = withdrawal_count + dictGet(_dict_backfill_electra_fork_epoch_events, 'withdrawal_count', (epoch_timestamp, validator_index)),
+    withdrawal_amount = withdrawal_amount + dictGet(_dict_backfill_electra_fork_epoch_events, 'withdrawal_amount', (epoch_timestamp, validator_index)),
+    roi_divisor = roi_divisor - dictGet(_dict_backfill_electra_fork_epoch_events, 'withdrawal_amount', (epoch_timestamp, validator_index))
```
remoterami: **question**: why do we subtract from `roi_divisor` here? Shouldn't we also add to `roi_dividend` instead, like in the prev PR?

remoterami (changes requested):

backend/pkg/commons/db/migrations/clickhouse/20250520135931_finalize_electra_fork_epoch_backfill.sql
InvisibleSymbol: Good catch, forgot to fix it here as well

remoterami (dismissed):
utAck

PR #1592, opened 2025-05-21
(BEDS-1519) activation estimation

remoterami:


backend/pkg/api/data_access/vdb_summary.go
```diff
@@ -1083,21 +1074,17 @@ func (d *DataAccessService) GetValidatorDashboardSummaryValidators(ctx context.C
...
-				validatorInfo.Timestamp = uint64(utils.EpochToTime(uint64(metadata.ActivationEpoch.Int64)).Unix())
-			} else if metadata.Queues.ActivationIndex.Valid {
-				queuePosition := uint64(metadata.Queues.ActivationIndex.Int64)
-				epochsToWait := (queuePosition - 1) / activationChurnRate
-				// calculate dequeue epoch
-				estimatedActivationEpoch := latestEpoch + epochsToWait + 1
-				// add activation offset
-				estimatedActivationEpoch += utils.Config.Chain.ClConfig.MaxSeedLookahead + 1
-				validatorInfo.Timestamp = uint64(utils.EpochToTime(estimatedActivationEpoch).Unix())
+			activationEpoch, err := d.getValidatorActivation(ctx, validatorIndex)
```
LuccaBitfly: **issue:** You're performing one query per pending validator. Wouldn't it be better to fetch the data for all pending validators at once?

backend/pkg/api/data_access/validators.go
```diff
@@ -137,3 +139,92 @@ func (d *DataAccessService) GetValidatorsByGraffiti(ctx context.Context, graffit
 
 	return runQueryRows[[]t.VDBValidator](ctx, d.alloyReader, validatorsDs)
 }
+
+// estimate activation for pending and deposited validators
+// TODO support estimates for validators without index
+func (d *DataAccessService) getValidatorActivation(ctx context.Context, validator uint64) (uint64, error) {
```
LuccaBitfly: **nitpick:** Func name doesn't tell you that you get an epoch. Maybe `getValidatorActivationEpochEstimate` or something would be better? Pls come up with something suited :sweat_smile:

backend/pkg/api/data_access/validators.go
```diff
@@ -137,3 +139,92 @@ func (d *DataAccessService) GetValidatorsByGraffiti(ctx context.Context, graffit
...
+		return 0, fmt.Errorf("validator index %d not found in validator mapping", validator)
+	}
+	metadata := validatorMapping.ValidatorMetadata[validator]
+	if metadata.ActivationEpoch.Valid {
+		return uint64(metadata.ActivationEpoch.Int64), nil
+	}
+
+	// estimate
+	latestEpoch := cache.LatestFinalizedEpoch.Get()
+	if d.config.ClConfig.ElectraForkEpoch > latestEpoch {
```
LuccaBitfly: **nitpick:** 
```suggestion
	if d.config.ClConfig.ElectraForkEpoch > latestEpoch {
	// pre pectra
```

backend/pkg/api/data_access/validators.go
```diff
@@ -137,3 +139,92 @@ func (d *DataAccessService) GetValidatorsByGraffiti(ctx context.Context, graffit
...
+		}
+		if !metadata.Queues.ActivationIndex.Valid {
+			return 0, fmt.Errorf("validator %d has no activation index", validator)
+		}
+		queuePosition := uint64(metadata.Queues.ActivationIndex.Int64)
+
+		latestStats := cache.LatestStats.Get()
+		activationChurnRate := uint64(4)
+		if latestStats.ValidatorActivationChurnLimit == nil {
+			log.Warnf("Activation Churn rate not set in config, using 4 as default")
```
LuccaBitfly: **issue:** Shouldn't `activationChurnRate` be initialized with `d.config.ClConfig.MinPerEpochChurnLimit`? Will this value ever be anything other than what's written in the config? If so, why not get it directly from the config? If not, why does the error message say it's not set in the config? If it's not set in the config, shouldn't this be an internal error?

backend/pkg/api/data_access/validators.go
```diff
@@ -10,7 +10,9 @@ import (
 	"github.com/ClickHouse/clickhouse-go/v2/ext"
 	"github.com/doug-martin/goqu/v9"
 	t "github.com/gobitfly/beaconchain/pkg/api/types"
+	"github.com/gobitfly/beaconchain/pkg/commons/cache"
```
LuccaBitfly, line 13: **nitpick:** Prob not your fault, but I don't like that we're relying on some global cache variable. I don't even understand where the churn rate is written into the cache.

backend/pkg/api/data_access/validators.go
```diff
@@ -137,3 +139,92 @@ func (d *DataAccessService) GetValidatorsByGraffiti(ctx context.Context, graffit
...
+	balance := metadata.Balance
+	upwardThreshold := utils.Config.Chain.ClConfig.EffectiveBalanceIncrement / utils.Config.Chain.ClConfig.HysteresisQuotient * utils.Config.Chain.ClConfig.HysteresisUpwardMultiplier
+	for _, deposit := range results {
+		balance += deposit.Amount
+		if effectiveBalance+upwardThreshold < balance {
+			effectiveBalance = balance - balance%utils.Config.Chain.ClConfig.EffectiveBalanceIncrement
+			if effectiveBalance >= utils.Config.Chain.ClConfig.MinActivationBalance {
+				return deposit.ClearEpoch, nil
+			}
+		}
```
LuccaBitfly: **question:** Do these if's need to be nested? Could we do this?
```go
		if effectiveBalance+upwardThreshold < balance {
			effectiveBalance = balance - balance%utils.Config.Chain.ClConfig.EffectiveBalanceIncrement
		}
		if effectiveBalance >= utils.Config.Chain.ClConfig.MinActivationBalance {
			return deposit.ClearEpoch, nil
		}
```

backend/pkg/api/data_access/validators.go
```diff
@@ -137,3 +139,92 @@ func (d *DataAccessService) GetValidatorsByGraffiti(ctx context.Context, graffit
...
+	}
+
+	return 0, fmt.Errorf("validator %d has not enough pending ETH deposits", validator)
```
LuccaBitfly: **question:** Do we want to throw an error here? This would result in a 500 response. Please confirm that we want that.

LuccaBitfly (changes requested):

backend/pkg/api/data_access/validators.go
remoterami: I suggested `getValidatorActivationEpochs`, as it only tries to return estimates when no activation epoch is available yet (and if possible)

backend/pkg/api/data_access/validators.go
remoterami: Yeah I think you're correct, should use config value as default, will change.
> If so, why not get it directly from the config?

Because it depends on chain state. In particular it depends on the amount of total staked ETH iirc, so we can't read this from config. The config only contains the min (and I think max?) value.

backend/pkg/api/data_access/validators.go
remoterami: Because we need the latest churn rate which is currently used in the network as a basis for estimate calculation. See explanation in above comment for context and here's the link to the beacon spec for more info https://github.com/ethereum/consensus-specs/blob/master/specs/phase0/beacon-chain.md#get_validator_churn_limit

backend/pkg/api/data_access/validators.go
remoterami: Prob not, I removed it, ptal

backend/pkg/api/data_access/validators.go
LuccaBitfly: Sure, my problem is more with the fact that we're relying on global state, not that we're getting current network data. Again, not your fault.

backend/pkg/api/data_access/validators.go
```diff
@@ -137,3 +139,120 @@ func (d *DataAccessService) GetValidatorsByGraffiti(ctx context.Context, graffit
 
 	return runQueryRows[[]t.VDBValidator](ctx, d.readerDb, validatorsDs)
 }
+
+// fills the activation epochs for a set of validators. Tries to estimate pending validators based on current queue simulation
+func (d *DataAccessService) getValidatorActivationEpochs(ctx context.Context, validators map[uint64]*uint64) error {
```
LuccaBitfly, line 144: **nitpick (non-blocking):** I'd probably do this to avoid dealing with pointers, but no strong opinion
```suggestion
func (d *DataAccessService) getValidatorActivationEpochs(ctx context.Context, validators []uint64) (map[uint64]uint64, error) {
```

backend/pkg/api/data_access/validators.go
```diff
@@ -137,3 +139,120 @@ func (d *DataAccessService) GetValidatorsByGraffiti(ctx context.Context, graffit
...
+				// probably not enough deposits yet
+				continue
+			}
+			if !metadata.Queues.ActivationIndex.Valid {
+				// could support estimates for validators without index
+				continue
+			}
+			queuePosition := uint64(metadata.Queues.ActivationIndex.Int64)
+
+			epochsToWait := (queuePosition - 1) / activationChurnRate
```
LuccaBitfly, line 187: **thought:** Since we're decrementing `queuePosition` here, pls confirm that `queuePosition` can never be 0. Otherwise, should prob. use `max(queuePosition, 1) - 1`.

backend/pkg/api/data_access/validators.go
```diff
@@ -137,3 +139,120 @@ func (d *DataAccessService) GetValidatorsByGraffiti(ctx context.Context, graffit
...
+		}
+
+		// post pectra
+		if constypes.ValidatorDbStatus(metadata.Status) != constypes.DbDeposited {
+			// should not happen since there's no more activation queue after deposits have been processed (see process_registry_updates)
+			return fmt.Errorf("validator %d has no activation epoch", validator)
+		}
+		pendingValidatorsPostPectra = append(pendingValidatorsPostPectra, validator)
+	}
+
```
LuccaBitfly, line 204: **suggestion:** Maybe an early return for when we're pre pectra?
```suggestion
		pendingValidatorsPostPectra = append(pendingValidatorsPostPectra, validator)
	}
	if len(pendingValidatorsPostPectra) == 0 {return nil}

```

LuccaBitfly (approved):
lgtm, still have minor comments left

backend/pkg/api/data_access/validators.go
remoterami: I see, but I'd avoid it just to save those memory allocations

PR #1615, opened 2025-06-02
(BEDS-1539) re-enable csrf injection

remoterami:


frontend/composables/useStripeProvider.ts
```diff
@@ -25,11 +25,19 @@ export function useStripeProvider() {
     )
   })
 
+  const { apiClient } = useRuntimeConfig().public
+  const csrfToken = ref()
```
marcel-bitfly, line 29: **note:** done here instead of `useCsrfStore` as I am about to remove it

enzo-bitfly: maybe we can ask the backend to add an endpoint to get a csrf after this? 🤷

marcel-bitfly, line 29: **note:** should not have to be a `ref()` but otherwise `typescript` is complaining about `headers: {} -> HeadersInit`  😑

enzo-bitfly (approved):

frontend/composables/useStripeProvider.ts
enzo-bitfly: maybe we can ask the api devs to add an endpoint to get a csrf after this? 🤷

marcel-bitfly: we removed the `csrf logic` but then realized we still need them, since we are using a `v1 api route` here.

So we only need it for `stripe` right now. `/i/api/pricing` returns `{data:null}` and the `csrf token header` we need to call the `route`

enzo-bitfly: I mean basically renaming the endpoint to make it clear what it does

remoterami: yeah the endpoint path might not be finalized yet, open to suggestions. Figured renaming should be an easy change

backend/pkg/api/router.go
```diff
@@ -32,6 +32,13 @@ func NewApiRouter(dataAccessor dataaccess.DataAccessor, dummy dataaccess.DataAcc
 	internalRouter.Use(sessionManager.LoadAndSave, getSlidingSessionExpirationMiddleware(sessionManager))
 
 	handlerService := handlers.NewHandlerService(dataAccessor, dummy, sessionManager, cfg)
+	// CSRF injection
+	if !(cfg.Frontend.CsrfInsecure || cfg.Frontend.Debug) {
+		csrfRouter := internalRouter.NewRoute().Subrouter()
+		csrfRouter.Use(getCsrfProtectionMiddleware(cfg), csrfInjecterMiddleware)
+
+		csrfRouter.HandleFunc("/pricing", handlers.Handle(http.StatusOK, handlerService.InternalGetPricing, false)).Methods(http.MethodGet)
```
LuccaBitfly: **thought (non-blocking):** Could also just do
```suggestion
		csrfRouter.HandleFunc("/pricing", handlers.Handle(http.StatusOK, func (w http.ResponseWriter, r *http.Request) {}).Methods(http.MethodGet)
```

LuccaBitfly (approved):
lgtm

PR #1620, opened 2025-06-03
(BEDS-1540) omitting missing efficiencies

remoterami:


backend/pkg/api/data_access/mobile.go
```diff
@@ -251,22 +252,33 @@ func (d *DataAccessService) GetValidatorDashboardMobileWidget(ctx context.Contex
...
 				Div(queryResult.RPLStake).
 				Mul(periodsPerYear).
 				Mul(decimal.NewFromInt(100)).InexactFloat64()
+			data.RplApr = &rplApr
 		}
 		return nil
 	})
 
-	retrieveApr := func(timeFrame enums.TimePeriod, apr *float64) {
+	retrieveApr := func(timeFrame enums.TimePeriod, apr **float64) {
```
LuccaBitfly: **issue:** Why do we pass double pointers here? Why not leave these funcs as they are, and rather call `retrieveApr(enums.Last30d, data.Last30dApr)`? Goes for all other double pointers in this PR

LuccaBitfly (changes requested):

backend/pkg/api/data_access/mobile.go
remoterami: Because we want to modify the original underlying value of the response struct, just like before. Essentially the first pointer is a reference if you're familiar with that concept from C, and we're just updating the data type from `float64` to `*float64` in order to allow returning undefined values

LuccaBitfly (approved):
lgtm

PR #1628, opened 2025-06-10
perf(migrations): dont proxy postgres tables access, use in-memory dict instead

invis-bitfly:
`$CLICKHOUSE_DB` seems to be sadly unavoidable, need to set the db or otherwise it cant find the dict, and cant use clickhouse native parameters because you cant load those with the output from `currentDatabase()`


invis-bitfly (commented):
lgtm

PR #1631, opened 2025-06-11
fix(DashboardChartSummary): use `null` for `nullable` values

enzo-bitfly:
See: BEDS-1548

PR #1632, opened 2025-06-16
(BEDS-1548) api: remove undefined

remoterami:


marcel-bitfly:
cherrypicked into #1631

PR #1643, opened 2025-07-02
Beds 99/further perf improvements

invis-bitfly:


invis-bitfly (commented):
checked if any usage of the rolling views in the api require FINAL. couldnt find anything, so should be gucci

PR #1644, opened 2025-07-02
feat(api): add endpoint to get truncated guest dashboard validators

LuccaBitfly:


backend/pkg/api/router.go
```diff
@@ -133,7 +133,10 @@ func addRoutes(hs *handlers.HandlerService, publicRouter, internalRouter *mux.Ro
 		{http.MethodPost, "/users/me/email", nil, hs.InternalPostUserEmail},
 		{http.MethodPut, "/users/me/password", nil, hs.InternalPutUserPassword},
 		{http.MethodGet, "/users/me/dashboards", hs.PublicGetUserDashboards, hs.InternalGetUserDashboards},
-		{http.MethodPut, "/users/me/notifications/settings/paired-devices/{client_id}/token", nil, hs.InternalPostUsersMeNotificationSettingsPairedDevicesToken},
```
remoterami, line 136: **question**: why was this changed, some modifications with the app? Seems like the method is not used anymore now, so it could be deleted

remoterami (changes requested):

backend/pkg/api/router.go
LuccaBitfly: Don't know what happened here, not intentional. Fixed, ptal

remoterami (approved):
lgtm

PR #1647, opened 2025-07-04
feat(api): provide flag if user has v1 specific notifications

LuccaBitfly:


backend/pkg/api/data_access/notifications.go
```diff
@@ -2112,3 +2112,18 @@ func (d *DataAccessService) QueueTestPushNotification(ctx context.Context, userI
...
+
+// hasUserV1NotificationSubscriptions checks if a user has any v1 notification subscriptions.
+// Since some notification events are indistinguishable from v1 and v2, only validator-related that are not v2 are checked.
+func (d *DataAccessService) hasUserV1NotificationSubscriptions(ctx context.Context, userId uint64) (bool, error) {
+	ds := goqu.Dialect("postgres").
+		Select(goqu.COUNT(goqu.I("id"))).
+		From(goqu.T("users_subscriptions")).
+		Where(
+			goqu.I("user_id").Eq(userId),
+			goqu.I("event_name").Like("validator%"),
```
manuelsc: **Performance Suggestion:** Instead of using like, we could just use all v1 notification events that are validator based. We probably won't change any v1 notifications in the future any more so this would be a finite list of all the events we are looking for in v1:

```  
AND event_name IN (
  'validator_balance_decreased',
  'validator_proposal_missed',
  'validator_proposal_submitted',
  'validator_attestation_missed',
  'validator_got_slashed',
  'validator_did_slash',
  'validator_is_offline',
  'validator_withdrawal',
  'validator_received_deposit',
  'validator_synccommittee_soon'
)

manuelsc: **observation (prob nothing to do):** While looking at all our events, I noticed `user_tax_report`. I guess you can't change or set this notification up in v2? 
But this is a curious one as you apparently can not manage it from the v1 notification settings page either. Instead it seems to only be available on this sub-site https://beaconcha.in/user/rewards where you enable/disable it.

Will briefly check with product but I think it's fine if we keep this separate (aka no change here), as it already is it's own thing.

LuccaBitfly: fixed, ptal

manuelsc (approved):
lgtm

PR #1655, opened 2025-07-22
API Key DB Schema

manuelsc:


backend/pkg/commons/db/migrations/postgres/20250722110556_api_keys.sql
```diff
@@ -0,0 +1,33 @@
+-- +goose Up
+-- +goose StatementBegin
+
+CREATE EXTENSION IF NOT EXISTS pgcrypto;
+
+CREATE TABLE user_api_keys (
```
mccreedy-bitfly: Why the rename to "user" api_keys? Personally I think this is extra information we don't need, and I guess I'd prefer simply "api_keys"

backend/pkg/commons/db/migrations/postgres/20250722110556_api_keys.sql
```diff
@@ -0,0 +1,33 @@
+-- +goose Up
+-- +goose StatementBegin
+
+CREATE EXTENSION IF NOT EXISTS pgcrypto;
```
mccreedy-bitfly, line 4: Whats this and why do we need it?

backend/pkg/commons/db/migrations/postgres/20250722110556_api_keys.sql
```diff
@@ -0,0 +1,33 @@
...
+
+CREATE TABLE user_api_keys (
+    id UUID PRIMARY KEY DEFAULT gen_random_uuid(),
```
mccreedy-bitfly: Can we instead call this `key_id`? While `id` will work just fine and that pattern is pretty common, I prefer having explicit names for the id columns so that it's easier to grep the code for them and also to identify which keys you are talking about if you ever need to join tables and stuff.

backend/pkg/commons/db/migrations/postgres/20250722110556_api_keys.sql
```diff
@@ -0,0 +1,33 @@
...
+    last_used_at TIMESTAMP,
+    disabled_at TIMESTAMP,
+    deleted_at TIMESTAMP
+);
+
+CREATE UNIQUE INDEX uniq_user_api_keys_name_active
+ON user_api_keys(user_id, name)
+WHERE deleted_at IS NULL;
+
+CREATE INDEX idx_user_api_keys_api_key ON user_api_keys(api_key);
```
mccreedy-bitfly: Can we make this a unique index

backend/pkg/commons/db/migrations/postgres/20250722110556_api_keys.sql
peterbitfly, line 4: where is this extension used?

backend/pkg/commons/db/migrations/postgres/20250722110556_api_keys.sql
```diff
@@ -0,0 +1,33 @@
+-- +goose Up
+-- +goose StatementBegin
+
+CREATE EXTENSION IF NOT EXISTS pgcrypto;
+
+CREATE TABLE user_api_keys (
+    id UUID PRIMARY KEY DEFAULT gen_random_uuid(),
+    user_id BIGINT NOT NULL REFERENCES users(id),
```
peterbitfly, line 7: user_id might become a string in the future (if we decide to implement an external auth solution)

peterbitfly: note: this is nothing we need to deal with right now as if we migrate at some point we will have this issue in more places than here

backend/pkg/commons/db/migrations/postgres/20250722110556_api_keys.sql
manuelsc: Fits better in our current db structure, we have `users` and tables that depend on it directly have so far been placed in the same scope aka same prefix. For example `users_subscriptions`, `users_devices`, `users_app_subscriptions`, `users_notification_channels` etc.

backend/pkg/commons/db/migrations/postgres/20250722110556_api_keys.sql
manuelsc: Needed for gen_random_uuid()

backend/pkg/commons/db/migrations/postgres/20250722110556_api_keys.sql
manuelsc: Would try to bring this up with the team, it's opinionated discussion that probably affects everyone how we define it in the future

backend/pkg/commons/db/migrations/postgres/20250722110556_api_keys.sql
mccreedy-bitfly: I don't think it makes sense to double-describe foreign key constraints in the table name for many reasons: 

1. Constraints can change, but table names can't (technically can but its hard because code uses the table names)
2. Tables can have many foreign keys, and so doesnt make sense to have it in the name
3. Our "UserDb" as we call it today is going to end up being more of an "AdminDB", which maintains many records, many of which are user related. So prefixing every single table with the "user" string is going to sorta remove any benefit of having it, it's basically a dead prefix.

backend/pkg/commons/db/migrations/postgres/20250722110556_api_keys.sql
mccreedy-bitfly: I'd like to avoid opening up a discussion unless we feel it is extremely important. Feel free to start a thread, but let's not waste too much time here. I strongly believe calling it explicitly "key_id" is the way to go.

backend/pkg/commons/db/migrations/postgres/20250722110556_api_keys.sql
mccreedy-bitfly: I also believe user_id should become a string. Even if it is actually a number, it should always be represented as a string because to treat it as an integer can introduce math-related unintentional bugs. However, I don't think it's in-scope now as you said

mccreedy-bitfly (approved):

PR #1675, opened 2025-08-13
fix: threshold sorting

remoterami:


sasha-bitfly (approved):

backend/pkg/api/data_access/notifications.go
```diff
@@ -754,14 +754,9 @@ func (d *DataAccessService) GetMachineNotifications(ctx context.Context, userId
 		ds = ds.Where(directions)
 	}
 
-	query, args, err := ds.Prepared(true).ToSQL()
-	if err != nil {
-		return nil, nil, fmt.Errorf("error preparing machine notifications query: %w", err)
-	}
-
-	err = d.userReader.SelectContext(ctx, &notificationHistory, query, args...)
+	notificationHistory, err := runQueryRows[[]NotificationHistory](ctx, d.userReader, ds)
```
sasha-bitfly, line 757: Can we write a small test to cover this call with a unit test?

PR #1679, opened 2025-08-19
API 54/search types

LuccaBitfly:
- **refactor: move search handler funcs into `searchTypeMap`**
- **feat: add search for address, transaction, block, epoch and token**


backend/pkg/api/data_access/user.go
```diff
@@ -433,6 +434,9 @@ func (d *DataAccessService) GetUserDashboards(ctx context.Context, userId uint64
 
 		result.ValidatorDashboards = append(result.ValidatorDashboards, *validatorDashboard)
 	}
+	slices.SortFunc(result.ValidatorDashboards, func(a, b t.ValidatorDashboard) int {
+		return int(a.Id) - int(b.Id)
+	})
```
LuccaBitfly, line 439: unrelated to ticket, but sneaking it into this PR (don't tell the reviewer)

backend/pkg/api/data_access/search.go
```diff
@@ -131,3 +140,90 @@ func (d *DataAccessService) GetSearchValidatorsByGraffitiHex(ctx context.Context
...
+		return nil, fmt.Errorf("failed to search for address %s: %w", hexutil.Encode(address), err)
+	}
+	if len(eth1AddressSearchItem) == 0 || eth1AddressSearchItem[0] == nil {
+		return nil, ErrNotFound
+	}
+	foundAddress := *eth1AddressSearchItem[0]
+	return &t.SearchAddress{
+		Address: t.Address{
+			Hash:       t.Hash("0x" + foundAddress.Address),
+			Label:      foundAddress.Name,
```
remoterami: **check**: I'm not sure if this is the intended label we want to return here. Didn't dig too deep but I fear it could be the contract name or smth. Afaik it should be a team-defined string instead, it's more of a new concept which isn't really used yet; I used it like this before, fetching it from psql: https://github.com/gobitfly/beaconchain/blob/API-54/search-types/backend/pkg/api/data_access/general.go#L39

backend/pkg/api/data_access/search.go
```diff
@@ -131,3 +140,90 @@ func (d *DataAccessService) GetSearchValidatorsByGraffitiHex(ctx context.Context
...
+			IsContract: foundAddress.Token != "",
```
remoterami: **issue**: this shouldn't only look at tokens, there are other contracts, too. See above message how to fetch the contract state

Would also be fine for me to omit these 2 fields for now and add a TODO, but shouldn't be wrong values; same applies to `GetSearchToken`

backend/pkg/api/handlers/search.go
```diff
@@ -31,55 +31,102 @@ const (
...
 type searchType struct {
 	regex        *regexp.Regexp
 	responseType string
+	handlerFunc  func(ctx context.Context, h *HandlerService, input string, chainId uint64) (*types.SearchResult, error)
+}
+
+var searchTypeMap map[searchTypeKey]searchType
+
+// using init to avoid initialization cycles, since handler functions may reference the map
+func init() {
```
remoterami, line 51: **praise**: nice to have this, the handler as a property makes it more clear what's happening

backend/pkg/api/data_access/search.go
```diff
@@ -131,3 +140,90 @@ func (d *DataAccessService) GetSearchValidatorsByGraffitiHex(ctx context.Context
 	}
 	return ret, nil
 }
+
+func (d *DataAccessService) GetSearchAddress(ctx context.Context, chainId uint64, address []byte) (*t.SearchAddress, error) {
+	eth1AddressSearchItem, err := d.bigtable.SearchForAddress(address, 1)
```
remoterami, line 145: **comment**: this will only find erc20 contracts, not erc721 or erc1155 as indicated in the interface definition. It's not trivial to retrieve that data afaik so fine for me to skip for now, but please add open TODOs and indicate this in the interface for frontend (comment and/or limit the enum)

backend/pkg/api/types/search.go
```diff
@@ -2,6 +2,12 @@ package types
 
 // search types to be used between the data access layer and the api layer, shouldn't be exported to typescript
 
+type PostSearchRequest struct {
+	Input    string        `json:"input"`
+	Networks []interface{} `json:"networks,omitempty" tstype:"(number | string)[]"`
+	Types    []string      `json:"types,omitempty"`
+}
+
```
remoterami: **suggestion**: This comment seems outdated please remove, previously the response types were not meant to be exported but now they obv should be.

Also we rarely export the request types but I guess it won't hurt.

remoterami (changes requested):
left some comments, ptal

backend/pkg/api/types/search.go
LuccaBitfly: Yeah FE explicitly asked for the search request type

backend/pkg/api/data_access/search.go
LuccaBitfly: `GetNamesAndEnsForAddresses` assumes that the addresses in the input already exist. I'm not stuck on having the label or token fields be filled, so I'd just remove them for lack of a better way of fetching address metadata.

backend/pkg/api/data_access/search.go
LuccaBitfly: I'll omit the fields for searching addresses. For tokens, since I early return when it's not a token and tokens are always a contract (correct me if I'm wrong) I'll leave them in.

backend/pkg/api/data_access/search.go
LuccaBitfly: done, ptal

backend/pkg/api/data_access/search.go
LuccaBitfly: done, ptal

backend/pkg/api/data_access/search.go
LuccaBitfly: done, ptal

backend/pkg/api/types/search.go
LuccaBitfly: done, ptal

frontend/types/api/search.ts
```diff
@@ -26,11 +32,27 @@ export interface SearchValidatorsByGraffiti {
...
+}
+export interface SearchBlock {
+  block_number: number /* uint64 */;
+}
+export interface SearchEpoch {
+  epoch: number /* uint64 */;
+}
+export interface SearchToken {
+  address: Address;
+  token: 'erc20' | 'erc721' | 'erc1155'; // currently only erc20 tokens can be found
```
enzo-bitfly: question: can tokens be found by name too or just by address?
and if i get this right, we cannot find e.g. USDC for now?

remoterami: USDC is an [ERC20 token](https://etherscan.io/token/0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48#readProxyContract), so you should be able to find this.

But only via address yes, searching via name depends on what `name` should refer to. If it's the contract name, it's likely not a good idea as anyone can name their contract however they want, so by searching for USDC you would probably get a list of 10k scam contracts and 1 legit one.
If `name` refers to some internal labeling system (which we've had ideas for floating around for a while) it would work in theory, but at least I don't see how that would be working by looking at the code, and it needs to be actively maintained.
If it refers to something else, it should be specified.

enzo-bitfly: thanks, today I learnt.
I meant searching for the token name (like Tether, USDC, etc).

backend/pkg/api/data_access/search.go
enzo-bitfly: question: Will we find addresses by ENS too?

backend/pkg/api/data_access/search.go
```diff
@@ -131,3 +140,91 @@ func (d *DataAccessService) GetSearchValidatorsByGraffitiHex(ctx context.Context
...
+	if foundAddress.Token == "" {
+		return nil, ErrNotFound
+	}
+	return &t.SearchToken{
+		Address: t.Address{
+			Hash:       t.Hash("0x" + foundAddress.Address),
+			IsContract: true,
+			Label:      foundAddress.Name,
+		},
+		Token: foundAddress.Token,
```
remoterami, line 228: nitpick: The [bigtable func](https://github.com/gobitfly/beaconchain/blob/API-54/search-types/backend/pkg/commons/db/bigtable_eth1.go#L2361) will return `ERC20` while the API specifies the tstype to be `erc20`, should streamline the capitalization

remoterami (approved):
left a small comment, otherwise utack

backend/pkg/api/data_access/search.go
LuccaBitfly: fixed, ptal

PR #1683, opened 2025-08-21
fix: truncated guest dashboard

LuccaBitfly:
Fixes bug where list that is equal to the max allowed EB would get truncated, when it shouldn't


remoterami (approved):
lgtm

PR #1684, opened 2025-08-21
feat: add premium tiers table

manuelsc:


backend/pkg/commons/db/migrations/postgres/20250819120437_premium_tiers.sql
```diff
@@ -0,0 +1,17 @@
+-- +goose Up
+-- +goose StatementBegin
+
+CREATE TABLE IF NOT EXISTS premium_tiers (
+    purchase_id VARCHAR(255) NOT NULL,
```
mccreedy-bitfly: subscription_id

backend/pkg/commons/db/migrations/postgres/20250819120437_premium_tiers.sql
```diff
@@ -0,0 +1,17 @@
...
+    tier_id VARCHAR(255) NOT NULL,
```
mccreedy-bitfly: tier_name

mccreedy-bitfly: Add constraints enforcing the tier names which match the code types

backend/pkg/commons/db/migrations/postgres/20250819120437_premium_tiers.sql
```diff
@@ -0,0 +1,33 @@
...
+-- create enum type for tier_name
+DO $$
+BEGIN
+    IF NOT EXISTS (SELECT 1 FROM pg_type WHERE typname = 'tier_enum') THEN
+        CREATE TYPE tier_enum AS ENUM ('FREE', 'HOBBYIST', 'BUSINESS', 'SCALE');
+    END IF;
+END$$;
+
+CREATE TABLE IF NOT EXISTS subscription_tiers (
+    price_id VARCHAR(255) NOT NULL,
```
mccreedy-bitfly, line 21: Call this explicitly `stripe_*`, because then whoever sees this variable is going to know exactly where it comes from. Otherwise, they have to guess.

mccreedy-bitfly: Also, I thought we were going to rename this to `stripe_subscription_id`. Are we sure `price` is right?

manuelsc: Yeah it was my bad, I confused subscription id and price id in the call but checked after to make sure I get this right. Price ID is the correct one (identifying what the user purchased, subscription is is just the id of that specific user subscription).

As for the stripe prefix. In theory we could use this table for a generic implementation. It's just a mapping between any third party identifier of a thing a user can by to our local reference for it. That's why I'd argue against tying it to strongly to stripe

backend/pkg/commons/db/migrations/postgres/20250819120437_premium_tiers.sql
```diff
@@ -0,0 +1,33 @@
...
+    tier_name tier_enum NOT NULL,
+    PRIMARY KEY (price_id)
```
mccreedy-bitfly, line 24: The primary key should be tier_name, not price_id, right? Otherwise it could be possible to accidentally add in multiple tiers with the same name but different price_ids, and the code will break.

mccreedy-bitfly (approved):

PR #1685, opened 2025-08-21
Df 7/guest dashboard truncation

marcel-bitfly:
> [!WARNING]  
> this PR will assume that the application will move from the `beta url` to `v1 domain` (beaconcha.in)


frontend/composables/useDashboardKeyProvider.ts
```diff
@@ -91,6 +100,14 @@ export function useDashboardKeyProvider(
...
     const filtered = list.filter(s => !!s).join(',')
     const key = encodeBase64Url(filtered)
+    if ('isTruncated' in route.query) {
+      router.push({
+        query: {
+          ...route.query,
+          isTruncated: undefined,
+        },
+      })
+    }
```
enzo-bitfly, line 110: question: is this still necessary after adding it in `setDashboardKey`?

frontend/middleware/dashboard-max-eb-param.ts
enzo-bitfly, line 1: would a name like `dashboard-truncation-from-param` be more explicit?

frontend/i18n/locales/en.json
```diff
@@ -215,6 +215,10 @@
             }
         },
         "title": "Dashboard",
+        "truncated_validators": {
+            "_link": "Learn more about how to upgrade and unlock full validator insights and premium features on our pricing page.",
+            "template": "This dashboard displays a truncated list of validators, as their maximum stake exceeds the free plan limit. {_link}"
```
enzo-bitfly: suggestion: change `maximum` for `total`, as the `maximum stake` is actually the `plan limit`, and here we mean the `accumulated validator stake` in the dashboard.

frontend/components/bc/DataWrapper.vue
```diff
@@ -14,9 +13,9 @@ await useAsyncData('latest_state', () => refreshLatestState(), {
   immediate: true,
   watch: [ counter ],
 })
-if (isLoggedIn) {
-  await useAsyncData('get_user', () => getUser())
-}
+await useAsyncData('get_user', () => getUser(), {
```
enzo-bitfly: suggestion: use `callOnce` because we get the user from the pinia store after this runs

enzo-bitfly (approved):
nice work

frontend/i18n/locales/en.json
marcel-bitfly: @Buttaa approves thx

PR #1690, opened 2025-08-26
fix/truncate guest dashboard 2

LuccaBitfly:
- **fix: fetch eb at exit time for truncate guest dashboard endpoint**
- **fix: enforce strict comparison for unarchiving dashboard eb check**


remoterami (approved):
lgtm

PR #1702, opened 2025-08-28
fix: cap chart timestamps to min and max possible ts

LuccaBitfly:
See: BEDS-875


remoterami (approved):
honestly not a big fan of that function, ie. the switching, variable names and min/max calculations. But lgtm and prob not worth to beautify, approved

PR #1713, opened 2025-09-09
API 54/search types 2

LuccaBitfly:
- **feat: add search for address, transaction, block, epoch and token**
- **feat: add search ts types for address, transaction, block, epoch and token**
- **feat: add search for ens name, slot by block & state root**


backend/pkg/api/data_access/search.go
```diff
@@ -158,6 +163,45 @@ func (d *DataAccessService) GetSearchAddress(ctx context.Context, chainId uint64
...
+	}
+	return &t.SearchAddress{
+		Address: t.Address{
+			Hash: t.Hash("0x" + hexutil.Encode(addressBytes)),
+			Ens:  ensName,
+		},
+	}, nil
+}
+
+func (d *DataAccessService) GetSearchEnsName(ctx context.Context, chainId uint64, ensName string) (*t.SearchEnsName, error) {
```
manuelsc, line 191: Question: What's the usecase for this, why not use `GetSearchAddressByEnsName` that returns the same info + more?

backend/pkg/api/data_access/search.go
```diff
@@ -187,6 +231,57 @@ func (d *DataAccessService) GetSearchBlock(ctx context.Context, chainId uint64,
...
+	return &t.SearchSlot{
+		Slot: slot,
+	}, nil
+}
+
+func (d *DataAccessService) GetSearchSlotByBlockRoot(ctx context.Context, chainId uint64, blockRoot []byte) (*t.SearchSlot, error) {
+	ds := goqu.Dialect("postgres").
+		Select(goqu.I("slot")).
+		From("blocks").
+		Where(goqu.I("blockroot").Eq(blockRoot))
```
manuelsc: Issue: Blocks table has no index for just `blockroot` but only in combination with status or slot. Depending on the usage of what we try to cover with this endpoint, it might be an option to restrict it to just searching by blockroot & `proposed` status (no missed/orphaned) or add a new index if there is a usecase  for missed & oprhaned

backend/pkg/api/data_access/search.go
LuccaBitfly: Searching for an ENS name should give you 2 search results: one that leads to the resolved addresses' page and one that leads to the registration overview.
image
It's true that this is a bit redundant but I imagine the extra DB load is not egregious (it should be the same load as v1 actually). In future we could probably optimize with `sync.Once` or something.

backend/pkg/api/data_access/search.go
LuccaBitfly: added proposed status to the query, ptal

backend/pkg/api/data_access/search.go
manuelsc: Will not block on this, but it seems odd that we introduce redundancy in the API when this is basically a frontend "how do they use the API" issue. They could use `GetSearchAddressByEnsName` for both usecases right? But I guess we probably have our reasons to help frontend out here

manuelsc (approved):
lgtm

PR #1721, opened 2025-09-12
chore: remove obsolete pages and add prefix to `primevue` components

enzo-bitfly:


frontend/utils/mobileAuth.ts
marcel-bitfly, line 1: **issue:** this `util` is used by `mobile app v2`

@manuelsc are you ok with us removing this?
You can always recover this via `git`

enzo-bitfly: In our project it was only used in the `register` page, which we never used and I remove in this PR

PR #1732, opened 2025-09-22
DF-97/add section stacking hub

rucister:


frontend/layers/base/app/assets/css/main.css
```diff
@@ -86,7 +87,10 @@
   --text-md--line-height: 1.5rem; /* 24px */
   --text-lg: 1.125rem; /* 18px */
   --text-lg--line-height: 1.75rem; /* 28px */
-
+  --text-xl: 1.25rem; /* 20px */
+  --text-xl--line-height: 1.5rem; /* 24px */
+  --text-2xl: 1.5rem ; /* 24px */
+  --text-2xl--line-height: 2rem; /* 32px */
   --text-2xl: 1.5rem ; /* 24px */
   --text-2xl--line-height: 2rem; /* 32px */
```
marcel-bitfly: **issue:** plz remove this lines as they are already there

frontend/layers/base/app/components/BaseButton.vue
```diff
@@ -1,29 +1,68 @@
...
+  variant === 'branded' && 'from-brand-500 to-brand-700 text-white hover:opacity-90',
+  size === 'md' && 'text-sm py-md px-sm',
+  size === 'xl' && 'text-md py-xl px-3xl',
+  full ? 'w-full' : 'w-fit',
+])
 </script>
 
 <template>
-  <button
-    class="flex justify-center items-center bg-linear-to-b rounded-full font-semibold disabled:opacity-40 aria-disabled:opacity-40 active:opacity-80 focus-visible:outline-2 focus-visible:outline-offset-2 focus-visible:outline-brand-300"
```
marcel-bitfly, line 18: **issue:** as we spoke about we want to go with putting as much as possible in the markup instead of the `script`

frontend/layers/base/app/components/BaseButton.vue
```diff
@@ -1,29 +1,68 @@
...
+  disabled?: boolean,
   full?: boolean,
   leadingIcon?: IconName,
   size?: 'md' | 'xl',
+  to?: NuxtLinkProps['to'],
   trailingIcon?: IconName,
   variant?: 'branded' | 'primary' | 'secondary',
 }>()
+
+// Determine component type
```
marcel-bitfly: **thought:** this code comment is not talking about the `why` and probably should go

frontend/layers/base/app/components/BaseButton.vue
```diff
@@ -1,29 +1,68 @@
...
+  if (to) {
+    // Link props
+    baseProps.to = to
+  }
+
+  if (disabled) {
+    if (to) {
+      // For links: use aria-disabled + tabindex
+      baseProps['aria-disabled'] = 'true'
+      baseProps.tabindex = -1
```
marcel-bitfly: **thought:** having a `disabled link` is a interessting frontend problem, which you may solve when the situation occurs. I would for now go with conditional types (removing the possiblity to add `to` and `disabled/aria-disabled` at the same time

frontend/layers/base/app/components/BaseCard.vue
```diff
@@ -0,0 +1,52 @@
...
+    | {
+      title?: never,
+      titleIs?: never,
+    }
+  )
+>()
+</script>
+
+<template>
+  <div class="bg-gray-950 p-3xl rounded-4xl flex flex-col gap-2xl">
```
marcel-bitfly: **todo:** since the `light mode` is not ready from a designer part yet, we agreed on putting the `dark modifier` only and put in the `light version` at a later point
```suggestion
  <div class="dark:bg-gray-950 p-3xl rounded-4xl flex flex-col gap-2xl">
```

frontend/layers/base/app/components/BaseCard.vue
```diff
@@ -0,0 +1,52 @@
...
+        <LazyBaseHeading
+          :is="titleIs"
+          v-if="title && titleIs"
+          size="xs"
+        >
+          {{ title }}
+        </LazyBaseHeading>
+      </slot>
+    </div>
+
```
marcel-bitfly: **nitpick:** 👀

frontend/layers/products/app/components/ProductLandingpageSectionStakinghub.vue
```diff
@@ -0,0 +1,172 @@
...
+const { url } = useV1Login()
+
+// Video ref and state
+const videoRef = useTemplateRef('videoRef')
+
+// Play video with error handling
+const playVideo = async () => {
+  if (videoRef.value) {
+    try {
+      await videoRef.value.play()
```
marcel-bitfly, line 9: **question:** plz check if there is the following issue for `ios Safari`: autoplayed videos sometimes tend to be full-size and make the `page` hard to use (esp. if there a several autoplayed videos)

frontend/layers/products/app/components/ProductLandingpageSectionStakinghub.vue
```diff
@@ -0,0 +1,172 @@
...
+    }
+    catch {
+      // Handle autoplay restrictions silently
+    }
+  }
+}
+
+// Intersection Observer - play once when entering viewport
+const { stop: stopIntersectionObserver } = useIntersectionObserver(
```
marcel-bitfly, line 18: **nitpick:** can you think about wrapping this third party code (vueuse) into its own composable?

frontend/layers/products/app/components/ProductLandingpageSectionStakinghub.vue
```diff
@@ -0,0 +1,172 @@
...
+  videoRef,
+  ([ entry ]) => {
+    if (entry && entry.isIntersecting) {
+      playVideo()
+    }
+  },
+  {
+    rootMargin: '0px 0px -50px 0px',
```
marcel-bitfly, line 26: **question:** : 👀 does this also work if the user changes `font-size` (rem instead of px?)

frontend/layers/products/app/components/ProductLandingpageSectionStakinghub.vue
```diff
@@ -0,0 +1,172 @@
...
+    <BaseButton
+      leading-icon="coins"
+      trailing-icon="arrow-up-right"
+      variant="branded"
+      size="xl"
+      :to="url"
+    >
+      {{ $t('products.landing_page.staking_hub.action.go_to_stakinghub') }}
+    </BaseButton>
+    <div class="grid grid-cols-1 md:grid-cols-3 md:flex-row gap-xl min-h-[495px] lg:min-h-[620px]">
```
marcel-bitfly: **question:** should we put it into semantic `css variables`? (e.g. `var(--landing-page-card-height)`)

frontend/layers/products/app/components/ProductLandingpageSectionStakinghub.vue
```diff
@@ -0,0 +1,172 @@
...
+      <BaseCard
+        title-is="h3"
+        title-icon="file-code"
+        :title="$t('products.landing_page.staking_hub.cards.notifications.title')"
+        class="min-h-[495px]"
```
marcel-bitfly: **question:** is it enough to put it on the container?

frontend/layers/products/app/components/ProductLandingpageSectionStakinghub.vue
```diff
@@ -0,0 +1,172 @@
...
+        <p class="text-md font-semibold">
+          {{ $t('products.landing_page.staking_hub.cards.notifications.subtitle') }}
+        </p>
+        <p class="text-sm">
+          {{ $t('products.landing_page.staking_hub.cards.notifications.description') }}
+        </p>
+        <div class="flex-grow px-5xl flex items-center justify-center">
+          <BaseIcon
+            name="bell"
+            class="w-[100px] h-[100px]"
```
marcel-bitfly: **question:**  what should happen for `bigger font sizes`?

```suggestion
            class="size-[100px]"
```

frontend/layers/products/app/components/ProductLandingpageSectionStakinghub.vue
```diff
@@ -0,0 +1,172 @@
...
+            {{ $t('products.landing_page.staking_hub.cards.notifications.action') }}
+          </BaseButton>
+        </template>
+      </BaseCard>
+      <BaseCard
+        title-is="h3"
+        title-icon="file-code"
+        :title="$t('products.landing_page.staking_hub.cards.staking_mobile_app.title')"
+        class="min-h-[495px] relative overflow-hidden"
+        tabindex="0"
```
marcel-bitfly, line 99: **issue:** 👀

frontend/layers/products/app/components/ProductLandingpageSectionStakinghub.vue
```diff
@@ -0,0 +1,172 @@
...
+        @mouseenter="playOnInteraction"
+        @focus="playOnInteraction"
+      >
+        <video
+          ref="videoRef"
+          poster="/assets-2usdf/img/rotato-frame.jpg"
```
marcel-bitfly: **question:** did you try to use `.webp` as a format? could be smaller

frontend/layers/products/app/components/ProductLandingpageSectionStakinghub.vue
```diff
@@ -0,0 +1,172 @@
...
+          class="absolute top-[0] left-[0] w-full h-full object-cover z-0"
```
marcel-bitfly, line 106: **suggestion (non-blocking):** you could also try doing the stacking with grid

frontend/layers/products/app/components/ProductLandingpageSectionStakinghub.vue
```diff
@@ -0,0 +1,172 @@
...
+          muted
+          playsinline
```
marcel-bitfly, line 108: **praise:** nice did not know about that 💪

frontend/layers/products/app/components/ProductLandingpageSectionStakinghub.vue
```diff
@@ -0,0 +1,172 @@
...
+        <template #footer>
+          <div class="flex gap-2xl justify-center items-center relative z-10">
+            <NuxtLink
+              to="https://apps.apple.com/app/beaconchain-dashboard/id1541822121"
+              target="_blank"
+            >
+              <img
+                src="/assets-2usdf/img/app-store-btn.svg"
+                alt="Download on App Store"
+                class="w-auto h-[41px]"
```
marcel-bitfly: **thought:** things should only be changed by `width` and sometimes `aspect-ratio` and almost never with `height`

frontend/layers/products/app/components/ProductLandingpageSectionStakinghub.vue
```diff
@@ -0,0 +1,172 @@
...
+        </div>
+        <template #footer>
+          <div class="flex gap-2xl justify-center items-center relative z-10">
+            <NuxtLink
+              to="https://apps.apple.com/app/beaconchain-dashboard/id1541822121"
+              target="_blank"
+            >
+              <img
+                src="/assets-2usdf/img/app-store-btn.svg"
+                alt="Download on App Store"
```
marcel-bitfly: **issue:** `alt text` also has to be translated

marcel-bitfly: better phrase it grammatically better like:

`Download the Beaconchain Dashboard App on the App Store`

frontend/layers/products/app/components/ProductLandingpageSectionTestimonials.vue
```diff
@@ -0,0 +1,36 @@
+<template>
+  <div class="py-3xl w-full">
+    <BaseCard>
+      <div class="flex flex-col text-center gap-y-4xl">
+        <p class="text-xl">
+          {{ $t('products.landing_page.testimonials.title') }}
+        </p>
+        <div class="flex flex-wrap sm:flex-row gap-x-5xl gap-y-3xl justify-center">
+          <!-- Replace with real logos -->
```
marcel-bitfly: **nitpick:** unnessesary I think

frontend/layers/products/app/components/ProductLandingpageSectionStakinghub.vue
marcel-bitfly: **issue**: you should probably try out what happens here for the `screen reader` and the `alt text` inside the link.
The safer route would be skip the `alt text` but but in a `sr-only` span

frontend/public/assets-2usdf/img/app-store-btn.svg
marcel-bitfly, line 1: **nitpick:** plz always minify `resources` 

I would use https://svgomg.net/

frontend/public/assets-2usdf/img/rotating-mobile.mp4
marcel-bitfly, line 1: **nitpick:** maybe a semantic name would be a good idea

marcel-bitfly (commented):
very good work 💪 

(esp. considering that this is your first issue)

I am looking forward to see more of your work

frontend/layers/products/app/components/ProductLandingpageSectionStakinghub.vue
marcel-bitfly: maybe make sure that the `focus style` also matches the `custom button focus style`

frontend/layers/base/app/components/BaseButton.vue
rucister: Refactored this to enforce for now not using `to` and `disabled` at the same time.

frontend/layers/base/app/components/BaseButton.vue
rucister: Done, put everything in template

frontend/layers/products/app/components/ProductLandingpageSectionStakinghub.vue
rucister: There is a potential issue in iOS safari where video might not play at all when entering viewport. I would need Tanya to test this. If the issue is confirmed then I need to implement an alternative approach with animated frames.

frontend/layers/products/app/components/ProductLandingpageSectionStakinghub.vue
rucister: Implemented the much cleaner variable approach

frontend/layers/products/app/components/ProductLandingpageSectionStakinghub.vue
rucister: It is required also on the card itself for mobile view where the card needs to have its own height to be compatible with deisng.

frontend/layers/products/app/components/ProductLandingpageSectionStakinghub.vue
rucister: Changed to webp

frontend/public/assets-2usdf/img/app-store-btn.svg
rucister: Minified using https://svgomg.net/

frontend/layers/products/app/components/ProductLandingpageSectionStakinghub.vue
rucister: In this case it is more straight-forward to use height in order to ensure exact compatibility with design.

frontend/layers/products/app/components/ProductLandingpageSectionStakinghub.vue
rucister: I used the safer approach, removing alt text and using a `sq-only` span.

frontend/layers/products/app/components/ProductLandingpageSectionStakinghub.vue
rucister: I did not have enough time to try this yet but will definitely do.

frontend/layers/base/app/components/BaseButton.vue
```diff
@@ -1,21 +1,33 @@
...
+} = defineProps<
+  {
+    full?: boolean,
+    leadingIcon?: IconName,
+    size?: 'md' | 'xl',
+    trailingIcon?: IconName,
+    variant?: 'branded' | 'primary' | 'secondary',
+  }
+  // Make `to` and `disabled` mutually exclusive
+  & (
```
marcel-bitfly, line 17: **question:** did you test your implementation? I think this should not be optional (but I could be wrong)
```suggestion
  & (
    | { disabled: boolean, to?: never }
    | { disabled?: never, to: NuxtLinkProps['to'] }
  )
```

marcel-bitfly: also I thought using the `generic` in `script tag` would be necessary (it could be that this is obsolete now with a newer version of `nuxt/typescript`
```<script setup lang="ts" generic="WORKAROUND_FOR_CONDITIONAL_PROPS">```

also mentioning the link is a good idea as in `frontend/layers/base/app/components/BaseCard.vue`
// https://github.com/vuejs/core/issues/8952

frontend/layers/products/app/components/ProductLandingpageSectionStakinghub.vue
marcel-bitfly: if the video is not playing, we should be aware of it, but it is not a problem for now...the real issue would be the auto fullscreen

PR #1738, opened 2025-09-25
feat: add `search` section to `landing page`

enzo-bitfly:
See: FE-92


frontend/layers/products/app/components/BlockchainSearchResultItem.vue
```diff
@@ -0,0 +1,125 @@
...
+  if (result.chain_id === 560048) return hoodi
+  return mainnet
+})
+</script>
+
+<template>
+  <NuxtLink
+    v-if="result.type === 'address'"
+    class="flex gap-md"
+    :to="`${baseUrl}/address/${result.value.address.hash}?chain_id=${result.chain_id}`"
```
enzo-bitfly, line 23: I decided to add the `baseUrl` here and skip using the `global redirect middleware` altogether, since otherwise I'd have needed to pass the `chain_id` as a query param for the middleware to handle, and make everything worse there. 
Open to discuss, preferably after merging this.

frontend/layers/base/app/assets/css/main.css
```diff
@@ -47,8 +52,9 @@
   --container-3xs: 16rem; /* 256px */
   /* --container-7xl: 80rem; 1280px -> this is in conflict with --spacing-7xl */
   --container-8xl: 90rem; /* 1440px */
+  --container-24xl: 80rem; /* 1280px */
```
marcel-bitfly, line 57: **note**: we should be able to remove `--spacing-7xl` and so on with `--spacing: 0.25rem;` now

which will mean we do not need these high `xl values` anymore
(this could be done in a separate pr though)

frontend/layers/base/app/components/BaseNetworkIcon.vue
```diff
@@ -0,0 +1,23 @@
+<script setup lang="ts">
+import type { ChainId } from '~/types/network'
+import { useNetwork } from '~/layers/base/app/composables/useNetwork'
+
+const { id } = defineProps<{
+  id: ChainId,
```
marcel-bitfly: ** thought**(non-blocking): maybe being more explicit here would be better? 

I probably do not know what you mean with `id of BaseNetworkIcon` (could also be an internal thing)


```suggestion
  chainId: ChainId,
```

frontend/layers/base/app/components/BaseNetworkIcon.vue
```diff
@@ -0,0 +1,23 @@
...
+const networkName = computed(() => getNetworkName(id))
+</script>
+
+<template>
+  <LazyNetworkIconCryptoEth
+    v-if="networkName === 'Ethereum'"
+    class="w-2xl"
+  />
+  <LazyNetworkIconCryptoHoodi
+    v-if="networkName === 'Ethereum Hoodi'"
```
marcel-bitfly: **np**: maybe `short name` is more convenient here? 
```suggestion
    v-if="shortName === 'Hoodi'"
```

frontend/layers/base/app/components/BaseSearchInput.vue
```diff
@@ -0,0 +1,168 @@
...
+watchDebounced(
+  input,
+  async () => {
+    if (input.value.length) {
+      emit('search', input.value)
+      hasSearched.value = true
+    }
+  },
+  {
+    deep: true,
```
marcel-bitfly: **question**: are you sure this is needed, it should be the `default`?

See: https://vuejs.org/guide/essentials/watchers#deep-watchers

frontend/i18n/locales/en.json
```diff
@@ -6,7 +6,9 @@
         "beaconchain_homepage": "Beaconchain Homepage",
         "common": {
             "close": "Close",
+            "error_retry": "An error occurred. Please try again.",
```
marcel-bitfly, line 9: **issue**: in the template you are referencing : `error_try_again`

frontend/layers/base/app/components/BaseSearchInput.vue
```diff
@@ -0,0 +1,168 @@
...
+  const groupedResults = Object.groupBy(results, result => result[groupBy] as PropertyKey)
+
+  return Object.entries(groupedResults) as [SearchResultType[GroupBy], SearchResultType[]][]
+})
+</script>
+
+<template>
+  <form
+    role="search"
+    class="p-2xl dark:bg-[rgba(16,16,16,0.2)] rounded-3xl"
```
marcel-bitfly: **question:** why is `dark:bg-[rgba(16,16,16,0.2)]` not part of the `design system`?

Maybe we could give feedback to the designers.

frontend/layers/base/app/components/BaseSearchInput.vue
```diff
@@ -0,0 +1,168 @@
...
+      ignore-filter
+    >
+      <RkLabel
+        for="blockchain-search-input"
+        class="absolute bottom-2xl left-2xl dark:text-gray-400 text-sm-tight"
+      >
+        {{ label }}
+      </RkLabel>
+      <RkComboboxInput
+        id="blockchain-search-input"
```
marcel-bitfly: **question**: wouldn't `search-input` be specific enough for a `BaseSearchInput` component?
```suggestion
        id="search-input"
```

frontend/layers/base/app/components/BaseSearchInput.vue
```diff
@@ -0,0 +1,168 @@
...
+        v-model.trim="input"
+        type="search"
+        auto-focus
+        :aria-busy="isLoading"
+        :placeholder
+        class="search-input w-full text-2xl font-semibold rounded-3xl pt-3xl pr-5xl pb-6xl pl-2xl dark:bg-gray-950
+        dark:focus:bg-black placeholder:dark:text-gray-500 dark:text-white border-1 dark:border-charcoal-500
+        dark:focus:border-charcoal-50 dark:focus-within:outline-0"
+        @update:model-value="(value) => { if (!value) hasSearched = false }"
```
marcel-bitfly, line 79: **np**: probably better to use a function here, since it might be interessting where `hasSearched` would be `reset`

frontend/layers/base/app/components/BaseSearchInput.vue
marcel-bitfly, line 1: **thought:**since this seems not to be a generic component, probably it is best to place it in `products layer`

frontend/layers/base/app/composables/useBaseUrls.ts
```diff
@@ -0,0 +1,15 @@
+export const useBaseUrls = () => {
```
marcel-bitfly: **thought**: `baseUrl` normally refers to the `url` your `app` is deployed to

maybe something like `useEndpoint` would be clearer?

frontend/layers/products/app/components/BlockchainSearchInput.vue
```diff
@@ -0,0 +1,79 @@
...
+})
+
+const handleSearch = (input: string) => {
+  searchParams.value.input = input
+  emit('search', searchParams.value)
+}
+</script>
+
+<template>
+  <BaseSearchInput
```
marcel-bitfly, line 34: **thought**: maybe this is a bad name, is `BaseSearchInput` to generic enough to be used for any `Search` in our application no matter what `results` are expected?

frontend/layers/products/app/components/BlockchainSearchInput.vue
```diff
@@ -0,0 +1,79 @@
...
+import type {
+  InternalPostSearchResponse, PostSearchRequest,
+} from '~/types/api/search'
+import type { ChainId } from '~/types/network'
+
+export type SearchParams = Omit<PostSearchRequest, 'networks'> & { networks: ChainId[] }
+export type SearchResult = ReplaceChainId<InternalPostSearchResponse['data'][number]>
+type ReplaceChainId<T> = T extends { chain_id: any }
+  ? Omit<T, 'chain_id'> & { chain_id: ChainId }
+  : T
```
enzo-bitfly: Forgot to replace this with the `InternalPostSearchResponseWithChainId` type from the `search` server route

frontend/layers/products/app/components/BlockchainSearchResultItem.vue
```diff
@@ -0,0 +1,125 @@
...
+  >
+    <BaseNetworkIcon :id="result.chain_id" />
+    {{ result.value.address.hash }}
+  </NuxtLink>
+
+  <NuxtLink
```
marcel-bitfly, line 29: **suggestion:** you could think about either `prefetch-on="interaction"` or `no-prefetch` for all your `NuxtLinks` here

as probably this would otherwise result in a lot of unnecessary traffic for the users

marcel-bitfly (approved):

frontend/layers/products/app/components/BlockchainSearchInput.vue
enzo-bitfly: That's the idea, yes: it's a combobox which displays search results passed in it's `results` prop. We may need to adapt where styles are applied later if another `search` has e.g. a different input style, but the functionality is agnostic.
One thing I missed is ensuring that the results are an array of objects, as that is a requirement of the component to work correctly. Added now: `generic="SearchResultType extends Record<string, any>"`

frontend/i18n/locales/en.json
enzo-bitfly: thanks

frontend/layers/base/app/components/BaseNetworkIcon.vue
enzo-bitfly: I agree. Changed to `chainId`

frontend/layers/base/app/components/BaseNetworkIcon.vue
enzo-bitfly: Done, added a utility for this in `layers/base/utils/networks.ts`

frontend/layers/base/app/components/BaseSearchInput.vue
enzo-bitfly: It is generic

frontend/layers/base/app/components/BaseSearchInput.vue
enzo-bitfly: Good catch. This is a residue from having the model as an earlier implementation. Removed.

frontend/layers/base/app/components/BaseSearchInput.vue
enzo-bitfly: It's the `--color-black` with 0.2 opacity. I changed it now to use a `:before`, which caused me to refactor the `index` page with div used for background effects.

frontend/layers/base/app/components/BaseSearchInput.vue
enzo-bitfly: good catch, missed that when refactoring

frontend/layers/base/app/composables/useBaseUrls.ts
enzo-bitfly: True. Calling it `useApiUrl` now, as endpoint is the specific endpoints in the `API`

frontend/layers/products/app/components/BlockchainSearchResultItem.vue
enzo-bitfly: There's no prefetching happening, as the links lead to pages in a different application.

enzo-bitfly (commented):
Great review, thank you

frontend/layers/base/app/utils/networks.ts
```diff
@@ -0,0 +1,11 @@
+const ChainIDs = {
```
marcel-bitfly: **np:** maybe we comply with the js convention for `constants`? 
```suggestion
const CHAIN_ID = {
```

marcel-bitfly (approved):

frontend/layers/base/app/utils/networks.ts
enzo-bitfly: Good call. Done

PR #1741, opened 2025-09-30
Df 98/add section api

marcel-bitfly:


frontend/layers/base/app/components/BaseCardLayout.vue
```diff
@@ -0,0 +1,19 @@
...
+    :class="[
+      layout === '2fr_1fr' && 'md:grid-cols-[2fr_1fr] md:[&>*:nth-child(3)]:col-span-2',
+      layout === '1fr_1fr_1fr' && 'md:grid-cols-[1fr_1fr_1fr]',
+    ]"
+  >
+    <slot />
+  </div>
+</template>
+
+<style scoped></style>
```
enzo-bitfly, line 19: np: I see we're starting to encapsulate layout things into components but I am not sure it's the way to go. Do we really know that this abstraction will be used more times in the future? Also, the coupling with `BaseCard` seems too lose. In theory you could put any element other than a `BaseCard` in here, right?

enzo-bitfly (approved):

PR #1751, opened 2025-10-01
Fe 93/search filters

enzo-bitfly:


frontend/layers/products/app/pages/products/index.vue
rucister, line 377: suggestion: this width is causing horizontal scroll on the page, removing just a few pixels with calc should work ok given the dark design.

frontend/layers/products/app/pages/products/index.vue
```diff
@@ -103,12 +103,13 @@ const {
           v-model="searchParams"
           class="w-screen sm:w-full max-w-[920px] -ml-md sm:mx-auto"
```
rucister: Now that search is inside `ProductLandingpageSection` which has paddings on its own, `w-screen` will cause issues in small screens with horizontal scroll. From my test using just `w-full max-w-[920px] -ml-md sm:mx-auto` seems to work OK here.

frontend/layers/products/app/pages/products/index.vue
rucister, line 1: question:  Would it be better to encapsulate all search functionality inside it's component since in this case there is already a dedicated component? Or this search behavior is specific to this landing page and in another page will behave differently?

frontend/layers/base/app/components/BaseChipGroup.vue
```diff
@@ -0,0 +1,89 @@
...
+}
+</script>
+
+<template>
+  <ul
+    class="flex gap-md px-2xl py-lg"
+    role="group"
+    :aria-label
+  >
+    <li>
```
rucister, line 68: suggestion: I believe the `select all` chip should be optional if we want to reuse this component in other scenarios, or event not include at all in here but handle this specific logic on the parent component.

frontend/layers/base/app/components/BaseChip.vue
```diff
@@ -0,0 +1,23 @@
+<script setup lang="ts">
+const {
+  size = 'sm',
```
rucister, line 3: question: I guess the list of available sizes is to be expanded later as needed by the design?

frontend/layers/base/app/components/BaseChipGroup.vue
```diff
@@ -0,0 +1,89 @@
+<script setup lang="ts" generic="ChipItemValue extends string">
+type ChipItem = {
+  label: string,
+  value: ChipItemValue,
+}
+const {
+  ariaLabel,
+  items,
+} = defineProps<{
+  ariaLabel: string,
```
rucister, line 10: question: what is the standard practice here, technically speaking you don't really need to define this as prop but I guess it's a mandatory prop to ensure accessibility?

rucister (commented):
- I believe that the `all` option of the `BaseChipGroup` should at least be optional if we consider to reuse this component.
- I see that the search call logic is included in `products/index.vue` and not inside the dedicated `BlockchainSearchInput.vue`. Having all search logic encapsulated inside a dedicated component, would it make it more maintainable?

frontend/layers/products/app/pages/products/index.vue
enzo-bitfly: As far as I could see, it' wasn't this width but the one from the search input mentioned in another comment

frontend/layers/products/app/pages/products/index.vue
enzo-bitfly: Good catch, I missed this when changing to the component. I went with removing the search sections' `px` for the mobile breakpoint (as a fall-through attr), which also allowed me to remove the negative margin

frontend/layers/products/app/pages/products/index.vue
enzo-bitfly: This `BlockchainSearcInput` will probably only be used in the landing page, but as a general rule, I find it better to keep api calls in the root page - I find this makes data easier to track and keeps the component dumb and testable.

frontend/layers/base/app/components/BaseChipGroup.vue
enzo-bitfly: I agree that it could be optional, and I thought about doing it, but we don't have that use case right now and it would be very easy to make it conditional later.
About putting it in the parent, I don't think so, as the toggling logic can be considered to be responsibility of the group.
Open to discuss if you believe it would be better otherwise

frontend/layers/base/app/components/BaseChip.vue
enzo-bitfly: Exactly. We add stuff as needed.

frontend/layers/base/app/components/BaseChipGroup.vue
enzo-bitfly: Right, the purpose of the prop is to enforce it.

enzo-bitfly (commented):
Thank you

frontend/layers/base/app/components/BaseChipGroup.vue
rucister: I think it's ok to keep it like this for now and implement it later. Just didn't feel very comfortable to make a change  later which will potentially cause a change also on the `search`.

PR #1755, opened 2025-10-06
Fe 108/api pricing hero

enzo-bitfly:


frontend/layers/products/app/pages/products/pricing.vue
```diff
@@ -44,7 +44,23 @@ const items: BaseNavigationItem[] = [
...
+          {{ $t('products.api.hero.title') }}
+        </BaseHeading>
+        <BaseText
+          is="p"
+          size="2xl"
+          font-weight="bold"
+          class="container text-center sm:text-balance"
+        >
+          {{ $t('products.api.hero.subtitle') }}
+        </BaseText>
```
marcel-bitfly, line 65: **thought:** makes me think if `BaseText` was a good idea in the first place 🤔 

The component makes sense if it is easy to use from looking at the design I guess.
Bad if the components props grow even more.

marcel-bitfly (approved):

frontend/layers/products/app/pages/products/pricing.vue
enzo-bitfly: Same, I have never used a component to style text tbh. It kind of defeats the purpose of helper classes.
On the other hand, if we decide to keep it, we can expect it to have different variant up to some point.

PR #1756, opened 2025-10-06
feat(exporter): add functionality for a no downtime migration of lega…

peterbitfly:
…cy balances from clickhouse to bigtable


backend/pkg/exporter/modules/slot_exporter.go
```diff
@@ -502,6 +502,15 @@ func ExportSlot(client rpc.Client, slot uint64, isHeadEpoch bool, tx *sqlx.Tx) e
 			return nil
 		})
 
+		// save the validator balances to clickhouse
+		g.Go(func() error {
+			err := db.SaveLegacyValidatorBalancesToClickhouse(uint32(epoch), block.Validators)
+			if err != nil {
+				return fmt.Errorf("error exporting validator balances to bigtable for slot %v: %w", block.Slot, err)
```
invis-bitfly: **typo:** 
```suggestion
				return fmt.Errorf("error exporting legacy validator balances to clickhouse for slot %v: %w", block.Slot, err)
```

PR #1757, opened 2025-10-07
feat: add `enterprise plan section` to `api pricing page`

enzo-bitfly:
See: BEDS-106

### DIFF
--- a/frontend/layers/base/app/components/BaseCard.vue
+++ b/frontend/layers/base/app/components/BaseCard.vue
@@ -4,8 +4,15 @@ import type { NuxtLinkProps } from '#app'
 import type { BaseHeadings } from '~/layers/base/app/components/BaseHeading.vue'
 import type { IconName } from '~/layers/base/app/components/BaseIcon.vue'
 
-defineProps<
+const {
+  borderRadius = '4xl',
+  elevation = '1',
+  gap = '2xl',
+} = defineProps<
   {
+    borderRadius?: '4xl' | 'xl',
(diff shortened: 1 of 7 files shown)


frontend/layers/base/app/components/BaseCard.vue
```diff
@@ -4,8 +4,15 @@ import type { NuxtLinkProps } from '#app'
...
-defineProps<
+const {
+  borderRadius = '4xl',
+  elevation = '1',
+  gap = '2xl',
+} = defineProps<
   {
+    borderRadius?: '4xl' | 'xl',
+    elevation?: '1' | '2',
+    gap?: '5xl' | 'xl',
```
marcel-bitfly, line 15: **issue:** these are too many moving pieces

we should either speak with the design team, if there could be fewer options or we should assume that only the `elevation level change` this is provided as prop then.
Otherwise having styling options as props on a `BaseCard` is kinda pointless and we rather slap on `classes`.

In the end for example we want to have 2 or 3 different versions of a `BaseCard`.
```
<BaseCard
  elevation="1 | 2 | 3"
...
```

enzo-bitfly: Agreed. The thing is, higher elevation also means a card is contained by another card of lower elevation, so it is of a smaller grouping scale and its spacing will likely be smaller too. 
We can either only use `elevation` as a prop like you said, and set gaps with fall-through classes, or let elevation affect both bg color and gap - but I'm not sure that will work.

PR #1759, opened 2025-10-07
fix: untrap search focus when clicking outside

enzo-bitfly:
- show search dropdown scrollbar

See: FE-142

### DIFF
--- a/frontend/layers/base/app/components/BaseSearchInput.vue
+++ b/frontend/layers/base/app/components/BaseSearchInput.vue
@@ -1,4 +1,6 @@
 <script setup lang="ts" generic="SearchResultType extends Record<string, any>">
+import type { PointerDownOutsideEvent } from 'reka-ui'
+
 type GroupBy = keyof SearchResultType
 
 const {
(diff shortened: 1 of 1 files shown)


frontend/layers/base/app/components/BaseSearchInput.vue
```diff
@@ -82,12 +93,17 @@ const groupedResults = computed(() => {
...
         class="absolute z-10 dark:bg-gray-950 mt-xl rounded-xl w-full max-h-[400px]"
+        @pointer-down-outside="handleClickOutside"
       >
         <template v-if="$slots['dropdown-fixed-header']">
           <slot name="dropdown-fixed-header" />
         </template>
 
-        <RkComboboxViewport tabindex="-1">
+        <div
+          role="presentation"
```
marcel-bitfly, line 103: **question**: what do you want to achieve with `role=presentation`?

the `div` is already a `generic container`. If you wanna make sure that this is not exposed to the `accessibility tree`, you could add a `aria-hidden="true"`

marcel-bitfly (approved):

frontend/layers/base/app/components/BaseSearchInput.vue
enzo-bitfly: Not really because I want the content to still be exposed

PR #1773, opened 2025-10-15
fix: current `color mode`

marcel-bitfly:
Saving the current `color mode` in a cookie to avoid `hydration issues`, as this way the `server` already knows the current `color mode`.

It was necessary to use `useCookie()` to save the `color mode` instead of the `storage setting` of `nuxtjs/color-mode`, as the cookie value otherwise will be applied to the current `/path`. Resulting in different `color modes` for different `paths.

See: https://github.com/nuxt-modules/color-mode/pull/301


frontend/layers/base/app/components/BaseSwitch.vue
```diff
@@ -12,37 +12,41 @@ const { values } = defineProps<{
...
-const modelValue = defineModel<typeof values[number]['key']>()
+const modelValue = defineModel<typeof values[number]['key']>({ required: true })
+const defaultValue = modelValue.value
 const { t: $t } = useTranslation()
-const thumb = useTemplateRef('thumb')
+const thumbs = useTemplateRef('thumb')
 const track = useTemplateRef('track')
 
 const moveThumb = () => {
+  // console.log('👉', thumbs.value)
```
enzo-bitfly: issue: please remove debug comment

frontend/layers/base/app/components/BaseSwitch.vue
```diff
@@ -104,6 +108,37 @@ watch(modelValue, () => {
           type="radio"
           :name
         >
+      </label> -->
```
enzo-bitfly: question: does this code need to stay? also, what's the reason to let the `BaseSwitch` map through any amount of values instead of fixing it to only two?

frontend/layers/base/app/components/BaseSwitch.vue
```diff
@@ -12,37 +12,41 @@ const { values } = defineProps<{
...
-  const animation = thumb.value.animate([ {
+  const animation = thumb.animate([ {
     transform: `translateX(${x - initialX}px)`,
   } ],
   { duration: 180 },
   )
   return animation.finished.then(() => {
-    activeTrackItem.appendChild(thumb.value!)
+    if (!thumb) return
+    activeTrackItem.appendChild(thumb)
```
enzo-bitfly, line 46: praise: thanks for crearing a local variable to remove the need for `!`

enzo-bitfly (approved):

frontend/layers/base/app/components/BaseSwitch.vue
marcel-bitfly: sorry I overlooked that

marcel-bitfly: I am using a loop instead of putting the thumb into the first element, as we also have to handle the 
`defaultPosition`

frontend/layers/base/app/components/BaseSwitch.vue
marcel-bitfly: sorry I removed it

frontend/layers/base/app/components/BaseColormode.vue
```diff
@@ -0,0 +1,47 @@
+<script setup lang="ts">
+const colorMode = useColorMode()
+const cookie = useBcCookie<'dark' | 'light'>('theme')
+colorMode.preference = cookie.value || 'dark'
+const handleUpdate = () => {
+  cookie.value = colorMode.preference as 'dark' | 'light'
+  // due to a bug in @nuxtjs/color-mode
+  // `data-theme` does not update when after user comes back to the app
```
enzo-bitfly:
```suggestion
  // `data-theme` does not update when user comes back to the app
```

PR #1778, opened 2025-10-16
feat: add `prefill buttons` to search on `product landing page`

marcel-bitfly:
See: FE-94

### DIFF
--- a/frontend/layers/products/app/pages/products/index.vue
+++ b/frontend/layers/products/app/pages/products/index.vue
@@ -44,12 +44,13 @@ const searchTypes: BlockchainSearchParams['types'] = [
   'validator_by_public_key',
   'validators_by_withdrawal_credential',
 ]
+const networks = [
+  1,
+  560048,
+] satisfies ChainId[]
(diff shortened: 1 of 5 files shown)


frontend/layers/products/app/pages/products/index.vue
```diff
@@ -71,6 +72,37 @@ const handleSearch = (input: string) => {
   }
   execute()
 }
+
+const handleExampleClick = (type: 'address' | 'token' | 'transaction' | 'validator') => {
```
rucister, line 76: suggestion: maybe it would it better to expose as type possible return values of `click:example` from `BlockchainSearchInput.vue` and reference here to ensure it satisfies the list.

rucister (approved):
Looks good to me with current implementation. Left just one very minor (non-blocking) suggestion.`

frontend/layers/products/app/pages/products/index.vue
marcel-bitfly: I will revisit this with FE-159

PR #1796, opened 2025-10-29
feat: add `registered trademark symbol` to `BeaconScore`

enzo-bitfly:
See: BC-47

### DIFF
--- a/frontend/components/bc/BcTranslation.vue
+++ b/frontend/components/bc/BcTranslation.vue
@@ -1,29 +1,40 @@
 <script lang="ts" setup>
+/**
+ * This component provides two approaches for translation interpolation:
+ * 1. Props (simple patterns): suppath, linkpath, boldpath, listpath
+ * 2. Named slots (complex/dynamic content): for computed values or multiple interpolations
+ *
+ * @example Simple - using props
+ * <BcTranslation
+ *   keypath="base.common.beaconscore"
+ *   suppath="base.common.registered_trademark_symbol"
+ * />
+ *
+ * @example Complex - using named slots
+ * <BcTranslation keypath="dashboard.validator.summary.tooltip.higher">
+ *   <template #sup><sup>®</sup></template>
+ *   <template #name>{{ groupName }}</template>
+ *   <template #average>{{ formatPercent(value) }}</template>
+ * </BcTranslation>
+ */
 defineProps<{
+  /** Translation key for bold text content */
   boldpath?: TranslationKey,
-  /**
-   * The path to the key in the translation file (e.g. en.json)
-   */
+  /** Translation key (required) */
   keypath: TranslationKey,
+  /** Translation key for link text (requires `to` prop) */
   linkpath?: TranslationKey,
+  /** Translation key for newline-separated list items */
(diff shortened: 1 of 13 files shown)


frontend/components/bc/BcTranslation.vue
```diff
@@ -32,40 +43,55 @@ defineProps<{
...
-    <template #bold>
-      <span
-        v-if="boldpath"
-        class="bc-translation-bold"
-      >{{ $t(boldpath) }}</span>
+    <!-- Forward custom named slots (use only for dynamic/complex content) -->
+    <template
+      v-for="(_, name) in slots"
+      :key="String(name)"
+      #[name]
```
rucister, line 50: After a further research on this, the below implementation seems to be the more correct one.
`
<template
  v-for="(_, name) in slots"
  :key="String(name)"
  #[name]="slotProps"
>
  <slot :name="String(name)" v-bind="slotProps || {}" />
</template>
`
The reasoning about this seems to be:
> <component :is="_" /> is incorrect because slots are functions, not components.

frontend/components/bc/BcTranslation.vue
```diff
@@ -32,40 +43,55 @@ defineProps<{
...
-          class="link"
-          target="_blank"
-          :to
-        >
-          {{ $t(linkpath) }}
-        </BcLink>
-      </slot>
+
+    <!-- Prop-based templates (preferred - use these via props when possible) -->
+    <template
```
rucister, line 58: question: is there a risk that these get rendered twice now that the component supports dynamic slots forwarding?

frontend/components/bc/BcTranslation.vue
```diff
@@ -1,29 +1,40 @@
...
+ *   suppath="base.common.registered_trademark_symbol"
+ * />
+ *
+ * @example Complex - using named slots
+ * <BcTranslation keypath="dashboard.validator.summary.tooltip.higher">
+ *   <template #sup><sup>®</sup></template>
+ *   <template #name>{{ groupName }}</template>
+ *   <template #average>{{ formatPercent(value) }}</template>
+ * </BcTranslation>
+ */
```
rucister, line 19: 🎉 Nice work!

rucister (changes requested):

frontend/components/bc/BcTranslation.vue
enzo-bitfly: that's right, so I am replacing `<component is...` by a slot now. 
But passing `slotProps` causes an error when the slot content is reactive and initially null|undefined (`Cannot read properties of null (reading 'key')` as we saw yesterday). `slotProps` are not needed anyway because `I18nT` doesn't pass scoped slot values in the slots.

frontend/components/bc/BcTranslation.vue
enzo-bitfly: Good question. No, because the explicitly defined templates only render if the slot for the given name is not defined: `v-if="subpath && !slots.sup"`

enzo-bitfly (commented):
Great review, thanks

rucister (approved):

